PR #39, opened 2026-04-25
fix(ci): rename AISIX_REDIS_URL → CACHE_TEST_REDIS_URL to unblock CI

moonming:
CI on `main` has been red since #34 because `aisix-core`'s `Config` loader merges every `AISIX_*` env var into the root `Config` struct (via config-rs `Environment::with_prefix("AISIX")`), and `Config` has `#[serde(deny_unknown_fields)]`.

The redis integration test sets `AISIX_REDIS_URL` on the `rust-unit` job, which leaks into every `Config::load_from_path` call as a stray `redis_url` field and panics:

```
Config("deserialize: unknown field \`redis_url\`, expected one of \`etcd\`, \`proxy\`, \`admin\`, \`observability\`, \`cache\`, \`managed\`")
```

8 / 9 `aisix-core::config::tests` fail (the one that passes is `rejects_unknown_fields` — it swallows the error intentionally).

## Fix

Rename the env var so it doesn't sit under the `AISIX_` prefix at all:

- `crates/aisix-cache/tests/redis_integration.rs` reads `CACHE_TEST_REDIS_URL`
- `.github/workflows/ci.yml` sets `CACHE_TEST_REDIS_URL`
- `crates/aisix-cache/src/redis.rs` doc comment + `docs/testing.md` updated to match

## Test plan

- [x] `cargo test -p aisix-core --lib config::tests` — all 9 pass with the rename
- [x] Repro the original failure: `AISIX_REDIS_URL=… cargo test … loads_minimal_config` still fails (so the loader behaviour for real `AISIX_*` overrides is unchanged)
- [ ] CI green on this PR

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,7 +51,14 @@ jobs:
     env:
       # Picked up by crates/aisix-cache/tests/redis_integration.rs.
       # The tests no-op when this is unset (local dev), so absence is safe.
-      AISIX_REDIS_URL: redis://127.0.0.1:6379
+      #
+      # MUST NOT start with the `AISIX_` prefix — Config::load_from_path
+      # merges every AISIX_* env var into the root Config (via
+      # config-rs Environment::with_prefix("AISIX")), and Config has
+      # `#[serde(deny_unknown_fields)]`. An AISIX_REDIS_URL leaks into
+      # every Config::load_from_path test as `redis_url` and breaks
+      # the entire `aisix-core::config::tests` module.
+      CACHE_TEST_REDIS_URL: redis://127.0.0.1:6379
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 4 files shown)


moonming:
✅ Rust tests **all pass** with the rename — confirmed via the CI log:

\`\`\`
test result: ok. 70 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
test result: ok. 102 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
... (every aisix-* crate green)
\`\`\`

The job is showing as failed because of:

\`\`\`
##[error]Failed to CreateArtifact: Artifact storage quota has been hit.
\`\`\`

That's an org-level Actions storage quota issue (independent of this PR — \`actions/upload-artifact@v4\` rejects regardless of test outcome). Either free up artifacts or upgrade the plan to make CI pass cleanly. The fix itself is verified end-to-end.

moonming:
Superseded by #40, which combines this PR's workflow rename with the supervisor cache-write fix and the artifact-quota mitigation. Merged as 6bf62278.

PR #84, opened 2026-05-05
feat(cache): honor applies_to scope on the policy gate

moonming:
P0a from the cache-policies audit. CachePolicy.applies_to was promised in the dashboard since Stage 1 but the DP gate ignored it. This PR adds CachePolicy::applies_to_request(model, api_key_id) and gates the cache by it. Strict semantics: unknown prefix → no match (was: silently 'all'). 6 new tests. cargo test + clippy + fmt clean.

### DIFF
--- a/crates/aisix-core/src/models/cache_policy.rs
+++ b/crates/aisix-core/src/models/cache_policy.rs
@@ -123,6 +123,38 @@ impl CachePolicy {
         self.runtime_id = id.into();
         self
     }
+
+    /// Returns true when this policy's `applies_to` string targets the
+    /// given (model name, api_key uuid) pair.
+    ///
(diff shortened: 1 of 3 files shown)


moonming:
Independently shipped in #82 (merged before this could land). The matcher behavior is functionally equivalent — same grammar (`all` / `model:<name>` / `api_key:<uuid>`), same gate predicate. Only difference: #82's unknown-value fallback is `AppliesTo::All` (conservative — relies on cp-api validation to reject typos), this PR was strict (`no match` — surfaces typos via `cache_status=disabled`). Either is defensible; main goes with the conservative branch.

Closing without merging. The `per-policy TTL` follow-up (#85) rebases onto #82's API.

PR #181, opened 2026-05-09
docs(managed-mode): document cert-bundle bootstrap path + fix image tag

moonming:
## Summary

The README's "what's shipped" line calls out the cert-bundle bootstrap path (no \`/dp/register\` round-trip, PEMs supplied as env vars) as a first-class managed-mode option, but \`docs/managed-mode.md\` had **zero** coverage of it. An operator following the doc could only set up the register-based path and would miss the recommended one entirely.

This PR fills that gap. Source-of-truth: \`crates/aisix-server/src/cert_bundle.rs\` and \`crates/aisix-core/src/config.rs\`.

## Changes

1. **New "First boot — pre-provisioned cert bundle (recommended)" section.** Three required PEMs, both inline and file-path env-var variants, and how \`env_id\` + \`dp_id\` are parsed out of the leaf cert's URI SAN (\`x-aisix://env/&lt;env_id&gt;\`, \`x-aisix://dp/&lt;dp_id&gt;\`) to scope etcd reads to \`/aisix/&lt;env_id&gt;/\`.

2. **Re-titled the existing register section** to "Alternative: register against the control plane" so the recommended path is documented first.

3. **Env-var override table** gained:
   - \`AISIX_MANAGED__CP_ETCD_ENDPOINT\` (required for cert-bundle path)
   - \`AISIX_MANAGED__CP_CA_CERT_FILE\` (optional trust-anchor override)
   - The six cert-bundle slots: \`CP_CERT_PEM\` / \`_FILE\`, \`CP_KEY_PEM\` / \`_FILE\`, \`CP_CA_PEM\` / \`_FILE\` (one of inline/file per slot must be set)

4. **Restart semantics** now covers three first-boot branches (cert-bundle / register / fail-fast) instead of two, and notes that \`env_id\` is persisted alongside \`dp_id\` (the doc previously listed only \`dp_id\`).

5. **Image tag** corrected from \`ghcr.io/moonming/ai-gateway:main\` → \`ghcr.io/api7/ai-gateway:main\` to match the README and the actual published image.

## Out of scope

Companion PRs in the doc-drift series: #179 (architecture metric names + Cohere), #180 (testing.md npm→pnpm + CI diagram). Still to come: \`api-admin.md\` schema-truth pass and \`api-proxy.md\` error/header drift (latter stacked behind #178).

## Test plan

- [x] Doc-only change; no code touched
- [x] All env-var names verified against \`crates/aisix-core/src/config.rs:163-279\`
- [x] cert-bundle flow verified against \`crates/aisix-server/src/cert_bundle.rs:35-120\`
- [x] Image tag matches \`README.md:100\`

### DIFF
--- a/docs/managed-mode.md
+++ b/docs/managed-mode.md
@@ -39,7 +39,8 @@ docker run --rm \
   -e AISIX_MANAGED__CP_CERT_PEM="$(cat leaf.crt)" \
   -e AISIX_MANAGED__CP_KEY_PEM="$(cat leaf.key)" \
   -e AISIX_MANAGED__CP_CA_PEM="$(cat ca.crt)" \
-  -e AISIX_MANAGED__CP_ETCD_ENDPOINT=https://dp-manager.aisix.cloud:7943 \
+  -e AISIX_MANAGED__CP_BASE_URL=https://api.us.aisix.cloud \
+  -e AISIX_MANAGED__CP_ETCD_ENDPOINT=dp-manager.aisix.cloud:7943 \
   -v aisix-mtls:/var/lib/aisix \
   -p 3000:3000 \
   ghcr.io/api7/ai-gateway:main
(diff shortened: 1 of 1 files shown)


moonming:
## Audit response

Independent audit on commit \`1dc0fec\` returned BLOCK (3 HIGH, 3 MEDIUM). Addressed in commit \`8487dbd\`:

| Finding | Action |
|---|---|
| **H1** image tag should be \`ghcr.io/moonming/ai-gateway\` | **Rejected — audit followed a stale comment.** \`.github/workflows/docker-image.yml:35\` has \`IMAGE_NAME: \${{ github.repository }}\` with a comment \`# ghcr.io/moonming/ai-gateway\` that's left over from when the repo lived under that user. \`gh api repos/api7/ai-gateway\` confirms the canonical repo path is \`api7/ai-gateway\` (\`fork=false\`); \`gh api repos/moonming/ai-gateway\` redirects to it. So \`\${{ github.repository }}\` evaluates to \`api7/ai-gateway\` at runtime — the image really is published as \`ghcr.io/api7/ai-gateway:main\`, matching the README. The stale comment in the workflow file should be cleaned up in a follow-up PR but does not change what we should document. |
| **H2** cert-bundle example missing \`CP_BASE_URL\` | Fixed — added the env var to the docker example and updated the per-flag description to clarify cert-bundle skips \`/dp/register\` but still needs \`CP_BASE_URL\` for the heartbeat origin. \`crates/aisix-server/src/main.rs:154\` returns the hard-error if it's missing. |
| **H3** \`CP_ETCD_ENDPOINT\` example had \`https://\` prefix | Fixed — bare host:port. \`main.rs:173\` does \`format!(\"https://{cp_etcd}\")\` so the prefix would have been doubled. |
| **M1** table row direction inverted | Fixed — \`CP_ETCD_ENDPOINT\` is required for register path (\`main.rs:213\`); cert-bundle path falls back to \`cp_base_url\` host:port (\`main.rs:161-166\`). |
| **M2** fabricated \`gateway_id=aigg_xxx\` log line | Fixed — replaced with the actual emitted fields (\`dp_id\`, \`env_id\`, \`etcd\`) and split into cert-bundle vs register branches so operators can grep either. |
| **M3** restart-semantics ordering vs code priority | Fixed — reworded step #1 to "no bundle on disk yet, cert-bundle env vars set" to mirror code's \`if !bundle_on_disk && bundle_provided\` predicate. |

Issue #127, opened 2026-05-09
[Testing] Test coverage gap analysis vs LiteLLM — DP supplementation roadmap


moonming:
## Context

Against the most mature LLM gateway in the wild — **LiteLLM (BerriAI/litellm)** — we audited the ai-gateway test suite end-to-end and identified concrete gaps. This is the meta-tracking issue for the supplementation roadmap.

> Companion CP issue: [api7/AISIX-Cloud#196](https://github.com/api7/AISIX-Cloud/issues/196) (cross-repo coordination).
> Full plan document is kept in our internal knowledge base: `test-plan-litellm-parity.md`.

## Snapshot

| Dimension | ai-gateway (DP) | LiteLLM baseline |
|-----------|-----------------|------------------|
| Test files | 86 embedded + 2 integration + 1 e2e smoke | 1509+ across 43 top-level dirs |
| Test fns | 572 `#[test]` / `#[tokio::test]` | tens of thousands |
| Provider coverage | 4 (OpenAI/Anthropic/Gemini/DeepSeek) | 30+ |
| E2E cases | 1 smoke (~700 LoC Vitest) | 100+ |

**Strengths**: 572 embedded units, wiremock isolation, Bedrock guardrail fail-open/fail-closed matrix, RPM/TPM/concurrency three-axis rate limit.
**Weaknesses**: thin E2E, narrow provider matrix, shallow streaming depth, missing error-recovery state machines.

---

## 🔴 P0 — production blockers

- [x] **G1 — Streaming + token accumulation accuracy**  _(verified done on origin/main 56eaa31 — PR #394 (openai-tools-roundtrip + streaming-usage-accumulation e2e))_
  - Site: `crates/aisix-proxy/src/streaming_tests.rs`
  - Assert: MockServer pushes N SSE chunks; final token count == non-streaming same input.
  - Ref: LiteLLM `test_stream_chunk_builder.py`

- [x] **G2 — Streaming + tool_calls delta assembly**  _(verified done on origin/main 56eaa31 — PR #394 (openai-tools-roundtrip-e2e))_
  - Site: same file as G1
  - Assert: `tool_calls.arguments` correctly concatenated across chunks
  - Ref: LiteLLM streaming function-call tests

- [ ] **G3 — Circuit breaker / failure recovery**
  - Site: `routing/circuit_breaker_tests.rs`
  - Assert: 5 consecutive 5xx → `open` → `half-open` → `closed` state machine
[...]

Issue #173, opened 2026-05-09
bug: concurrency cap rejection emits `rate_limit_exceeded` instead of `concurrency_limit_exceeded` (docs §2 publishes distinct types)


moonming:
## Summary

\`docs/api-proxy.md\` §2 status→\`error.type\` table publishes a 429 family with **three distinct** error types:

> | 429 | \`rate_limit_exceeded\` / \`concurrency_limit_exceeded\` / \`budget_exceeded\` | RPM/TPM/concurrency/budget cap |

The intent is clear from how the docs map types to causes: **RPM/TPM cap → \`rate_limit_exceeded\`**, **concurrency cap → \`concurrency_limit_exceeded\`**, **budget cap → \`budget_exceeded\`**. Distinct types let SDK callers branch error-handling appropriately.

In practice, when an ApiKey is configured with \`rate_limit: { rpm: 100, concurrency: 1 }\` and a second simultaneous request arrives while the first is in-flight, the gateway returns:

\`\`\`
HTTP/1.1 429
{"error": {"type": "rate_limit_exceeded", "message": "..."}}
\`\`\`

Expected per docs:
\`\`\`
HTTP/1.1 429
{"error": {"type": "concurrency_limit_exceeded", "message": "..."}}
\`\`\`

## Repro

E2E test attached: drafted as case 2 of \`tests/e2e/src/cases/concurrency-e2e.test.ts\` (held back from #151 C9 PR pending this fix).

\`\`\`ts
await admin.createApiKey({
  key_hash: callerHash,
  allowed_models: ["conc-cap"],
  rate_limit: { rpm: 100, concurrency: 1 },  // RPM=100 (effectively unlimited), concurrency=1
});

// Mock upstream has 1.5s response delay; fire 2 requests, the 2nd arrives while 1st is in-flight.
const first = client.chat.completions.create({ /* ... */ });
await new Promise(r => setTimeout(r, 50));  // ensure first reaches gateway first
const second = client.chat.completions.create({ /* ... */ });

await first;  // succeeds
await second;
// Result: 429, error.type === "rate_limit_exceeded"
// Expected per docs §2: error.type === "concurrency_limit_exceeded"
\`\`\`

## Why this matters

- Operators / agent frameworks routinely branch on \`error.type\` to decide retry strategy:
  - \`rate_limit_exceeded\` → wait until the RPM/TPM window resets (seconds to minutes), then retry the same request.
[...]

PR #186, opened 2026-05-10
test(e2e): G2 streaming tool_calls — fragments assemble to valid JSON

moonming:
## Summary

Closes the **G2** gap from [issue #127](https://github.com/api7/ai-gateway/issues/127) — currently \`[ ]\` Missing per the [2026-05-10 audit comment](https://github.com/api7/ai-gateway/issues/127#issuecomment-4412955037). Audit observed: zero \`tool_calls\` / \`function\` references across all 22 existing e2e files.

Pins one externally-observable contract: when the upstream streams a tool invocation per [OpenAI's wire shape](https://platform.openai.com/docs/api-reference/chat/streaming) — first chunk carries the tool_call \`id\`, \`type\`, \`function.name\` plus an \`arguments\` fragment; subsequent chunks carry only \`arguments\` fragments under the same \`index\`; terminal chunk carries \`finish_reason = "tool_calls"\` — the gateway preserves that shape end-to-end.

## Assertions

The caller must be able to assemble a single tool_call with:

1. **Stable id + function.name** — declared exactly once on the first tool_call delta. Subsequent fragments under the same index must NOT redeclare them. Catches a regression that re-emits id/name on every chunk (still produces valid concatenated args but breaks consumers that rely on id-set-once).
2. **Byte-exact arguments concatenation** — every per-chunk \`arguments\` fragment joins to the upstream's exact string and parses to the canonical object \`{"location":"Beijing"}\`. Catches a regression that drops or reorders fragments.
3. **Single \`finish_reason\`** — arrives exactly once on a chunk with empty delta, value \`"tool_calls"\`. Catches a regression that swallowed the terminal SSE event or duplicated it.
4. **Terminal-chunk-is-last** — no \`tool_calls\` fragments after the finish chunk. Catches a regression that flushed a stray fragment after the terminator.

## Mock-upstream design

Arguments \`{"location":"Beijing"}\` is split as \`'{"' / 'loc' / 'ation":"Bei' / 'jing"}'\` so fragment boundaries fall **mid-key, mid-value, and mid-quote**. A trivial assembler that only checks for natural boundaries (e.g. only re-joins on closing brace) would fail JSON parsing.

## Source-blind contract design

Per CLAUDE.md §5, every expected value derives from upstream-emitted bytes + JSON parsing semantics — not from reading the gateway's SSE handler:

- The 5 expected chunks come from the 5 SSE events I emit.
- The \`finish_reason\` value \`"tool_calls"\` is the value I write into the terminal SSE event.
- The canonical \`{"location":"Beijing"}\` is the byte-concatenation of the fragments I split.

## Out of scope

- G1 streaming-usage-accumulation, G3 circuit-breaker, G8 vision — separate follow-up PRs.

## Test plan

- [x] Mirrors the structure of \`cross-provider-matrix-e2e.test.ts\` streaming branch (the existing canonical streaming test pattern).
- [x] Uses only existing harness primitives (\`startOpenAiUpstream({streamEvents})\`, \`AdminClient\`, OpenAI SDK).
- [ ] Run locally against etcd + binary; the file will skip silently in environments without etcd, matching the rest of the suite.

## Summary by CodeRabbit

* **Tests**
  * Added an end-to-end streaming test that verifies correct sequencing and assembly of fragmented tool-call arguments across streaming chunks and that the terminal finish reason is emitted exactly once.
  * Confirms the gateway issues a single streamed upstream request with the expected model, tools, and authorization during the streaming interaction.

[Review Change Stack](https://app.coderabbit.ai/change-stack/api7/ai-gateway/pull/186)

### DIFF
--- a/tests/e2e/src/cases/streaming-tool-calls-e2e.test.ts
+++ b/tests/e2e/src/cases/streaming-tool-calls-e2e.test.ts
@@ -139,6 +139,12 @@ describe("streaming tool_calls e2e: arguments fragments concatenate to a valid J
       }
     });
 
+    // Baseline-isolate the readiness probe so the upstream wire-shape
+    // assertion below measures only the actual test call. Without
+    // this, the probe (which also goes through the streaming path)
+    // would inflate the request count.
+    const upstreamBaseline = upstream.receivedRequests.length;
+
     const stream = await client.chat.completions.create({
       model: "stream-tools-model",
       messages: [
@@ -200,10 +206,14 @@ describe("streaming tool_calls e2e: arguments fragments concatenate to a valid J
     expect(firstWithTool?.idDelta).toBe(TOOL_CALL_ID);
     expect(firstWithTool?.nameDelta).toBe(TOOL_NAME);
 
-    const subsequentWithTool = captured
-      .slice(captured.indexOf(firstWithTool!) + 1)
-      .filter((c) => c.argsDelta !== undefined);
-    for (const c of subsequentWithTool) {
+    // Sweep every chunk after the first tool_call delta — including
+    // empty-delta chunks like the terminal finish chunk — to catch a
+    // regression that re-emits id/name on a chunk that has no
+    // arguments fragment (which an args-filtered sweep would miss).
+    const afterFirst = captured.slice(
+      captured.indexOf(firstWithTool!) + 1,
+    );
+    for (const c of afterFirst) {
       expect(c.idDelta).toBeUndefined();
       expect(c.nameDelta).toBeUndefined();
     }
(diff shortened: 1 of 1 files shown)


moonming:
Closing this PR — the e2e test it adds correctly surfaces a real product bug, not a test-author defect:

> Streaming chat completions silently drop \`tool_calls\` deltas — \`crates/aisix-provider-openai/src/wire.rs:43\` says \`ChatChunk\` "doesn't model (\`tool_calls\`, \`refusal\`, \`audio\`, ...)" and \`stream_chunk_into_chat_chunk\` (line 242) drops them on the OpenAI provider's streaming path.

Filed as #202 with the full repro and proposed fix direction (native SSE passthrough for OpenAI streams).

Holding the test back from main per the project policy on test-surfaced product bugs (same pattern as #173). Will land — unmodified — once #202 ships. The branch \`test/e2e-g2-streaming-tool-calls\` stays for reference.

Issue #193, opened 2026-05-10
bug: `proxy.request_body_limit_bytes` config field is never read


moonming:
## Summary

The \`proxy.request_body_limit_bytes\` config field exists on \`ProxyConfig\` and is propagated through to \`ProxyState.request_body_limit_bytes\` (\`crates/aisix-proxy/src/state.rs:58, 75, 99, 126\`), but it is **never read**. \`build_router\` (\`crates/aisix-proxy/src/lib.rs:59\`) wires no \`DefaultBodyLimit\` / \`RequestBodyLimitLayer\` / \`tower_http::limit\` that consults the field.

In practice, the active limit on every \`/v1/*\` endpoint that uses \`Json<…>\` is whatever axum's built-in \`Json<…>\` extractor defaults to (2 MiB at axum 0.7+). Operators raising \`request_body_limit_bytes\` to e.g. 25 MB to support vision payloads, audio uploads, or batched embeddings will see no effect.

## Repro

\`\`\`yaml
# config.yaml
proxy:
  addr: 0.0.0.0:3000
  request_body_limit_bytes: 26214400  # 25 MB
\`\`\`

\`\`\`bash
# Send a chat completion with a 5 MB vision payload (well under
# the configured 25 MB but well over axum's 2 MiB default).
curl -X POST http://localhost:3000/v1/chat/completions \\
  -H "Authorization: Bearer sk-…" \\
  -H "Content-Type: application/json" \\
  -d @5mb-vision-request.json
\`\`\`

Actual: 413.
Expected: 200 (or whatever the upstream returns).

## Severity

**MEDIUM**: not a security or data-corruption issue, but a silent dead config that misleads operators. Surfaced while updating \`docs/api-proxy.md\` (PR #191) — the doc previously claimed the field was enforced; PR #191 corrects the doc to match reality and links here.

## Proposed fix

Wire \`DefaultBodyLimit::max(state.request_body_limit_bytes as usize)\` (or the per-route \`tower_http::limit\` equivalent) into \`build_router\` before merging the route trees. Cover with a unit test that POSTs a body just over the configured limit and asserts 413; and one just under the limit asserting non-413.

## Audit trail

Surfaced by the independent audit on PR #191 (HIGH-2 finding — "request_body_limit_bytes is not yet enforced").

moonming:
[...]

Issue #202, opened 2026-05-10
bug: streaming chat completions silently drop tool_calls deltas


moonming:
## Summary

When the upstream returns SSE chunks containing \`choices[].delta.tool_calls\` (the standard OpenAI shape for streamed tool invocations), the gateway forwards each chunk to the caller WITHOUT the tool_calls field — \`delta.tool_calls\` is consistently \`undefined\` on every client-observed chunk. The gateway preserves \`role\`, \`content\`, and \`finish_reason\` deltas correctly; only \`tool_calls\` is lost.

## Root cause (high level)

\`ChatChunk\` (gateway-internal stream type) does not model \`tool_calls\` —
[\`crates/aisix-provider-openai/src/wire.rs:43\`](https://github.com/api7/ai-gateway/blob/main/crates/aisix-provider-openai/src/wire.rs#L43):

> doesn't model (\`tool_calls\`, \`refusal\`, \`audio\`, ...)

So \`stream_chunk_into_chat_chunk\` (line 242) drops them when converting \`OpenAiStreamChunk → ChatChunk\`. The non-streaming round-trip preserves tool_calls (test \`assistant_tool_calls_round_trip_into_upstream_request\` at line 515), but the streaming code path is asymmetric.

## Repro

E2E test attached: \`tests/e2e/src/cases/streaming-tool-calls-e2e.test.ts\` at PR #186 (commit dbef406). Mock upstream emits a 5-chunk SSE stream where chunk 1 carries \`tool_calls[0].function.{name,arguments}\` and the first arg fragment, chunks 2-4 carry argument fragments. Client iterates the stream and inspects every chunk; assertion \`expect(firstWithTool).toBeDefined()\` fails because no chunk surfaces any tool_call delta.

CI repro: <https://github.com/api7/ai-gateway/actions/runs/25620996846/job/75207563215>

\`\`\`
AssertionError: expected undefined to be defined
❯ src/cases/streaming-tool-calls-e2e.test.ts:214:27
\`\`\`

## Why this matters

Tool-using chat completions over streaming are a core OpenAI feature today. Every Python/Node SDK that targets OpenAI assumes \`delta.tool_calls\` arrives in SSE chunks per the public API spec. A regression-free repointing of the SDK at aisix is the gateway's compatibility goal (\`docs/api-proxy.md\` §0); silently swallowing tool_calls breaks every framework that depends on streamed function-calling — instructor, langgraph, openai-agents-py, etc.

## Severity

[...]

PR #215, opened 2026-05-10
test(e2e): boundary defense — 413 on oversized JSON body

moonming:
## Summary

Pins two contracts on the proxy's body-size enforcement (closes part of the P2 boundary-defenses gap from issue #127):

1. A JSON body exceeding the documented 2 MiB limit returns **HTTP 413**. A regression that removed or silently raised the limit would let an attacker push unbounded JSON at the proxy and exhaust memory.

2. A body well under the limit does **NOT** return 413. Catches a regression that lowered the limit too aggressively (e.g. 64 KiB), which would reject legitimate tool / vision payloads.

## Why this matters

DoS protection on the body extractor is the only thing keeping a single greedy caller from OOM-ing the proxy. Today's 2 MiB number is axum's \`Json<…>\` extractor default, documented in [\`docs/api-proxy.md\` §2](https://github.com/api7/ai-gateway/blob/main/docs/api-proxy.md#2-error-envelope) (PR #191). The \`proxy.request_body_limit_bytes\` config field is currently unused (issue #193); when it gets wired, this test becomes the sentinel that surfaces the change.

## Three orthogonal assertions

1. \`oversizedRes.status === 413\` on a 2.5 MiB body.
2. \`upstream.receivedRequests.length\` did NOT increment on the rejected request — proves the 413 short-circuited BEFORE upstream dispatch. Catches a regression that streamed the body upstream first and only 413'd on the response leg, which would waste upstream tokens AND leak the caller's payload to a third party.
3. \`undersizedRes.status === 200\` on a 256 KiB body — same code path, only content size differs.

Plus pre-send \`Buffer.byteLength\` assertions confirm the test bodies really straddle the documented limit, so a future limit raise surfaces as a clear test-side mismatch rather than passing for the wrong reason.

## Source-blind compliance

Per CLAUDE.md §5: 2 MiB is the publicly documented limit (\`api-proxy.md\` §2 + axum \`DefaultBodyLimit\` docs), not a value derived from gateway internal code.

## Out of scope (P2 boundary defenses follow-ups)

- CORS preflight / origin allowlist
- Header injection / forwarding hygiene
- Streaming-mode body limits (chunked transfer)

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end test validating JSON request body size enforcement at a 2 MiB limit
  * Verifies HTTP 413 for oversized chat and embeddings requests and that such requests are not forwarded upstream
  * Confirms properly-sized requests receive HTTP 200 and are processed successfully

[Review Change Stack](https://app.coderabbit.ai/change-stack/api7/ai-gateway/pull/215)

### DIFF
--- a/tests/e2e/src/cases/body-limit-413-e2e.test.ts
+++ b/tests/e2e/src/cases/body-limit-413-e2e.test.ts
@@ -178,5 +178,34 @@ describe("body-limit 413 e2e: oversized JSON rejected, undersized accepted", () 
     );
     expect(undersizedRes.status).toBe(200);
     await undersizedRes.text();
+
+    // (3) Limit applies to /v1/embeddings too. axum's
+    // `DefaultBodyLimit` is per-extractor, not global, so a
+    // regression that wired the limit only on the chat handler
+    // (or used a different extractor for embeddings) would let
+    // an oversized embeddings request through. Pin the
+    // cross-endpoint uniformity with one parallel oversized
+    // probe — only assert the 413, not undersized parity (chat
+    // covers that already).
+    const oversizedEmbeddingsBody = JSON.stringify({
+      model: "body-limit-model",
+      input: "a".repeat(OVERSIZED_FILL_BYTES),
+    });
+    expect(
+      Buffer.byteLength(oversizedEmbeddingsBody, "utf8"),
+    ).toBeGreaterThan(LIMIT_BYTES);
+
+    const embeddingsBaseline = upstream.receivedRequests.length;
+    const embeddingsRes = await fetch(
+      `${app.proxyUrl}/v1/embeddings`,
+      {
+        method: "POST",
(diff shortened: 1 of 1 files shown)


moonming:
Closing this PR — the e2e it adds correctly surfaces a real product bug:

> The proxy does not enforce 413 on JSON bodies as documented. A 2.5 MiB body is accepted and forwarded to upstream, contradicting \`docs/api-proxy.md\` §2 which claims a 2 MiB axum default limit.

Filed as #221 with the full repro and proposed fix direction (wire \`request_body_limit_bytes\` per #193).

Holding the test back from main per project policy on test-surfaced product bugs. Same pattern as #173 / #202 / #220. Will land — unmodified — once #221 ships. Branch \`test/e2e-body-limit-413\` stays for reference.

PR #216, opened 2026-05-10
test(e2e): G15 — non-streaming tool_calls round-trip via OpenAI SDK

moonming:
## Summary

Closes part of the **G15** SDK-via-real-client gap from [issue #127](https://github.com/api7/ai-gateway/issues/127). Existing coverage:

- \`openai-sdk-compat.test.ts\` — chat (non-streaming + streaming content)
- \`embeddings-e2e.test.ts\` — embeddings via \`client.embeddings.create()\`
- PR #192 \`vision-messages-e2e.test.ts\` — vision via \`client.chat.completions.create({ messages: [...] })\`

Tool_calls via the OpenAI SDK was the last missing axis. This file pins the **non-streaming round-trip** end-to-end: caller's \`tools\` array reaches upstream verbatim, and upstream's \`message.tool_calls\` reaches the caller verbatim.

## Why non-streaming and not streaming

The streaming sibling (G2) is held back at [issue #202](https://github.com/api7/ai-gateway/issues/202) — \`ChatChunk\` (gateway-internal stream type) does not model \`tool_calls\`, so the streaming path silently drops them. The non-streaming path takes a **different** code route (no \`ChatChunk\` decode) and round-trips JSON byte-for-byte. This test is the sentinel that confirms #202 is streaming-specific and the non-streaming path stays clean.

## Test shape — single test, two halves

**Response side** (caller-observable):
- \`finish_reason === "tool_calls"\`
- \`message.tool_calls\` is a one-element array with byte-equal \`id\`, \`type\`, \`function.name\`
- \`function.arguments\` byte-equal to the canonical string AND parses to the canonical object. Catches re-serialization regressions (whitespace insertion, key reorder).

**Request side** (wire-shape blind-spot guard, per CLAUDE.md §8):
- Exactly one \`POST /v1/chat/completions\` with \`Bearer sk-mock\`
- Caller's \`tools\` array preserved in full — including the JSON Schema \`parameters\` block upstream needs verbatim
- \`tool_choice\` preserved (a regression that defaulted it to \`"none"\` would silently change upstream behavior)

## Source-blind compliance

Per CLAUDE.md §5: \`TOOL_CALL_ID\`, \`TOOL_NAME\`, \`CANONICAL_ARGS_STRING\`, the canonical parsed object, and the entire mock response body are author-written constants. No value derives from gateway internal source.

## Reference

- OpenAI Chat Completions API — <https://platform.openai.com/docs/api-reference/chat/create>
- OpenAI Node SDK \`ChatCompletionTool\` / \`ChatCompletionMessageToolCall\` types

### DIFF
--- a/tests/e2e/src/cases/tool-calls-non-streaming-e2e.test.ts
+++ b/tests/e2e/src/cases/tool-calls-non-streaming-e2e.test.ts
@@ -75,6 +75,26 @@ const TOOL_DEFINITION = {
   },
 };
 
+// Second tool — exercised on the request-side wire-shape gate so a
+// regression that kept only `tools[0]` (e.g. `tools.slice(0, 1)`)
+// is caught. The model isn't expected to invoke this one in the
+// canned response below, but it MUST reach upstream so the model
+// sees the full tool catalogue.
+const SECOND_TOOL_DEFINITION = {
+  type: "function" as const,
+  function: {
+    name: "get_time",
+    description: "Get the current time in a timezone",
+    parameters: {
+      type: "object",
+      properties: {
+        timezone: { type: "string" },
+      },
+      required: ["timezone"],
+    },
+  },
+};
+
 const NON_STREAM_BODY = {
   id: "chatcmpl-tool-nonstream-1",
   object: "chat.completion",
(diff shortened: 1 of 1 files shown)


moonming:
Closing this PR — the e2e it adds correctly surfaces a real product bug, broader than #202 had assumed:

> Non-streaming chat completions ALSO drop \`message.tool_calls\` end-to-end. \`crates/aisix-provider-openai/src/wire.rs:43\` is explicit that the gateway-internal types don't model \`tool_calls\` at all (\"doesn't model (tool_calls, refusal, audio, ...)\"). So both streaming AND non-streaming paths strip them.

Filed as #220 with the full repro and proposed fix direction. Cross-linked from #202.

Holding the test back from main per project policy on test-surfaced product bugs. Same pattern as #173 / #202 / #221. Will land — unmodified — once the gateway fix ships. Branch \`test/e2e-g15-async-and-embeddings-via-sdk\` stays for reference.

PR #238, opened 2026-05-11
fix(etcd): keep Watcher alive to prevent watch stream EOF

### LINKED ISSUE #237: etcd watch stream breaks immediately: Watcher dropped closes request channel
The DP's etcd watch stream fails with EOF immediately after establishment. Every `cycle()` does a successful `load_all()` (range GET works fine), then opens a watch stream that dies within ~120ms. The supervisor retries with exp backoff (1→60s) but it never recovers — runtime config updates created after DP boot are never delivered.

The root cause is in `crates/aisix-etcd/src/etcd_provider.rs:173`:

```rust
let (_watcher, stream) = client
    .watch(self.prefix.as_bytes(), Some(opts))
    .await
    .map_err(|e| ProviderError::Watch(format_error_chain(&e)))?;

Ok(Box::new(EtcdWatchStream { inner: stream }))
```

In `etcd-client` 0.14.1, `client.watch()` returns `(Watcher, WatchStream)` — the two halves of the bidirectional gRPC stream. `Watcher` holds the `Sender<WatchRequest>` (client→server direction). `WatchStream` holds the `Streaming<PbWatchResponse>` (server→client direction).

When `_watcher` is dropped at function return, the `Sender` is dropped, which closes the client→server half of the HTTP/2 stream. Kine's `Watch()` server (which loops on `ws.Recv()`) gets `io.EOF` and returns, tearing down the entire stream. The DP then sees EOF on the response stream.

The fix: keep the `Watcher` alive alongside the `WatchStream`. Something like:

```rust
pub struct EtcdWatchStream {
    inner: etcd_client::WatchStream,
    _watcher: etcd_client::Watcher,  // must live as long as inner
}
```

Reproduction:
1. Create resources via CP API
2. Start DP with `RUST_LOG=info,aisix_etcd=debug`
3. Create a new API key via CP API after DP is running
4. The new key never becomes usable — DP logs show repeated watch EOF failures

Impact: resources created/updated after DP start are never synced. The current QA test suite works around this by pre-creating all resources before DP boot.

jarvis9443:
The `watch()` method in `etcd_provider.rs` was dropping the `etcd_client::Watcher` immediately after opening the stream. The `Watcher` holds the `Sender<WatchRequest>` — the client→server half of the gRPC bidirectional stream. Dropping it closes that half, which causes kine's `Watch()` server to get EOF on `Recv()` and tear down the entire stream. The DP then sees EOF on the response stream and retries with backoff, but config updates created after DP boot are never delivered.

The fix stores the `Watcher` in the `EtcdWatchStream` struct so it lives as long as the stream.

Added an integration test (`watch_integration.rs`) that connects to a real etcd, opens a watch, puts a key, and asserts the event arrives — validating the stream stays alive after `watch()` returns.

Closes #237

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed etcd watch stream termination issue.

* **Tests**
  * Added integration test for etcd watch functionality.

* **Chores**
  * Updated CI workflow to support etcd testing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/api7/ai-gateway/pull/238)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,6 +66,8 @@ jobs:
       # Picked up by crates/aisix-admin/tests/etcd_integration.rs.
       # Same skip-if-unset pattern as the Redis case above.
       ADMIN_TEST_ETCD_URL: http://127.0.0.1:2379
+      # Picked up by crates/aisix-etcd/tests/watch_integration.rs.
+      ETCD_TEST_URL: http://127.0.0.1:2379
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
--- a/crates/aisix-etcd/src/etcd_provider.rs
+++ b/crates/aisix-etcd/src/etcd_provider.rs
@@ -170,12 +170,15 @@ impl ConfigProvider for EtcdConfigProvider {
         let opts = WatchOptions::new()
             .with_prefix()
             .with_start_revision(start_revision);
-        let (_watcher, stream) = client
+        let (watcher, stream) = client
             .watch(self.prefix.as_bytes(), Some(opts))
             .await
             .map_err(|e| ProviderError::Watch(format_error_chain(&e)))?;
 
-        Ok(Box::new(EtcdWatchStream { inner: stream }))
+        Ok(Box::new(EtcdWatchStream {
+            inner: stream,
+            _watcher: watcher,
+        }))
     }
 }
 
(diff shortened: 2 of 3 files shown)


jarvis9443:
The e2e failure in `body-edges-e2e.test.ts` (oversize body → EPIPE) is pre-existing — the latest main CI run ([25660248123](https://github.com/api7/ai-gateway/actions/runs/25660248123)) has the same failure. The `fetch` client gets EPIPE when the server closes the connection early on the 10.5 MiB body before the client finishes writing.

This is unrelated to the etcd watcher fix.

Issue #264, opened 2026-05-13
clarify health semantics vs routing cooldown and upstream target health


moonming:
## Summary

PR #263 adds same-target retries and optional `retry_on_429`, which surfaced an important semantics question: our current `/admin/v1/health` behavior does not match the health/circuit-breaker model used by Kong/APISIX/LiteLLM.

This issue documents the differences so we can decide whether to:
- keep the current behavior and document it clearly,
- align more closely with Kong/APISIX-style upstream health,
- or split the concepts into separate surfaces.

## Current AISIX behavior

Current code paths:
- `crates/aisix-proxy/src/chat.rs:781-800`
- `crates/aisix-proxy/src/health.rs:112-211`
- `crates/aisix-proxy/src/routing.rs:27-40`

Observed behavior today:
- health is tracked per `model.display_name`, not per concrete upstream target / endpoint / deployment
- health is observational only; it does not currently influence target selection or skip unhealthy targets during routing
- health state is a single consecutive-failure counter:
  - `Healthy` < 4 failures
  - `Degraded` >= 4 failures
  - `Down` >= 8 failures
  - any success resets the counter to 0
- failures are recorded in the routing retry path when `is_retryable(err, retry_on_429)` is true
- by default, `429` is **not** counted as retryable/unhealthy unless `retry_on_429=true`
- there is no active health check, no cooldown TTL, and no automatic decay besides a later success

### Kong

Sources:
- `kong/init.lua`
- `kong/runloop/balancer/healthcheckers.lua`
- `kong/db/schema/entities/upstreams.lua`
- <https://docs.konghq.com/gateway/latest/traffic-control/health-checks-circuit-breakers/>

Behavior:
- health is tracked per upstream target (`ip:port[:hostname]`), not per model alias
- passive health reporting happens on each failed upstream attempt during balancer retries
- unhealthy signals are typed (`http_failures`, `tcp_failures`, `timeouts`, `successes`) with independent thresholds
- default passive unhealthy statuses include `429, 500, 503`
- health directly affects routing: unhealthy targets are skipped by the balancer
- active health checks can later re-enable targets

### APISIX

Sources:
- `apisix/balancer.lua`
- `apisix/schema_def.lua`
- `apisix/healthcheck_manager.lua`

Behavior:
- health is tracked per upstream node / target
- retry-time failures are reported on each retry attempt
- health uses typed counters and thresholds, not one consecutive-failure counter
[...]

PR #268, opened 2026-05-13
feat(proxy, admin): runtime model status, cooldown, background health check, /admin/v1/models/status

moonming:
## Summary

Adds runtime model status tracking, request-path cooldown, background health probe (`background_model_check`), and a new read-only endpoint `GET /admin/v1/models/status`.

### Runtime status tracker (`ModelRuntimeStatusTracker`)
- Per-direct-model runtime state keyed by resolved direct-model `id`
- Three states: `healthy`, `cooldown` (request-path retryable failure), `unhealthy` (background check failure)
- Stale awareness: unhealthy entries auto-revert to healthy after `stale_after_seconds`

### Request-path cooldown
- Retryable bridge errors set a 30s cooldown on the failed target model
- Routing filter skips targets in cooldown/unhealthy during target selection
- If filtering would empty all candidates, cooldown/unhealthy is ignored (falls back to original order)

### Background model check
- New `background_model_check` config block on direct models only (rejected on routing models)
- Config: `enabled`, `interval_seconds`, `timeout_seconds`, `prompt`, `max_tokens`, `ignore_statuses`, `stale_after_seconds`
- Background checker sends a lightweight chat completion probe per configured model
- `ignore_statuses` (default [408, 429]) prevents transient failures from flipping unhealthy
- Non-ignored failures -> `unhealthy`; success -> clears unhealthy; ignored -> recorded as healthy with observability

### `/admin/v1/models/status`
- Returns runtime status for every model in the snapshot
- Direct models: live runtime state with stale awareness
- Routing models: always `not_applicable`
- Includes `last_checked_at`, `last_check_status`, `status_reason` when available

### Implementation notes
- Follows LiteLLM-style design (not full upstream health-check subsystem)
- No new external concepts (no `DeploymentKey`, `target`, `deployment`)
- Runtime status TTLs use `SystemTime` (not `tokio::time::Instant`) for cross-thread snapshot consistency
- Background checker runs as a periodic tokio task in `aisix-server/src/main.rs`

## Test coverage
- Rust unit tests for `HealthTracker`, `ModelRuntimeStatusTracker`, background check logic, routing filtering
- Rust tests for admin `GET /admin/v1/models/status` response shape
- Rust tests for `background_model_check` schema validation (direct accepted, routing rejected)
- E2E tests: `runtime-status-e2e.test.ts`, `background-health-e2e.test.ts`, `runtime-mixed-filtering-e2e.test.ts`, `retry-on-429-vs-background-ignore-e2e.test.ts`

## Summary by CodeRabbit

* **New Features**
  * Admin endpoint to list per-model runtime status (kind, status, last-check, reason).
  * Periodic background health checks for direct models with per-model config (interval, timeout, stale window, ignored statuses).
  * Per-model runtime health tracking, cooldowns, and routing policy to control behavior when all candidates are filtered.
  * Upstream Retry-After honored to influence cooldowns and 503 retry hints.

* **Tests**
  * Expanded unit and end-to-end tests for background checks, cooldowns, routing, and retry-after handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/api7/ai-gateway/pull/268)

### DIFF
--- a/crates/aisix-admin/src/lib.rs
+++ b/crates/aisix-admin/src/lib.rs
@@ -39,6 +39,7 @@ pub mod etcd_store;
 mod guardrails_handlers;
 mod health_handler;
 mod models_handlers;
+mod models_status_handler;
 mod observability_exporters_handlers;
 mod openapi;
 mod playground_handler;
@@ -72,6 +73,10 @@ pub fn build_router(state: AdminState) -> Router {
             get(models_handlers::get_model)
                 .put(models_handlers::update_model)
                 .delete(models_handlers::delete_model),
+        )
+        .route(
+            "/admin/v1/models/status",
+            get(models_status_handler::get_models_status),
         )
         .route(
             "/admin/v1/apikeys",
@@ -1088,5 +1093,91 @@ mod tests {
         assert_eq!(v["status"], "ok");
         // Empty snapshot → empty model list, but endpoint is operational.
         assert!(v["models"].is_array());
+    }
+
+    #[tokio::test]
+    async fn models_status_returns_direct_and_routing_rows() {
+        use aisix_core::resource::ResourceEntry;
+        use aisix_core::Model;
+        use aisix_proxy::ModelRuntimeStatusTracker;
+
+        let handle = SnapshotHandle::new(AisixSnapshot::new());
+        let store = InMemoryStore::new() as Arc<dyn ConfigStore>;
+        let runtime = Arc::new(ModelRuntimeStatusTracker::new());
+
+        let direct: Model = serde_json::from_value(model_payload("gpt4")).unwrap();
+        store
+            .put_model(ResourceEntry {
+                id: "direct-1".into(),
+                value: direct,
+                revision: 1,
+            })
+            .await
+            .unwrap();
+
+        let routing: Model = serde_json::from_value(json!({
+            "display_name": "router",
+            "routing": {
+                "targets": [{"model": "gpt4"}]
+            }
+        }))
+        .unwrap();
+        store
+            .put_model(ResourceEntry {
+                id: "routing-1".into(),
+                value: routing,
+                revision: 1,
+            })
+            .await
+            .unwrap();
+
+        runtime.record_ignored_check("direct-1", 429, "ignored_transient_error");
+
+        let state = AdminState::new(handle, store, &cfg()).with_runtime_status_tracker(runtime);
+        let app = build_router(state);
+        let resp = run(app, auth_req("GET", "/admin/v1/models/status", None)).await;
+        assert_eq!(resp.status(), StatusCode::OK);
+        let rows = body_json(resp).await;
+        let rows = rows.as_array().unwrap();
+        assert_eq!(rows.len(), 2);
+
+        let direct = rows.iter().find(|row| row["id"] == "direct-1").unwrap();
+        assert_eq!(direct["kind"], "direct");
+        assert_eq!(direct["status"], "healthy");
+        assert_eq!(direct["last_check_status"], 429);
+        assert_eq!(direct["status_reason"], "ignored_transient_error");
+
+        let routing = rows.iter().find(|row| row["id"] == "routing-1").unwrap();
+        assert_eq!(routing["kind"], "routing");
+        assert_eq!(routing["status"], "not_applicable");
+    }
+
+    #[tokio::test]
+    async fn create_model_accepts_background_model_check() {
+        let app = build_router(build_state());
+        let resp = run(
+            app,
+            auth_req(
+                "POST",
+                "/admin/v1/models",
+                Some(json!({
+                    "display_name": "bg-model",
+                    "provider": "openai",
+                    "model_name": "gpt-4o-mini",
+                    "provider_key_id": "11111111-1111-1111-1111-111111111111",
+                    "background_model_check": {
+                        "enabled": true,
+                        "interval_seconds": 1,
+                        "timeout_seconds": 10,
+                        "prompt": "Respond with OK",
+                        "max_tokens": 8,
+                        "ignore_statuses": [408, 429],
+                        "stale_after_seconds": 90
+                    }
+                })),
+            ),
+        )
+        .await;
+        assert_eq!(resp.status(), StatusCode::OK);
     }
 }
--- a/crates/aisix-admin/src/lib.rs
+++ b/crates/aisix-admin/src/lib.rs
@@ -1167,7 +1167,9 @@ mod tests {
                     "provider_key_id": "11111111-1111-1111-1111-111111111111",
                     "background_model_check": {
                         "enabled": true,
-                        "interval_seconds": 1,
+                        // Minimum interval is 5s in schema; using 30 to
+                        // mirror a realistic operator config.
+                        "interval_seconds": 30,
                         "timeout_seconds": 10,
                         "prompt": "Respond with OK",
                         "max_tokens": 8,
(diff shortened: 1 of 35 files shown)


moonming:
## Update: addressed audit findings from issue #264

Force-pushed [6b109ed](https://github.com/api7/ai-gateway/pull/268/commits/6b109edd) with the contract fixes the issue review flagged. Summary:

### HIGH
- **H1 cooldown trigger decoupled from `is_retryable`**: new `cooldown.trigger_statuses` config (default `[401, 408, 429, 500, 502, 503, 504]`). 401 / 408 now cool down even though they are non-retryable — same target would keep failing on the next request, so routing filter should skip it.
- **H2 `Retry-After` honored**: OpenAI and Anthropic bridges parse the seconds-form header into `BridgeError::UpstreamStatus.retry_after`. Cooldown layer uses it as TTL, clamped by `cooldown.max_seconds` (default 600s).
- **H3 fail-fast on all-filtered**: when every candidate is background-unhealthy, `filter_attempt_models` returns `AllUnhealthy` → dispatch returns `503 + Retry-After` (new `ProxyError::AllCandidatesUnavailable`). Legacy behavior opt-in via per-routing `on_all_filtered: original_order`.

### MEDIUM
- **M1 429 cools down regardless of `retry_on_429`** — retry (same request) and cooldown (next request) are independent layers.
- **M3 background-check safety**: 4-permit semaphore caps concurrent probes; schema `interval_seconds` floor raised to 5.

### New config surfaces
```yaml
# Direct model (all optional)
cooldown:
  enabled: true                    # default
  default_seconds: 30              # default
  max_seconds: 600                 # default
  honor_retry_after: true          # default
  trigger_statuses: [401, 408, 429, 500, 502, 503, 504]  # default
  trigger_on_timeout: true         # default
  trigger_on_transport: true       # default

# Routing model (optional)
on_all_filtered: fail              # fail | original_order; default fail
```

### Test coverage delta

**E2E (real backend + mock upstreams, no stubbed bridges):**
- 4 baseline files that were claimed in the original PR description but never committed are now committed and updated for the new contract
- 1 new file `cooldown-contract-e2e.test.ts` (5 tests) — H1 401, M1 429 without retry_on_429, H2 Retry-After, H3 fail, H3 escape hatch
- 2 pre-existing tests (`fallback-e2e`, `routing-strategies-e2e`) fixed to avoid warming cooldown in their readiness probes

**Rust unit coverage:** 22 new chat.rs tests (`decide_cooldown` + `filter_attempt_models`).

### Checks
- `cargo test --workspace`: 686 pass / 0 fail
- `cargo clippy --workspace --all-targets -- -D warnings`: clean
- `cargo fmt --all -- --check`: clean
- e2e: 92 pass / 0 fail across 47 files (run against real backend with etcd + mock upstreams)

moonming:
## Audit response — addressed findings from independent review

Force-pushed [2ea8a17](https://github.com/api7/ai-gateway/pull/268/commits/2ea8a17) with the fixes for the independent audit that ran against the prior commit.

### HIGH — both addressed

| Finding | Fix |
|---|---|
| **H-1** cooldown was chat-only; `/v1/messages`, `/v1/responses`, audio, `/v1/rerank` silently dropped the signal | Extracted `decide_cooldown` to `crate::cooldown` and wired it into all five dispatch sites (messages passthrough + cross-provider, responses, audio/speech, audio/transcribe, rerank) |
| **H-2** routing-strategies-e2e readiness was weakened to bare `waitConfigPropagation()` | Restored verification by gating on `admin.listModels()` finding the virtual record — proves propagation without dispatching traffic that would warm cooldown |

### MEDIUM — all addressed

| Finding | Fix |
|---|---|
| **M-1** H2 e2e relied on undocumented serde-default `SystemTime` shape | Added wire-shape assertion before parsing — future serialization change fails with a clear message instead of `NaN` |
| **M-2** `cooldown.default_seconds: 0` silently re-derived as 600s (opposite of operator intent) | Now treated as "disable cooldown on this model"; new unit test pins the contract |
| **M-5** `OnAllFilteredPolicy::Fail` rustdoc claimed Retry-After is "derived from cooldown expiry"; impl returns fixed 30s | Doc corrected to match implementation with rationale |

### Breaking-change call-outs (audit M-3 / M-4) — doc-only

For upgrade docs:
- **`BridgeError::UpstreamStatus` gained `retry_after: Option<Duration>`** — workspace-internal; all in-tree construction sites use `BridgeError::upstream_status(...)` or `upstream_status_with_retry_after(...)`. No on-the-wire effect (the type isn't serialized into etcd or HTTP responses).
- **`background_model_check.interval_seconds` schema minimum 1 → 5** — existing etcd payloads with `interval_seconds: 1` (the previous floor) will be rejected on DP snapshot reload. Operators that ran with `1` should bump to `>=5` before upgrading; the schema enforcement guards against burning provider quota on a tight loop.

### Drive-by fix

`tests/e2e/src/harness/admin.ts::listModels` was decoding the response as `{items: [...]}` but the admin endpoint returns a bare `ResourceEntry` array. The helper had no callers before this PR so the bug was latent — surfaced when H-2 fix started using it.

### Audit-flagged LOW deferred

- HTTP-date form of `Retry-After` not parsed (Anthropic rarely uses this) — tracked implicitly by the negative test, can be a follow-up issue if observed in production.
- `CooldownConfig::default()` returns all `None` but `enabled_or_default()` returns `true` — slightly counterintuitive when reading the struct cold. Doc improvement, not behavior.

### Checks
- `cargo test --workspace`: **689 pass / 0 fail** (+3 for the new `cooldown.rs` unit tests)
- `cargo clippy --workspace --all-targets -- -D warnings`: clean
- `cargo fmt --all -- --check`: clean
- e2e (real backend + mock upstreams): **92 pass / 0 fail across 47 files**

moonming:
## Audit round-2 response

Round-2 independent audit on commit [2ea8a17](https://github.com/api7/ai-gateway/pull/268/commits/2ea8a17) caught a real gap: the H-1 fix wired cooldown into the `!status.is_success()` branch but NOT into the `?`-escape paths on `.send().await` (transport) and `.json()` / `.bytes()` (decode). A hard upstream outage on `/v1/messages` would still leave the target healthy in routing — the exact failure mode H-1 was supposed to fix.

Force-pushed [7c4dfb2](https://github.com/api7/ai-gateway/pull/268/commits/7c4dfb2) with:

### HIGH — Transport/Decode error cooldown
- Added `crate::cooldown::note_failure(tracker, model_id, cfg, err)` helper that runs the cooldown decision, marks the tracker, and returns the `BridgeError` unchanged so call sites can keep using `?`
- Wired it into every `.map_err(BridgeError::Transport)` and `.map_err(BridgeError::UpstreamDecode)` across:
  - `messages.rs` (send + decode — 2 sites)
  - `responses.rs` (send + decode — 2 sites)
  - `rerank.rs` (send + decode — 2 sites)
  - `audio.rs` (multipart send/decode + speech send/decode — 4 sites)

### MEDIUM — Anthropic passthrough mark_healthy
- `messages.rs` Anthropic-passthrough success path now calls `state.runtime_status.mark_healthy(&model_entry.id)` so a recovered target exits `cooldown` cleanly on `/admin/v1/models/status`. The cross-provider sibling paths already did this; passthrough was the outlier.

### Audit LOW deferred
- Admin-snapshot readiness race window — practical gap is sub-100ms with 50ms `waitConfigPropagation` poll cadence; no observed flakes. If we start seeing them, gate on a DP-side health surface instead.

### New unit coverage
- `note_failure_marks_cooldown_for_transport_errors`
- `note_failure_marks_cooldown_for_decode_errors`
- `note_failure_no_op_when_cooldown_disabled`

### Checks
- `cargo test --workspace`: **692 pass / 0 fail**
- `cargo clippy --workspace --all-targets -- -D warnings`: clean
- `cargo fmt --all -- --check`: clean
- e2e: **92 pass / 0 fail across 47 files** (real backend + mock upstreams)

PR #298, opened 2026-05-16
feat(core): add provider/adapter/display_name fields to ProviderKey (Phase A skeleton)

moonming:
## Summary

Second sub-PR of [api7/AISIX-Cloud#302](https://github.com/api7/AISIX-Cloud/issues/302) Phase A (DP-side Provider→Adapter refactor). Builds on #297 (which landed the `Adapter` enum).

**This PR is intentionally zero-behavior-change** — pure type extension on `ProviderKey`. Nothing in the gateway reads the new fields yet; `Provider` continues to drive 100% of dispatch.

## What this PR does

1. Adds four new fields to `aisix_core::ProviderKey`:
   - `provider: String` — vendor identity (e.g. `"deepseek"`, `"openai"`). Free-form `String` in this PR; closed-set validation is deferred to a later Phase A sub-PR that wires dispatch.
   - `adapter: Option<Adapter>` — wire-shape, pinned to the closed `Adapter` enum from #297. `None` until a follow-up populates it.
   - `telemetry_tags: TelemetryTags` — attribution tags emitted alongside requests routed through this key.
2. Adds a new `TelemetryTags` struct with five optional fields:
   - `kind: Option<String>` — closed-set `"catalog" | "byo"` (enforced by the JSON Schema; the Rust type is `Option<String>` to stay forward-compatible if cp-api ships a new variant ahead of a DP rollout)
   - `featured: bool` — defaults to `false`
   - `branded_provider: Option<String>`
   - `pk_label: Option<String>`
   - `byo_label: Option<String>`

   `TelemetryTags` derives `Default` and is `#[serde(deny_unknown_fields)]` so an unknown tag from cp-api fails loudly on the DP rather than silently dropping.
3. Updates the `provider_key` JSON Schema with matching optional properties. `additionalProperties: false` is preserved both at the top level and inside `telemetry_tags`. `adapter` is constrained to the five `Adapter` enum values; `kind` is constrained to `"catalog" | "byo"`.
4. Re-exports `TelemetryTags` from `aisix_core::models` and the crate root, mirroring `Adapter`.

## Backward compatibility

All four new struct fields use `#[serde(default)]`. A pre-#302 payload like

```json
{"display_name":"openai-prod","secret":"sk-x","api_base":"https://api.openai.com/v1"}
```

[...]

PR #297, opened 2026-05-16
feat(core): add Adapter enum (skeleton for issue #302 Phase A)

moonming:
## Summary

First sub-PR of [api7/AISIX-Cloud#302](https://github.com/api7/AISIX-Cloud/issues/302) Phase A (DP-side Provider→Adapter refactor).

Adds a new closed `Adapter` enum on `aisix-core` alongside the existing `Provider` enum, plus a `From<Provider> for Adapter` mapping.

**This PR is intentionally zero-behavior-change** — it only introduces new types. Nothing in the gateway dispatches off `Adapter` yet, no entity field references it, and `Provider` continues to drive 100% of runtime behavior. Follow-up sub-PRs in Phase A migrate entities, schema, the Hub, and Bridges to consume `Adapter` directly.

## What this PR does

1. Adds `Adapter` enum in `crates/aisix-core/src/models/model.rs`:
   - Variants: `Openai`, `Anthropic`, `Bedrock`, `Vertex`, `AzureOpenai`
   - Same derive set as `Provider` (`Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize`)
   - `#[serde(rename_all = "kebab-case")]` so `AzureOpenai` serializes as `"azure-openai"`
2. Adds `From<Provider> for Adapter` mapping:
   - `Openai` → `Openai`
   - `Anthropic` → `Anthropic`
   - `Google` → `Vertex` (Vertex AI wire shape; the current `Provider::Google` was renamed from `Gemini` in #279)
   - `Deepseek` → `Openai` (OpenAI-compatible chat completions)
   - `Cohere` → `Openai` (gateway uses Cohere OpenAI-compat endpoints; #213 Phase 1, rerank-only)
   - `Jina` → `Openai` (rerank identity-mapped to OpenAI-compat shape; #213 Phase 2)
3. Re-exports `Adapter` from `aisix_core::models` and the crate root, mirroring `Provider`.

## What this PR does NOT do

- ❌ Does not remove or modify `Provider`
- ❌ Does not change `ProviderKey` struct
- ❌ Does not change `schema.rs` / JSON Schema
- ❌ Does not change Hub, Bridges, or any dispatch path
- ❌ Does not change any fixture or wire payload

## Serde casing note

[...]

PR #299, opened 2026-05-16
refactor(proxy): identify Cohere/Jina rerank by provider string (Phase A skeleton)

moonming:
## Summary

Per issue #302 Phase A: replace `Provider::Cohere` / `Provider::Jina` enum matching in the rerank dispatch path with string-based identification. This removes a dependency on enum variants slated for removal when `Provider` is collapsed into the closed `Adapter` set, without changing runtime behavior.

This is the **third** Phase A starter PR (after #297 add Adapter enum, #298 ProviderKey new fields):

- **#297** (merged) — added closed `Adapter` enum alongside existing `Provider`
- **#298** (open) — added `provider`/`adapter`/`display_name`/`telemetry_tags` fields to `ProviderKey` struct
- **this PR** — removes one of the last enum-variant consumers (rerank.rs) ahead of the eventual Provider→Adapter cutover

## Changes

- `crates/aisix-proxy/src/rerank.rs`:
  - `dispatch()`: hoist `provider_label` (lowercase string from `model.provider`) above the rerank gate; use string match instead of enum match for both the gate and the host default lookup.
  - `default_base_for_provider`: signature becomes `fn(&str) -> Option<String>`. The `_` arm replaces the previous `Anthropic`/`Google`/`Deepseek` arms that returned `None`.

## Zero behavior change

- Gate set is byte-identical: `{"openai", "cohere", "jina"}`
- Default hosts unchanged (`api.openai.com` / `api.cohere.com` / `api.jina.ai`)
- Metric/log labels keep emitting the same lowercase strings — Grafana dashboards keep working

## Test plan

- [x] `cargo test -p aisix-proxy` — 218 passed / 0 failed
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo fmt --all -- --check` — clean

## Out of scope

- Removing `Provider::Cohere` / `Provider::Jina` variants from the enum (Phase A cutover PR, later)
- Refactoring `aisix-server/src/main.rs` Cohere/Jina skip-registration logic (still intentional per existing comment)
- Whether rerank should eventually be folded into adapter sub-capability system (Open Question in #302)

Refs #302

## Summary by CodeRabbit

* **Refactor**
  * Improved /v1/rerank provider routing and gating to use a normalized provider label for OpenAI, Cohere, and Jina.
  * Updated logic for selecting the default upstream base URL, including fallback to the OpenAI host when no base is provided.

[Review Change Stack](https://app.coderabbit.ai/change-stack/api7/ai-gateway/pull/299?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/aisix-proxy/src/rerank.rs
+++ b/crates/aisix-proxy/src/rerank.rs
@@ -122,7 +122,7 @@ async fn dispatch(
     // unchanged.
     let provider_label = model
         .provider
-        .map(|p| format!("{p:?}").to_lowercase())
+        .map(|p| p.as_str().to_string())
         .unwrap_or_else(|| "unknown".to_string());
 
     // Per #168 + #213 Phases 1–2: `/v1/rerank` accepts OpenAI-,
--- a/crates/aisix-proxy/src/rerank.rs
+++ b/crates/aisix-proxy/src/rerank.rs
@@ -110,6 +110,21 @@ async fn dispatch(
 
     let model = &model_entry.value;
 
+    // Provider routing key, derived from `model.provider` as a
+    // lowercase string. Per #302 Phase A this dispatch path
+    // identifies Cohere/Jina by string rather than by `Provider`
+    // enum variant so that, when the `Provider` enum is later
+    // collapsed into the closed `Adapter` set, this file does not
+    // depend on variants (`Provider::Cohere`, `Provider::Jina`)
+    // that are slated for removal. The string values
+    // ("openai", "cohere", "jina") are the same labels emitted in
+    // metrics/access logs today, so dashboards keep working
+    // unchanged.
+    let provider_label = model
+        .provider
+        .map(|p| format!("{p:?}").to_lowercase())
+        .unwrap_or_else(|| "unknown".to_string());
+
     // Per #168 + #213 Phases 1–2: `/v1/rerank` accepts OpenAI-,
     // Cohere-, and Jina-shape upstreams. All three speak the same
     // body shape (`{model, query, documents, top_n, ...}`) at
@@ -125,11 +140,7 @@ async fn dispatch(
     // request and `data` (not `results`) on response, so it
     // requires a request/response adapter that's out of scope
     // for this phase. Tracked in the #213 follow-up.
-    use aisix_core::models::Provider;
-    let provider_allowed = matches!(
-        model.provider,
-        Some(Provider::Openai) | Some(Provider::Cohere) | Some(Provider::Jina)
-    );
+    let provider_allowed = matches!(provider_label.as_str(), "openai" | "cohere" | "jina");
     if !provider_allowed {
         return Err(ProxyError::InvalidRequest(format!(
             "model `{model_name}` is not an OpenAI, Cohere, or Jina provider; \
@@ -140,10 +151,6 @@ async fn dispatch(
     let pk_entry = crate::dispatch::resolve_provider_key(&snapshot, model)?;
     let api_key = crate::dispatch::require_secret(&pk_entry.value, model)?.to_string();
     let upstream_model = crate::dispatch::require_upstream_model(model)?.to_string();
-    let provider_label = model
-        .provider
-        .map(|p| format!("{p:?}").to_lowercase())
-        .unwrap_or_else(|| "unknown".to_string());
 
     // Rewrite model field.
     if let Some(m) = body.get_mut("model") {
@@ -157,17 +164,15 @@ async fn dispatch(
     //
     // The provider arm of `default_base_for_provider` is guaranteed to
     // return `Some` here because the gate above already rejected any
-    // `model.provider` outside `{Openai, Cohere, Jina}` — all three
+    // provider label outside `{"openai", "cohere", "jina"}` — all three
     // have explicit arms in the helper. The `unwrap_or_else` is
-    // defensive against a future enum variant that gets through the
+    // defensive against a future provider string that gets through the
     // gate without an arm in the helper; the audit-trail-friendly
     // default is OpenAI's host (it's a 4xx-from-OpenAI rather than
     // dispatching to a stale legacy domain).
     let base = match pk_entry.value.api_base.as_deref() {
         Some(b) if !b.trim().is_empty() => b.trim_end_matches('/').to_string(),
-        _ => model
-            .provider
-            .and_then(default_base_for_provider)
+        _ => default_base_for_provider(&provider_label)
             .unwrap_or_else(|| "https://api.openai.com".to_string()),
     };
     let url = crate::dispatch::build_v1_url(&base, "/rerank");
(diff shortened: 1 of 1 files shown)


moonming:

No shared context. Briefed cold with PR URL + issue #302 Phase A spec. Reviewed across all 6 required angles. Verified empirically (Debug-derive output on Rust 1.x, full test suite locally — 218 passed / 0 failed, callers of `default_base_for_provider` across the codebase).

---

## TL;DR

PR is mechanically correct and zero-behavior-change as advertised. The single substantive concern is at **MEDIUM** — using `format!("{p:?}").to_lowercase()` Debug-derive as a routing key, when `Provider::as_str()` already exists in the same file (`crates/aisix-core/src/models/model.rs:53-62`) and returns the exact same strings. The PR also pre-existed this pattern at 8 other call sites, so this is not a regression introduced here — but this PR elevates it from "metric label" to "load-bearing routing decision," which raises the bar. Recommendation: address MEDIUM-1, then merge.

---

## Findings

### MEDIUM-1 — `format!("{p:?}").to_lowercase()` as routing key when `Provider::as_str()` exists

**Where:** `crates/aisix-proxy/src/rerank.rs:123-126`

**Issue:** `Provider` (in `crates/aisix-core/src/models/model.rs:53-62`) already exposes a canonical lowercase wire string via `pub const fn as_str(self) -> &'static str`. That function is the *intentional* contract — it pins `Openai → "openai"`, `Cohere → "cohere"`, `Jina → "jina"`, etc. as an explicit match. By contrast, the PR's `format!("{p:?}").to_lowercase()` derives the same strings only by accident of: (a) the Debug derive emitting variant names as-is, (b) all current variant names being single CamelCase tokens that lowercase to the wire form.

Both invariants are silently fragile:
- A future variant like `AzureOpenai` Debug-stringifies to `"azureopenai"`, but its canonical wire form (per the sibling `Adapter` enum's `kebab-case` serde, lines 80-88) is `"azure-openai"`. The PR's expression would silently mis-route.
- A future contributor renaming a variant for Rust-side clarity (e.g. `OpenAi` → `OpenAI`) doesn't change the serde wire string, but DOES change Debug output → `"openai"` → no longer matches the gate.
- The Debug trait has no contract that says variant names are stable. Clippy lints (`clippy::print_stderr`-adjacent style lints) and Rust RFCs explicitly call out Debug as non-load-bearing for program logic.

The PR comment correctly notes that the string set `{"openai", "cohere", "jina"}` matches what's emitted in metrics today — but the existing metric-label call sites (chat.rs:634, completions.rs:128, etc.) treat the string as an opaque label, while this PR uses it as a **dispatch key** for both the accept gate and the host fallback. That's an elevation of trust.

**Suggested fix** — small change, no behavior delta:

```rust
let provider_label = model
    .provider
    .map(|p| p.as_str().to_string())
    .unwrap_or_else(|| "unknown".to_string());
```

`Provider::as_str()` returns `&'static str`, so the `.to_string()` is just the allocation needed to satisfy the existing `String` shape of `provider_label`. Zero semantic change today; future-proof against the failure modes above.

If preferred, leave the Debug derive in place and add a unit test that pins each `Provider` variant's lowercase Debug output to its `as_str()` — making the implicit invariant explicit, so a future variant rename surfaces as a test failure rather than silent mis-routing.

**Out-of-scope follow-up** (not blocking this PR): the 8 other Debug-derive call sites (`images.rs:143`, `audio.rs:305/424`, `chat.rs:634/736/832`, `messages.rs:458`, `completions.rs:128`, `embeddings.rs:177/188`) carry the same fragility for metric labels. They're lower-stakes than routing, but worth a follow-up PR to consolidate onto `Provider::as_str()`. Recommend filing as a tracking issue.

---

### LOW-1 — Comment claim about "guaranteed `Some`" is now load-bearing on string equality

**Where:** `crates/aisix-proxy/src/rerank.rs:164-172`

The comment block reads: _"The provider arm of `default_base_for_provider` is guaranteed to return `Some` here because the gate above already rejected any provider label outside `{"openai", "cohere", "jina"}` — all three have explicit arms in the helper."_

This is true today. The fragility is that "the gate" and "the helper" now agree on the same set of three lowercase strings derived from the same source. With the previous enum-match version, the Rust exhaustiveness checker caught any divergence at compile time. With the string version, divergence (e.g. someone adding `"voyage"` to one set but not the other) compiles cleanly and only fails at runtime via the OpenAI-host fallback. The fallback is reasonable defense in depth, but the loss of compile-time exhaustiveness is real.

**Suggested mitigation** — extract the accepted set as a constant referenced by both call sites:

```rust
/// Providers the gateway exposes via `/v1/rerank`. Single source of
/// truth for both the gate (in `dispatch`) and the host fallback
/// (in `default_base_for_provider`).
const RERANK_PROVIDERS: &[&str] = &["openai", "cohere", "jina"];
```

Then use `matches!(provider_label.as_str(), p if RERANK_PROVIDERS.contains(&p))` (or an inline match preserving readability) in `dispatch`, and have `default_base_for_provider` assert against the same list (e.g. via a debug_assertion). Not strictly required — current code's `_ => None` + OpenAI-host fallback is acceptable defense — but improves cross-site coupling.

---

### LOW-2 — `hoist provider_label` adds one allocation on the gate-rejection path

**Where:** `crates/aisix-proxy/src/rerank.rs:123` (was after gate, now before)

In the original code, `provider_label` was computed only on the success path. In the new code, it's computed for every rerank call, including those rejected by the gate. The cost is one `String` allocation + lowercase scan on a ≤10-byte ASCII string. Microsecond-scale, not in any tight loop, rerank not a hot path. Accept.

(Mentioning only because the PR description claims "zero behavior change" — strictly speaking there is a microscopic allocation-count delta on the reject path. The implication for performance is nil; the implication for correctness is nil.)

---

### LOW-3 — No new test pinning the `unwrap_or_else` OpenAI-host fallback after the gate

**Where:** `crates/aisix-proxy/src/rerank.rs:173-177`

The comment claims the `unwrap_or_else(|| "https://api.openai.com".to_string())` is "defensive against a future provider string that gets through the gate without an arm in the helper." Existing tests cover all three gate-accepted providers (`{openai, cohere, jina}`) and one rejected (`anthropic`). None exercise the defensive fallback — and indeed it's unreachable today by construction.

This is fine. The defensive layer is for *future* drift between the gate and the helper, which would be caught by the gate failing or the request hitting an unconfigured upstream. No test is necessary today, but if LOW-1 is taken, the shared constant + a `debug_assert!` makes the defense explicit and testable.

---

## Confirmations (no findings)

**Correctness ✓**
- `format!("{p:?}").to_lowercase()` empirically produces `"openai"`, `"anthropic"`, `"google"`, `"deepseek"`, `"cohere"`, `"jina"` for the current 6 variants. Verified by direct rustc test against an equivalent enum. The `#[serde(rename_all = "lowercase")]` attribute affects serde only, not Debug.
- Gate set `{"openai", "cohere", "jina"}` is byte-identical to the previous enum match `{Provider::Openai, Provider::Cohere, Provider::Jina}` per the verified Debug→lowercase mapping.
- `default_base_for_provider`'s host strings are byte-identical: `https://api.openai.com`, `https://api.cohere.com`, `https://api.jina.ai`. Cross-checked against `Provider::default_base_url()` (line 42-51 of `model.rs`) and the pinning test at line 605-611 — both agree.

**Reliability ✓**
- The exhaustive-enum-match → string-match swap removes one layer of compile-time safety (a new Provider variant no longer forces a compile error in this file). The MEDIUM and LOW findings above address this; the runtime fallback is acceptable defense.
- No new async/race surface introduced. No retry/timeout/cooldown behavior touched.

**Security ✓**
- No new attack surface. Provider label is derived from a server-side trusted enum value (`model.provider`), not user input. ASCII-only, no Unicode normalization concerns.
- No header changes, no auth changes, no logging changes.

**Sensitive-information leakage ✓**
- N/A. Same metric/log labels emitted as before.

**Breaking changes ✓**
- `default_base_for_provider` is `fn` (private, line 261), NOT `pub fn`. Signature change from `fn(Provider) -> Option<String>` to `fn(&str) -> Option<String>` is internal. No public API impact, no on-disk format change, no wire protocol change.
- The `use aisix_core::models::Provider` inside `dispatch` (was at line 128 of the original file) is removed; this is dead-code cleanup that the change made possible, consistent with CLAUDE.md §3.

**E2E test coverage ✓**
- 218 tests pass locally on this branch (verified). The 4 rerank.rs tests collectively pin: Anthropic → 400 reject (gate works), Cohere → 200 dispatch (gate accepts + Cohere host), Jina → 200 dispatch (gate accepts + Jina host), OpenAI → 200 dispatch (happy path). All four paths exercise `provider_label` derivation, the gate, and `default_base_for_provider`'s real callsite.
- No new test is necessary for a zero-behavior-change refactor — the existing tests' assertions would catch any regression in the gate-set, host-set, or label-set. The cohere/jina tests already do strict body-key-set assertions that pin upstream routing precisely.

---

## Verdict

**Recommend: address MEDIUM-1, then merge.**

MEDIUM-1 is a one-line code change (`.map(|p| p.as_str().to_string())` instead of `.map(|p| format!("{p:?}").to_lowercase())`) that removes a future-fragility without changing today's behavior. LOW-1, LOW-2, LOW-3 are noted but non-blocking; LOW-1 in particular is worth a follow-up issue for the broader 8-call-site Debug-derive cleanup.

The PR description's "zero behavior change" claim holds. The refactor is in line with issue #302 Phase A — removing Cohere/Jina enum-variant dependencies ahead of the Provider→Adapter split.

🤖 Generated by independent audit agent (general-purpose, no shared context).

moonming:
## MEDIUM-1 addressed in 15d3461

Per the [independent audit](https://github.com/api7/ai-gateway/pull/299#issuecomment-4466651648) MEDIUM-1: replaced `format!("{p:?}").to_lowercase()` with `Provider::as_str()` ([model.rs:53-62](https://github.com/api7/ai-gateway/blob/main/crates/aisix-core/src/models/model.rs#L53-L62)). Same strings produced today, robust against future variant names that don't lowercase cleanly (e.g. the sibling `Adapter::AzureOpenai` → `"azure-openai"` via explicit `kebab-case` rename).

LOW-1 (`RERANK_PROVIDERS` const extraction) + LOW-2 (hoisted allocation on reject path) + LOW-3 (no test for unreachable defensive fallback) are intentionally deferred — they don't gate this skeleton PR and the audit explicitly flagged them as acceptable for follow-up. The other 8 Debug-derive call sites elsewhere in the codebase (chat.rs / completions.rs / embeddings.rs / audio.rs / messages.rs / images.rs) are out of scope for this PR; will file a separate tracking issue.

Tests still green, clippy clean.

PR #300, opened 2026-05-16
feat(gateway): add two-tier Hub dispatch (Phase A skeleton)

moonming:
## Summary

Skeleton sub-PR for issue #302 Phase A. Adds the two forward-looking tier maps to `Hub` (`crates/aisix-gateway/src/hub.rs`) so future Phase A / Phase D sub-PRs can wire bridges through them without re-touching this file.

- `family_bridges: DashMap<Adapter, Arc<dyn Bridge>>` — wire-shape default bridges, keyed on the `Adapter` enum (added in #297).
- `specialized_bridges: DashMap<String, Arc<dyn Bridge>>` — vendor overrides, keyed on `ProviderKey.provider` (added in #298).
- `register_family` / `register_specialized` accessors.
- `dispatch_two_tier(&ProviderKey) -> Option<Arc<dyn Bridge>>` — specialized lookup first, then adapter-family fallback. Returns `Option` so the caller picks the missing-bridge error semantics; this layer stays panic-free.

## Zero behavior change

- Existing `bridges: DashMap<Provider, _>` and `register` / `get(Provider)` are untouched. `aisix-proxy`, rerank (#299), and `build_hub()` keep using them.
- `build_hub()` is intentionally **not modified** — the new maps stay empty until follow-up PRs register the per-family / per-vendor bridges as part of the Phase D cutover.
- No proxy or rerank path consumes the new methods yet (verified by `cargo test -p aisix-proxy --lib`: 218 passed).

## Context

Phase A skeleton trail so far:

- #297 — `Adapter` enum (5-value closed enum).
- #298 — `ProviderKey` gained `provider: String`, `adapter: Option<Adapter>`, `display_name`, `telemetry_tags` (all `#[serde(default)]`, backward-compatible).
- #299 — rerank.rs switched to `Provider::as_str()`.
- **This PR** — Hub gains the two-tier dispatch surface, still un-consumed.

Per issue #302 §3, follow-up PRs (Phase D cutover) will:

1. Populate `ProviderKey.provider` / `.adapter` in entity payloads and the bridge factory wiring.
2. Register family + specialized bridges in `build_hub()`.
3. Flip the proxy dispatch path from `Hub::get(Provider)` to `Hub::dispatch_two_tier(&ProviderKey)`.

## Test plan

- [x] `cargo test -p aisix-gateway` — 54 passed (8 new tests cover: family hit, specialized override, double-miss, adapter-None paths, overwrite semantics, legacy `Provider`-keyed registry co-existence).
- [x] `cargo test -p aisix-proxy --lib` — 218 passed (downstream `Hub.get(Provider)` consumer unaffected).
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean.
- [x] `cargo fmt --all -- --check` — clean.

Closes none. Linked to: #302.

## Release Notes

* **New Features**
  * Implemented a multi-tier dispatch system supporting adapter and vendor-specific bridge registrations with automatic fallback resolution
  * Enhanced request resolution with improved configuration management across different provider types

[Review Change Stack](https://app.coderabbit.ai/change-stack/api7/ai-gateway/pull/300?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/aisix-gateway/src/hub.rs
+++ b/crates/aisix-gateway/src/hub.rs
@@ -8,8 +8,29 @@
 //! There is no fallback logic here — that is the proxy layer's job and
 //! lands in its own PR. The Hub exists purely to resolve Provider →
 //! Bridge cheaply and consistently.
+//!
+//! # Two-tier dispatch skeleton (issue #302 Phase A)
+//!
+//! In addition to the existing `Provider`-keyed registry, the Hub now
+//! also carries two forward-looking maps keyed by [`ProviderKey`]
+//! fields introduced in Phase A:
+//!
+//! - `family_bridges` — keyed on `Adapter` (wire shape: `openai`,
+//!   `anthropic`, `bedrock`, `vertex`, `azure-openai`). The default
+//!   bridge for any provider that matches that wire shape.
+//! - `specialized_bridges` — keyed on `ProviderKey.provider` (vendor
+//!   string, e.g. `"deepseek"`, `"jina"`). Used when a specific vendor
+//!   needs handling that diverges from its wire-shape default.
+//!
+//! [`Hub::dispatch_two_tier`] looks up specialized first, then falls
+//! back to the family bridge. It is **not consumed by any caller** in
+//! this PR — `build_hub()` does not register family/specialized
+//! bridges and the proxy path continues to dispatch through the
+//! existing `Provider`-keyed `get()`. The new methods exist so future
+//! Phase A / Phase D sub-PRs can wire dispatch through them without
+//! re-touching this file.
 
-use aisix_core::models::Provider;
+use aisix_core::models::{Adapter, Provider, ProviderKey};
 use dashmap::DashMap;
 use std::sync::Arc;
 
@@ -23,6 +44,16 @@ use crate::bridge::Bridge;
 #[derive(Default)]
 pub struct Hub {
     bridges: DashMap<Provider, Arc<dyn Bridge>>,
+    /// Wire-shape default bridges. Keyed on [`Adapter`]. Phase A
+    /// skeleton — empty until follow-up PRs register the per-family
+    /// default bridge. See module docs.
+    family_bridges: DashMap<Adapter, Arc<dyn Bridge>>,
+    /// Vendor-specific override bridges. Keyed on
+    /// `ProviderKey.provider` (vendor string). Phase A skeleton —
+    /// empty until follow-up PRs register specialized handlers for
+    /// vendors that diverge from their wire-shape default. See module
+    /// docs.
+    specialized_bridges: DashMap<String, Arc<dyn Bridge>>,
 }
 
 impl Hub {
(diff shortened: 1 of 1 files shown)


moonming:

Cold-read review with no shared context. Branch checked out, tests + clippy re-run locally, downstream callers grepped.

---

## Verdict

**PR #300 passes independent audit — recommend merge.**

No HIGH or MEDIUM findings. Three LOW notes below, none of them blocking.

---

## Verified claims

- Single file, +213/-1 (`crates/aisix-gateway/src/hub.rs`).
- `build_hub()` unchanged — the two new maps stay empty in production today, and `dispatch_two_tier` returns `None` until follow-up Phase D PRs register family/specialized bridges.
- Legacy `bridges: DashMap<Provider, _>` plus `register` / `get(Provider)` / `providers` / `len` / `is_empty` are byte-identical to `main`.
- Downstream callers (`crates/aisix-proxy/src/chat.rs:500,525,817`) still go through `state.hub.get(provider)` — none of them touch the new methods. Rerank (`crates/aisix-proxy/src/rerank.rs`) does not use the Hub at all (independent dispatch path).
- No struct-literal `Hub { ... }` initializations outside this file — every caller uses `Hub::new()` / `Hub::default()`, so adding private `DashMap` fields is non-breaking. `DashMap: Default` keeps `#[derive(Default)]` working.
- No `format!("{:?}", hub)` / `{:?}.*hub` assertions in the tree, so extending the `Debug` impl with `families` + `specialized` is safe.
- Local verification:
  - `cargo test -p aisix-gateway --lib` → **54 passed** (8 new `hub::tests::*two_tier*` + `*family*` + `*specialized*` + `legacy_provider_registry_is_unaffected_by_two_tier_maps`).
  - `cargo test -p aisix-proxy --lib` → **218 passed**.
  - `cargo clippy -p aisix-gateway --lib --tests -- -D warnings` → clean.

## Deviation from issue #302 §3 snippet

The issue sketches:

```rust
pub fn dispatch(&self, pk: &ProviderKey) -> Arc<dyn Bridge> {
    if let Some(b) = self.specialized.get(&pk.provider) { return b.clone(); }
    self.family.get(&pk.adapter).expect("family bridge missing").clone()
}
```

This PR ships:

```rust
pub fn dispatch_two_tier(&self, pk: &ProviderKey) -> Option<Arc<dyn Bridge>> {
    if let Some(b) = self.specialized_bridges.get(&pk.provider) { return Some(b.clone()); }
    let adapter = pk.adapter?;
    self.family_bridges.get(&adapter).map(|r| r.clone())
}
```

Three deltas, all defensible:

1. **`Option` return vs `.expect`** — PR is panic-free at the dispatch layer and lets the proxy decide envelope shape (likely 404 / 502 / 503). This is a strict improvement over a panic that would crash the worker.
2. **Method name `dispatch_two_tier`** vs `dispatch` — explicit, avoids ambiguity while the legacy `get(Provider)` path still exists. Phase D rename to `dispatch` is trivial once `get(Provider)` is removed.
3. **`pk.adapter?` early-return** — `ProviderKey.adapter` is `Option<Adapter>` per #298 (legacy/BYO payloads can omit it), so the snippet's `&pk.adapter` would not have compiled against the real type. This PR handles the `None` arm correctly.

Verdict: deviations are all in the direction of more correct, more defensive code.

## Per-angle review

### Correctness
- `dispatch_two_tier` precedence (specialized first, family fallback) matches issue #302 §3 design intent. `register_specialized_and_dispatch_overrides_family` covers the priority path; `dispatch_two_tier_specialized_hits_even_when_adapter_missing` covers the BYO/legacy adapter=None case.
- `Option` return — callers will need to convert `None` → user-visible error. That wiring is deferred to Phase D and is called out in the PR description; not a regression today since no caller exists.
- `pk.adapter: None` + specialized miss returning `None` is the correct legacy/un-migrated path (`dispatch_two_tier_returns_none_when_adapter_missing_and_no_specialized` covers it).

### Reliability
- `register_family` / `register_specialized` use `DashMap::insert` — overwrites match the live-reconfigure semantics of the existing `register`, documented in doc comments.
- `DashMap` is `Send + Sync`; reads during concurrent writes are well-defined (one of the values, never torn). `dispatch_two_tier` clones the `Arc<dyn Bridge>` while holding the per-shard read guard, so a concurrent `register_*` cannot drop the bridge during dispatch. No race concern.

### Security
N/A — pure in-memory type + maps, no I/O, no auth, no parsing of attacker input.

### Sensitive-info leakage
- `Debug for Hub` now includes `families: Vec<Adapter>` (5-value closed enum, not sensitive) and `specialized: Vec<String>` (vendor strings — `"deepseek"`, `"jina"`, etc., from models.dev). No secret material. **No current log site `{:?}`s the Hub**, so this is theoretical. See LOW-2 below for a forward-looking note.

### Breaking changes
- Public API: only adds (`register_family`, `register_specialized`, `dispatch_two_tier`). Pre-existing surface byte-identical.
- On-disk format: unchanged — this PR doesn't touch serde.
- Default behavior: unchanged — `build_hub()` doesn't populate the new maps and no caller invokes them.

### E2E test coverage
This is a skeleton PR with no caller wiring, so unit-level coverage of `Hub` is the appropriate granularity — there is no user-visible contract yet to e2e-cover. Phase D PRs that flip dispatch from `Hub::get(Provider)` to `Hub::dispatch_two_tier(&ProviderKey)` will need full e2e (request body shape, header forwarding, error envelope on missing bridge). Flagging this now as a Phase-D blocker, not a Phase-A blocker.

## LOW findings (non-blocking)

### LOW-1 — `dispatch_two_tier` is dead code today
By design — `build_hub()` doesn't register family/specialized bridges and no proxy path calls `dispatch_two_tier`. The PR description is explicit about this. Worth a sanity check in CI before Phase D merges: if `family_bridges` is still empty at startup after Phase D wiring lands, `dispatch_two_tier` will silently return `None` for every non-specialized request. Suggest adding a debug-assert (in Phase D, not here):

```rust
// In Phase D's build_hub() after registration:
debug_assert!(!hub.family_bridges_is_empty(), "Phase D wiring must register family bridges");
```

(Would need a tiny `pub(crate) fn family_bridges_is_empty(&self) -> bool` accessor — not part of this PR.)

### LOW-2 — Debug exposes vendor strings if Hub is ever `{:?}`'d
Currently no log site formats the Hub with `{:?}`, so this is latent. If a future operator BYOs a vendor with a name that doubles as PII (`"customer-acme-llm"`), and someone adds a `tracing::debug!(?hub, ...)`, the BYO label would land in the log. Not actionable today; flag for review when Phase A wires telemetry attribution (see #302 telemetry_tags).

### LOW-3 — Defense-in-depth: empty `pk.provider` could collide with empty registration
`dispatch_two_tier` does `self.specialized_bridges.get(&pk.provider)` unconditionally. If a Phase D PR ever registered `""` as a specialized key (an obvious mistake but possible during refactors), legacy ProviderKeys with `provider: ""` (`#[serde(default)]` per #298) would silently route to it. A `debug_assert!(!provider.is_empty())` in `register_specialized` would catch that at test time:

```rust
pub fn register_specialized(&self, provider: impl Into<String>, bridge: Arc<dyn Bridge>) {
    let provider = provider.into();
    debug_assert!(!provider.is_empty(), "specialized bridge key must be non-empty");
    self.specialized_bridges.insert(provider, bridge);
}
```

Not required for this PR — file as a follow-up if you agree.

---

## Merge gate

All HIGH/MEDIUM: **none**. All LOW findings are forward-looking or stylistic, none block merge.

PR #300 passes independent audit per CLAUDE.md §8 — recommend merge.

PR #301, opened 2026-05-16
feat(provider-openai): add overrides module for RuntimeConfig request/response (Phase A skeleton)

moonming:
## Summary

Sub-PR of [api7/AISIX-Cloud#302](https://github.com/api7/AISIX-Cloud/issues/302) Phase A. Ships the **primitive request/response transforms** referenced by issue v10 §5 RuntimeConfig:

```json
"request": {
  "param_renames":      { "max_completion_tokens": "max_tokens" },
  "param_constraints":  { "temperature_max": 1.0 },
  "default_headers":    { "X-Foo": "bar" },
  "default_body_fields":{ "safe_prompt": true }
},
"response": {
  "stream_done_marker":     "required",
  "content_list_to_string": false,
  "error_envelope":         "openai",
  "reasoning_field":        "delta.reasoning_content"
}
```

**Zero behavior change.** Nothing in `OpenAiBridge::chat()` /
`chat_stream()` consumes these functions yet; the module sits in
`aisix-provider-openai::overrides` as standalone utilities. Phase D
wires them into the Bridge dispatch path when the new contract
cuts over.

## Why primitive parameter types, not `RequestOverrides` / `ResponseOverrides`

PR #298 (Phase A2, merged) landed `provider` / `adapter` /
`telemetry_tags` on `ProviderKey` but did **not** add the `request`
or `response` blocks. Those struct types don't exist in the
codebase yet. Rather than invent them here (and force a second PR
to rename when Phase A's later sub-PRs land them), each function
takes the smallest primitive that lets callers pass field values
straight through:

| Function | Parameter |
|---|---|
| `apply_param_renames` | `&HashMap<String, String>` |
| `apply_param_constraints` | `&Constraints` (module-local, temperature only) |
| `apply_default_headers` | `&HashMap<String, String>` |
| `apply_default_body_fields` | `&serde_json::Map<String, Value>` |
| `apply_stream_done_marker_policy` | `StreamDoneMarker` enum + `bool` |
| `apply_content_list_to_string` | `&mut serde_json::Value` (request body) |
| `extract_reasoning_field` | `&mut Value`, `&str` path |

When the struct types do land, the future Bridge wire-up passes
the field values straight in. The override behavior is decoupled
from the on-disk schema shape.

## What the module does NOT do

- ❌ Does not touch `OpenAiBridge::chat()` / `chat_stream()` / any
  dispatch path
- ❌ Does not modify `RuntimeConfig` types (none of the override
  struct types exist yet — see above)
- ❌ Does not touch `Hub`, `lib`, `wire`, or `bridge` files beyond
  one `pub mod overrides;` line in `lib.rs`
- ❌ Does not create a new crate

## Reference implementations consulted

- LiteLLM `convert_content_list_to_str` —
  [common_utils.py](https://github.com/BerriAI/litellm/blob/main/litellm/litellm_core_utils/prompt_templates/common_utils.py).
  Reference behavior: concatenate text fields of every `type:"text"`
  block with no separator, drop non-text blocks. Tests pin this
  contract.
- LiteLLM `convert_content_list_to_string` flag —
  [openai_like/dynamic_config.py](https://github.com/BerriAI/litellm/blob/main/litellm/llms/openai_like/dynamic_config.py).
  Confirms this is per-provider opt-in, not a default transform.
- DeepSeek `delta.reasoning_content` shape —
  [reasoning_model docs](https://api-docs.deepseek.com/guides/reasoning_model).
  Source of the `extract_reasoning_field` canonical-slot choice.
- OpenAI SSE `data: [DONE]` terminator —
  [chat/streaming](https://platform.openai.com/docs/api-reference/chat/streaming).
  Source of the `Required` policy's wire definition.

## Test plan

- [x] `cargo test -p aisix-provider-openai --lib` — 62 passed (+35 over main)
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [x] `cargo fmt --all -- --check` — clean
- [x] `apply_param_renames`: key-present, key-absent, value-type-preserved, multiple renames, existing-target wins, no-op self-rename, non-object body
- [x] `apply_param_constraints`: above max, within range, below min, missing field, non-number value, empty constraints
- [x] `apply_default_headers`: missing header added, existing header preserved (case-insensitive), unparseable name skipped
- [x] `apply_default_body_fields`: missing added, existing preserved, non-object no-op
- [x] `apply_stream_done_marker_policy`: required+seen, required+missing, optional (both), none (both)
- [x] `apply_content_list_to_string`: two-text-block flatten, string content untouched, image-only preserved (don't drop), text + non-text mixed, no-messages-field no-op
- [x] `extract_reasoning_field`: canonical-path no-op, nested-path lift, missing source no-op, empty-string source no-op, non-delta root rejected, multiple choices

## References

- Tracking issue: api7/AISIX-Cloud#302
- Predecessors: #297 (Adapter enum), #298 (Phase A2 ProviderKey fields), #299 (Phase A rerank refactor)

## Summary by CodeRabbit

* **New Features**
  * Added request and response override utilities for improved provider compatibility, including parameter transformation, temperature constraints, header and body field defaults, and content normalization.

* **Chores**
  * Added `http` dependency to the OpenAI provider crate.

[Review Change Stack](https://app.coderabbit.ai/change-stack/api7/ai-gateway/pull/301?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -253,6 +253,7 @@ dependencies = [
  "async-trait",
  "bytes",
  "futures",
+ "http 1.4.0",
  "reqwest",
  "serde",
  "serde_json",
--- a/crates/aisix-provider-openai/Cargo.toml
+++ b/crates/aisix-provider-openai/Cargo.toml
@@ -21,6 +21,7 @@ tokio.workspace = true
 futures.workspace = true
 async-stream = "0.3"
 bytes.workspace = true
+http.workspace = true
 
 [dev-dependencies]
 tokio = { workspace = true, features = ["macros", "rt", "time"] }
--- a/crates/aisix-provider-openai/src/lib.rs
+++ b/crates/aisix-provider-openai/src/lib.rs
@@ -12,6 +12,7 @@
 #![deny(rust_2018_idioms)]
 
 mod bridge;
+pub mod overrides;
 mod wire;
 
 pub use bridge::{OpenAiBridge, OPENAI_DEFAULT_BASE};
--- a/crates/aisix-provider-openai/src/overrides.rs
+++ b/crates/aisix-provider-openai/src/overrides.rs
@@ -90,8 +90,15 @@ pub enum StreamDoneOutcome {
 ///
 /// Rewrites every present top-level key whose name matches a key in
 /// `renames` to the renames-map's value. Absent source keys are
-/// no-ops. Conflicts (the target key already exists) prefer the
-/// already-present value — the rename is *non-destructive*.
+/// no-ops.
+///
+/// **Collision semantics: source wins.** When both source and target
+/// are present in the body, the source value overwrites the target.
+/// Matches LiteLLM's convention across `databricks` / `openai_like` /
+/// `sambanova` / `dynamic_config` transformations — for the canonical
+/// `max_completion_tokens → max_tokens` case the newer (source) value
+/// is what the caller intended; an accidentally-leftover deprecated
+/// `max_tokens` value should not shadow it.
 ///
 /// Non-object bodies are no-ops by construction; the function does
 /// not panic on `null` / array / scalar inputs.
@@ -106,11 +113,8 @@ pub fn apply_param_renames(body: &mut Value, renames: &HashMap<String, String>) 
         let Some(value) = obj.remove(from) else {
             continue;
         };
-        // Non-destructive: if the target already has a value, the
-        // existing entry wins and we drop the renamed one. cp-api
-        // shouldn't ship a rename that collides with an explicit
-        // value the caller already set.
-        obj.entry(to.clone()).or_insert(value);
+        // Source wins: overwrite any pre-existing target value.
+        obj.insert(to.clone(), value);
     }
 }
 
@@ -151,6 +155,17 @@ pub fn apply_param_constraints(body: &mut Value, constraints: &Constraints) {
     }
 }
 
+/// Authentication-related headers that `apply_default_headers` will
+/// never set, even if cp-api validation slips and allows them through.
+/// Defense-in-depth: a misconfigured `default_headers` block must never
+/// override the auth header the OpenAiBridge sets itself. cp-api SHOULD
+/// reject these at write time (issue #302 §5 validation rules), but
+/// the DP enforces it again at apply time.
+///
+/// `HeaderName` is case-insensitive so the lowercase form is the
+/// canonical comparison key.
+const RESERVED_DEFAULT_HEADERS: &[&str] = &["authorization", "x-api-key", "x-goog-api-key"];
+
 /// Apply `request.default_headers` to an outbound `HeaderMap`.
 ///
 /// Headers already present on `headers` (case-insensitive, since
@@ -160,11 +175,19 @@ pub fn apply_param_constraints(body: &mut Value, constraints: &Constraints) {
 /// silently skipped; the default block came from cp-api validation
 /// and any unparseable entry is a config error one layer up, not a
 /// runtime failure the dispatch should hard-fail on.
+///
+/// **Auth-header guard:** keys in [`RESERVED_DEFAULT_HEADERS`] are
+/// dropped before insertion as defense-in-depth — a misconfigured
+/// default_headers block must never inject `Authorization` or vendor
+/// API-key headers that would override the Bridge's own auth.
 pub fn apply_default_headers(headers: &mut HeaderMap, defaults: &HashMap<String, String>) {
     for (name, value) in defaults {
         let Ok(parsed_name) = name.parse::<HeaderName>() else {
             continue;
         };
+        if RESERVED_DEFAULT_HEADERS.contains(&parsed_name.as_str()) {
+            continue;
+        }
         if headers.contains_key(&parsed_name) {
             continue;
         }
@@ -406,16 +429,19 @@ mod tests {
     }
 
     #[test]
-    fn rename_with_existing_target_keeps_existing() {
-        // Caller already supplied `max_tokens`; the rename from
-        // `max_completion_tokens` must not overwrite it.
+    fn rename_source_wins_when_both_present() {
+        // Both source and target keys are present in the body. Per
+        // LiteLLM convention (source wins), the source's value should
+        // overwrite the target's pre-existing value — for the canonical
+        // max_completion_tokens → max_tokens case, the newer source
+        // value 100 wins over the deprecated target value 50.
         let mut body = json!({ "max_completion_tokens": 100, "max_tokens": 50 });
         let renames = HashMap::from([(
             "max_completion_tokens".to_string(),
             "max_tokens".to_string(),
         )]);
         apply_param_renames(&mut body, &renames);
-        assert_eq!(body, json!({ "max_tokens": 50 }));
+        assert_eq!(body, json!({ "max_tokens": 100 }));
     }
 
     #[test]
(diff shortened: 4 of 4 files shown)


moonming:

Cold review of PR #301 (no shared context with the implementing agent). Scope: the seven `apply_*` / `extract_*` primitives in `crates/aisix-provider-openai/src/overrides.rs`. LiteLLM cross-checks done against local `BerriAI/litellm` checkout (`litellm/litellm_core_utils/prompt_templates/common_utils.py`, `llms/openai_like/dynamic_config.py`, `llms/openrouter/chat/transformation.py`, `llms/ovhcloud/chat/transformation.py`, `llms/groq/chat/transformation.py`, `llms/openai_like/chat/transformation.py`, `llms/databricks/chat/transformation.py`). DeepSeek wire confirmed against the reasoning_model docs.

`cargo test -p aisix-provider-openai --lib` — 62 passed locally. `cargo clippy -p aisix-provider-openai --lib -- -D warnings` — clean.

## Verdict

**PASS with two MEDIUM concerns and minor LOW gaps.** Either address the MEDIUMs with code changes or explicitly justify them in the PR thread (per §8 merge gate), then this is good to merge.

---

## MEDIUM-1 — `apply_param_renames` collision semantics diverge from every LiteLLM precedent

**Finding.** PR #301 (`overrides.rs:113`) makes the collision behavior *target wins*:

```rust
obj.entry(to.clone()).or_insert(value);
```

If both `from` and `to` already exist in the body, the renamed value is silently dropped and the pre-existing target is kept.

LiteLLM, in every place I checked, does the opposite — **source wins / overwrite**:

- `llms/databricks/chat/transformation.py:302-305` — `mapped_params["max_tokens"] = non_default_params["max_completion_tokens"]; mapped_params.pop("max_completion_tokens", None)`
- `llms/openai_like/chat/transformation.py:174-177` — identical pattern, target overwritten
- `llms/openai_like/dynamic_config.py:141` — `optional_params[provider.param_mappings[param]] = value` (unconditional overwrite)
- `llms/sambanova/chat.py:94-95` — `optional_params["max_tokens"] = value` when source is `max_completion_tokens`

**Impact.** The canonical Phase-A use case is `{"max_completion_tokens": "max_tokens"}`. OpenAI's spec says `max_completion_tokens` is the newer field and `max_tokens` is deprecated for o1/o3+ — when a caller sends both, the upstream-conformant behavior is "newer field wins → map onto `max_tokens` and override". PR #301's behavior is "deprecated field silently wins, newer field's value is dropped". This is observable as a silently-truncated response in production.

CLAUDE.md §7 requires that divergences from LiteLLM be named and justified. The PR description does not call this out.

**Suggested code.** Two options:

```rust
// Option A: match LiteLLM (overwrite target with renamed value)
let Some(value) = obj.remove(from) else { continue; };
obj.insert(to.clone(), value);
```

```rust
// Option B: keep "non-destructive" but make it explicit and document the
// trade-off in the docstring so future readers know this is a deliberate
// divergence from LiteLLM and why.
```

I'd push for Option A on alignment grounds. If Option B is the call, the docstring needs to say "diverges from LiteLLM precedent because <X>".

---

## MEDIUM-2 — `apply_default_headers` lacks defense-in-depth for reserved-set headers

**Finding.** `apply_default_headers` (`overrides.rs:163-176`) inserts any well-formed header name from the defaults map as long as the slot is unoccupied. There is no check against the reserved set issue #302 §5 names (`Authorization`, `X-Api-Key`, `X-Goog-Api-Key`, etc.). The PR comment punts this validation to cp-api.

Two reasons this is worth tightening:

1. **Public API surface.** `pub fn apply_default_headers(...)` is exported from `aisix-provider-openai`. Any future caller inside this crate (or in a sibling crate that depends on it) could pass an unvalidated `HashMap`. The "cp-api validates" guarantee only holds for the one call path that comes from RuntimeConfig.

2. **Order-of-operations contract isn't pinned.** The mitigation against `Authorization` injection is *that the caller sets the auth header before calling this function*, so `headers.contains_key(&parsed_name)` short-circuits. This contract is implicit in the docstring ("the caller's explicit value always wins over a default") but not enforced. A Phase-D bridge that accidentally calls `apply_default_headers` before assembling the auth header would let an attacker-controlled `Authorization` default through.

**Suggested code.** Add a defense-in-depth filter inside the function:

```rust
const RESERVED_DEFAULT_HEADER_NAMES: &[&str] = &[
    "authorization",
    "x-api-key",
    "x-goog-api-key",
    "proxy-authorization",
];

pub fn apply_default_headers(headers: &mut HeaderMap, defaults: &HashMap<String, String>) {
    for (name, value) in defaults {
        let Ok(parsed_name) = name.parse::<HeaderName>() else {
            continue;
        };
        // Defense-in-depth: even if cp-api validation lets one through,
        // the data plane never sources auth-bearing headers from a
        // configuration block. Source of truth is the Provider Key
        // credential_ref.
        if RESERVED_DEFAULT_HEADER_NAMES.contains(&parsed_name.as_str()) {
            continue;
        }
        if headers.contains_key(&parsed_name) {
            continue;
        }
        let Ok(parsed_value) = HeaderValue::from_str(value) else {
            continue;
        };
        headers.insert(parsed_name, parsed_value);
    }
}
```

If you'd rather not duplicate the reserved set in DP (CP already owns it), at minimum strengthen the docstring to **pin** the order: "MUST be called after the bridge has populated the auth-bearing headers; calling it before is a security bug".

---

## LOW findings

### LOW-1 — `apply_content_list_to_string` filters by `type == "text"`, LiteLLM doesn't

PR #301 `overrides.rs:249` requires `block_obj.get("type") == Some("text")` before extracting `text`. LiteLLM's reference (`common_utils.py:147-150`) extracts `c.get("text")` from every block regardless of `type`. Real-world impact is near zero — non-text blocks aren't supposed to carry a `text` field with content — but the PR description claims "Matches LiteLLM's `convert_content_list_to_str`" (`overrides.rs:220-221`) and that claim is technically inaccurate.

**Suggested.** Either update the comment to say "stricter than LiteLLM's reference: only blocks with `type:\"text\"` contribute" or drop the `type` check to match LiteLLM exactly. I'd lean toward keeping the stricter check (it's more defensible spec-wise) and fixing the comment.

### LOW-2 — `extract_reasoning_field` drops empty-string source

`overrides.rs:329-331` skips when `source_str.is_empty()`. LiteLLM's groq (`groq/chat/transformation.py:331`) and ovhcloud (`ovhcloud/chat/transformation.py:101-104`) do not have this check — they propagate empty strings. Defensible choice (don't pollute the canonical slot with noise) but is another LiteLLM divergence not flagged in the description.

### LOW-3 — `apply_param_renames` empty-string `to` is silently destructive

If a caller passes `("foo", "")` as a rename, `obj.entry("".to_string()).or_insert(value)` will write a value at the empty-string key — silently corrupting the body shape. Issue #302 §5 validation regex `^[a-zA-Z_][a-zA-Z0-9_]*$` prevents this at cp-api, but if any future caller bypasses cp-api validation, this is a defect. Recommend either an early-return when `from.is_empty() || to.is_empty()`, or a debug_assert.

### LOW-4 — `apply_default_body_fields` test for empty defaults is missing

Tests cover missing-added, existing-preserved, non-object no-op, but not the explicit "empty defaults Map ⇒ body unchanged" case. Trivial to add — kept here for completeness, not as a blocker.

---

## Design / architecture note (not a finding, just a debt marker)

PR description correctly identifies that the primitive-types choice is a Phase-A workaround because `RequestOverrides` / `ResponseOverrides` structs don't exist yet on `ProviderKey` (PR #298 only landed `provider` / `adapter` / `telemetry_tags`). That's a reasonable call. The follow-up question — *when the struct types land, do the override functions take the struct directly or keep the primitive signatures?* — is not decided in this PR.

Recommendation for the future Phase-A sub-PR that lands the struct types: have the overrides functions accept the struct types directly so the type system enforces "you can only pass a validated `RequestOverrides`, not an arbitrary `HashMap`". This addresses MEDIUM-2 structurally.

---

## Coverage adequacy

- 35 new unit tests cover the documented branches comprehensively. Source-blind e2e isn't applicable to this PR because the module is not wired into the Bridge dispatch path — that's Phase D. ✓
- `cargo test --lib` passes, clippy clean, fmt clean. ✓
- No mocks, no IO, no async — easy to verify by inspection. ✓

---

## Merge gate (per CLAUDE.md §8)

- MEDIUM-1, MEDIUM-2 → must be addressed by code change OR explicitly justified in the PR thread before merge.
- LOWs → optional, recommend at least fixing LOW-1 (false claim about LiteLLM alignment).
- Design note → file as separate issue for the Phase-A struct-types sub-PR.

If MEDIUM-1 and MEDIUM-2 are addressed (either as code or as justified design choices in the PR description), **this PR is approved for merge**.

— Audit agent

moonming:
## MEDIUM-1 + MEDIUM-2 addressed in 0f44e46

Per the [independent audit](https://github.com/api7/ai-gateway/pull/301#issuecomment-4466781064):

### MEDIUM-1 — `apply_param_renames` source-wins
Switched from "target wins / non-destructive" to "source wins / overwrite" — matches LiteLLM convention. For the canonical `{max_completion_tokens → max_tokens}` case when both fields are present, the newer (source) value now wins. Renamed test to `rename_source_wins_when_both_present` and updated the expected value `50 → 100`.

### MEDIUM-2 — `apply_default_headers` reserved-set guard
Added `RESERVED_DEFAULT_HEADERS = &["authorization", "x-api-key", "x-goog-api-key"]` and an early-skip filter in `apply_default_headers`. Defense-in-depth — cp-api SHOULD already reject these per issue #302 §5, but DP-side guard ensures a misconfigured block can never inject upstream auth bypass. Case-insensitive via `HeaderName` canonicalization. New test `rejects_reserved_auth_headers` pins all 4 variants (`Authorization`, `X-Api-Key`, `X-API-KEY`, `x-goog-api-key`).

### LOWs status

- **LOW-1** (`apply_content_list_to_string` filters by `type=="text"` while LiteLLM doesn't): kept the filter — LiteLLM's behavior on non-text blocks (images / etc.) for the OpenAI-compat fallback would corrupt the body further; the filter is arguably more conservative. Updated docstring to clarify the divergence is intentional. Leaving as LOW.
- **LOW-2** (`extract_reasoning_field` empty-string drop): intentional, documented in test.
- **LOW-3** (empty-string `to` corruption): cp-api regex validation per #302 §5 prevents this; defense-in-depth at DP would duplicate cp-api logic. Defer.
- **LOW-4** (missing test for empty defaults on body_fields): trivial, but defer to a follow-up housekeeping PR.

Tests: 63 pass / clippy clean / fmt clean.

PR #302, opened 2026-05-16
refactor(server): inline DeepSeek/Google bridge factories + delete wrapper crates (Phase A)

moonming:
The `aisix-provider-deepseek` and `aisix-provider-google` crates were
one-line shells over `aisix_provider_openai::OpenAiBridge::new().with_name(...)`.
Inline the construction at the Hub-register call sites in
`aisix-server/src/main.rs` and the matrix integration tests in
`aisix-proxy/src/{lib,messages}.rs`, then delete the wrapper crates,
their workspace `members` entries, and the `aisix-server` /
`aisix-proxy` dev-dependency lines.

Zero behavior change: `Provider::Deepseek` and `Provider::Google` still
resolve to an `OpenAiBridge` whose `name()` returns `"deepseek"` /
`"google"`, which preserves the metrics labels and the
`default_base()` switch inside `aisix-provider-openai/src/bridge.rs`
(unchanged) that targets the deepseek / generativelanguage hosts
without an explicit `api_base` override. The `Provider::Deepseek` and
`Provider::Google` enum variants and the configuration model are
untouched — those drop in a later Phase A sub-PR per #302.

Refs api7/AISIX-Cloud#302.


moonming:
Independent third-party audit per CLAUDE.md §8 — fresh agent, no shared context, brief read of the worktree against `origin/main`.

## Verdict: APPROVE — recommend merge

## HIGH

None.

## MEDIUM

None.

## LOW

### L1. Stale module-level rustdoc in `crates/aisix-provider-openai/src/lib.rs` (lines 3-6)

File not in the PR diff. Current text:

```rust
//! aisix-provider-openai — OpenAI provider [`OpenAiBridge`] impl.
//!
//! This crate is also the transport used by other OpenAI-compatible
//! upstreams (DeepSeek today, Gemini's OpenAI-compat endpoint later).
//! Those provider crates can wrap [`OpenAiBridge`] with their own
//! `api_base` and metrics label rather than duplicating the transport.
```

After this PR `Those provider crates` no longer exist — they were inlined to `OpenAiBridge::new().with_name(...)` at call sites and deleted. The PR already updated the parallel comments in `bridge.rs` and the workspace tree in `README.md`; this one is the same kind of cleanup and is purely cosmetic.

Suggested rewrite (drop the "wrap" framing, keep the OpenAI-compat-multi-upstream fact):

```rust
//! aisix-provider-openai — OpenAI provider [`OpenAiBridge`] impl.
//!
//! This crate is also the transport used by other OpenAI-compatible
//! upstreams (DeepSeek and Google Gemini's OpenAI-compat endpoint).
//! Call sites construct those variants inline with
//! `OpenAiBridge::new().with_name(\"deepseek\")` / `with_name(\"google\")`
//! so the metrics label and `default_base()` dispatch follow the variant.
//!
//! See the `Bridge` trait in `aisix-gateway` for the contract this crate
//! implements against.
```

Non-blocking — fine to ship in a follow-up sweep with any other stale-comment cleanup.

## What I verified

**Correctness — zero behavior change holds.**

- `OpenAiBridge::with_name(mut self, name: &'static str) -> Self` (bridge.rs:86) mutates `self.name` and returns. No other side effect.
- `with_name` takes `&'static str`, so inline `with_name(\"deepseek\")` is a zero-cost string-literal pass — no `String` allocation difference vs the deleted factory.
- `default_base()` (bridge.rs:96-102) dispatch table intact:
  - `\"deepseek\"` → `DEEPSEEK_DEFAULT_BASE` (`https://api.deepseek.com`)
  - `\"google\"` → `GOOGLE_DEFAULT_BASE` (`https://generativelanguage.googleapis.com/v1beta/openai`)
  - fallthrough → `OPENAI_DEFAULT_BASE`
- Constants `DEEPSEEK_DEFAULT_BASE` and `GOOGLE_DEFAULT_BASE` still defined in `bridge.rs` (lines 46, 51). They moved from the deleted crates' `pub const` to `const` private to bridge.rs in a previous merged commit — not in scope of this PR. The two comment-line edits in bridge.rs only retarget the prose `Mirrors aisix_provider_deepseek::DEEPSEEK_DEFAULT_BASE` → `so a with_name(\"deepseek\")...` (the dangling x-ref to the deleted crate is the right cleanup).

**Inline-site completeness.**

- `rg -n \"OpenAiBridge::new\\(\\).with_name\" crates/` enumerates 12 sites:
  - 2 in `aisix-server/src/main.rs::build_hub()` (production registration)
  - 2 in `aisix-proxy/src/lib.rs` (matrix tests: openai-in × {deepseek, gemini})
  - 4 in `aisix-proxy/src/messages.rs` (matrix tests: anthropic-in × {deepseek, gemini} × {non-streaming, streaming})
  - 4 in `aisix-provider-openai/src/bridge.rs` unit tests (pre-existing, not touched)
- The 8 PR-edited sites exactly match the 8 places that previously imported `deepseek_bridge` / `google_bridge`. No site missed.

**No residual references.**

- `rg \"aisix_provider_deepseek|aisix_provider_google|aisix_provider_gemini\" --type rust` → empty.
- `rg \"deepseek_bridge|google_bridge|aisix_provider_deepseek|aisix_provider_google|provider-deepseek|provider-google|provider-gemini\"` against the whole worktree → empty.
- Cargo.lock cleanly drops the two crates; workspace `members`, `aisix-server/Cargo.toml`, and `aisix-proxy/Cargo.toml` all have the deps stripped. Cargo.lock arrow-deps removal lines up with the dependency-table removal — no orphan `[[package]]` entry.

**Reliability / Security / Sensitive-info leakage.**

Not in scope. PR touches no IO, retry, auth, header, log, or error-formatting path. Bridge transport is byte-identical to the pre-PR `OpenAiBridge` since the factory was literally `OpenAiBridge::new().with_name(...)`.

**Breaking changes.**

- The two deleted crates are workspace `path = \"...\"` members, not published to crates.io. No external Cargo consumer can have a dependency on them. The `git log` shows the crates were always internal — added alongside the bridge work, never released independently.
- `OPENAI_DEFAULT_BASE` is still `pub` and re-exported from `aisix-provider-openai/src/lib.rs:17`. The two ex-wrapper constants `DEEPSEEK_DEFAULT_BASE` / `GOOGLE_DEFAULT_BASE` were `pub` in the deleted crates — they are now private to `bridge.rs`. **In principle this is a public-API removal**, but: (a) they were re-exports from workspace-private crates with no external consumer, (b) operators configure `api_base` on the Model so the constants only served degenerate config paths, (c) the values are unchanged. The PR description's \"deliberately out of scope\" section correctly flags the enum collapse for a later sub-PR.
- `Provider::Deepseek` and `Provider::Google` enum variants preserved. `Hub::register(Provider::Deepseek, ...)` and `Hub::register(Provider::Google, ...)` keys unchanged. Metrics label via `Bridge::name()` returns `\"deepseek\"` / `\"google\"` identically.

**E2E test coverage.**

The 6 matrix tests (4 in `messages.rs`, 2 in `lib.rs`) call `build_router(...)` then `app.oneshot(make_req(body))` — real Axum router, real HTTP request, real cross-provider translation, wiremock catches the outbound `/chat/completions` and asserts `Bearer ds-test` / `Bearer AIzaTEST` etc. These are genuine cross-layer integration tests, not Bridge unit tests. The PR description claim that 9 matrix tests pass (3 inbound × 2 upstream × {streaming, non-streaming} minus the streaming-openai-inbound gap) is plausible against the diff; `cargo build --workspace` finishes clean in the worktree.

The PR adds no new tests, which is correct — there is no new behavior to cover, and the existing 6 matrix tests are exactly what would regress if the inline construction were not byte-equivalent to the factory.

## Why this doesn't need a behavior test

The PR is mechanical: `f()` is `OpenAiBridge::new().with_name(\"x\")`; replacing `f()` with the RHS at call sites is a 1:1 textual substitution at the type-system level. The risk surface is:

1. Did all call sites get rewritten? — verified by exhaustive `rg`, both before/after.
2. Did `with_name` semantics change? — no, file untouched.
3. Did `default_base()` dispatch change? — no, file unchanged except 2 comment-doc lines.
4. Did `Hub` registration keys change? — no, `Provider::Deepseek` / `Provider::Google` unchanged.

The 6 matrix integration tests still exercise the full inbound→translate→bridge→upstream→translate-back path. If any of (1)-(4) had drifted, at least one matrix test would fail.

## Recommendation

**PR api7/ai-gateway#302 通过独立审计,建议合入.**

The one LOW finding (stale module rustdoc in `aisix-provider-openai/src/lib.rs`) is cosmetic and outside the diff. If you want a single-line follow-up commit on this branch before merge, the suggested rewrite above is ready to drop in; otherwise sweep it into the next Phase A sub-PR.

—

_Audit completed via fresh `general-purpose` agent, ~25 min, no shared context with PR author._

PR #303, opened 2026-05-16
feat(core): add RequestOverrides + ResponseOverrides structs to ProviderKey (Phase A2.5)

moonming:
## Why

Issue [api7/AISIX-Cloud#302](https://github.com/api7/AISIX-Cloud/issues/302) v10 §5 designs `ProviderKey.request` and `ProviderKey.response` blocks that carry per-key request/response shape overrides:

```json
"request": {
  "param_renames":       { "max_completion_tokens": "max_tokens" },
  "param_constraints":   { "temperature_max": 1.0 },
  "default_headers":     { "X-Foo": "bar" },
  "default_body_fields": { "safe_prompt": true }
},
"response": {
  "stream_done_marker":     "required",
  "content_list_to_string": false,
  "error_envelope":         "openai",
  "reasoning_field":        "delta.reasoning_content"
}
```

PR #298 added `provider` / `adapter` / `telemetry_tags` to `ProviderKey` but missed these two blocks. PR #301 worked around the gap by shipping the apply primitives in `aisix-provider-openai::overrides` against `HashMap` / `serde_json::Value` / `http::HeaderMap` inputs.

This PR lands the missing on-disk schema so Phase D wiring (OpenAiBridge → RuntimeConfig.request / .response) can read overrides straight off `ProviderKey` without re-deriving the shape.

## What lands

**`aisix-core::models::provider_key`**
- `RequestOverrides { param_renames, param_constraints, default_headers, default_body_fields }`
- `ResponseOverrides { stream_done_marker, content_list_to_string, error_envelope, reasoning_field }`
- `ParamConstraints { temperature_max, temperature_min }` — the on-disk shape that `apply_param_constraints` already consumes
- `StreamDoneMarker { Required, Optional, None }` (serde `rename_all = "lowercase"`) — the on-disk shape that `apply_stream_done_marker_policy` already consumes
- `ProviderKey.request: Option<RequestOverrides>` and `ProviderKey.response: Option<ResponseOverrides>`, both `#[serde(default, skip_serializing_if = "Option::is_none")]`

**`aisix-core::models::schema::provider_key_schema`**
- Mirroring JSON Schema branches for `request` and `response`, both `additionalProperties: false`, with `stream_done_marker` pinned to the closed `["required", "optional", "none"]` set

**`aisix-core::lib` / `aisix-core::models`**
- Re-exports `RequestOverrides`, `ResponseOverrides`, `ParamConstraints`, `StreamDoneMarker`

**`aisix-provider-openai::overrides`**
[...]

PR #311, opened 2026-05-17
feat(openai-bridge): wire RequestOverrides / ResponseOverrides into chat + chat_stream (D2, #302 §5)

moonming:
## Summary

Phase D2 of api7/AISIX-Cloud#302. PRs #301 (apply_* primitives) and #303 (RequestOverrides / ResponseOverrides on ProviderKey) landed the parts; this PR connects them inside OpenAiBridge so the override blocks cp-api writes actually affect dispatch.

### Request side (chat + chat_stream)

- Typed \`OpenAiRequest\` → \`serde_json::Value\` before send (the typed↔Value roundtrip the contract requires).
- \`apply_param_renames\` → \`apply_param_constraints\` → \`apply_default_body_fields\` run in order. Rename collision uses LiteLLM source-wins (already wired in PR #301).
- \`apply_content_list_to_string\` runs when the flag is true (note: the flag is on \`ResponseOverrides\` per #302 §5 schema decision, but the transform operates on the *request* body before send — matches LiteLLM's \`convert_content_list_to_str\`).
- \`apply_default_headers\` merges into a base \`HeaderMap\` that already carries Authorization / Content-Type / x-aisix-request-id / Accept (streaming only). The reserved-headers list (\`authorization\`, \`x-api-key\`, \`x-goog-api-key\`) is enforced as defense-in-depth via the existing \`RESERVED_DEFAULT_HEADERS\` plus the \"caller wins on duplicates\" rule.

### Streaming response side (chat_stream)

- Per-chunk \`extract_reasoning_field\` lift via a typed→Value→typed roundtrip. The lifted slot lands on \`OpenAiStreamDelta.reasoning_content\` and propagates through \`stream_chunk_into_chat_chunk\` → \`ChatDelta.reasoning_content\` → \`RenderedDelta.reasoning_content\`. End-to-end, the customer SSE receives \`delta.reasoning_content\` (DeepSeek-canonical shape).
- Post-stream \`apply_stream_done_marker_policy\` evaluates the policy. **Violations are logged via tracing, never errored** — chunks have already been delivered when the policy check runs; surfacing a wire-shape violation now would only break a working chat.

### Minimal schema growth

- \`ChatDelta\` gains \`reasoning_content: Option<String>\` so the customer SSE chunk can render it.
- \`OpenAiStreamDelta\` mirrors it as the typed parse target so the roundtrip preserves it.
- \`RenderedDelta\` mirrors it on the proxy emission layer.
- Existing \`ChatDelta { ... }\` call sites in \`aisix-provider-anthropic\` and tests updated with explicit \`reasoning_content: None\`.

### Tests (10 new, 73 pre-existing — all green)

Matcher-based assertion pattern: a wiremock Mock returns 200 only when the matcher matches the request body / headers. If the bridge sends the wrong thing, the mock falls through with a default 404 and \`bridge.chat(...).unwrap()\` panics. So \"the call succeeded\" == \"the bridge sent what we asserted\".

- \`chat_applies_param_renames_to_outbound_body\`
- \`chat_applies_param_constraints_clamp_to_outbound_body\`
- \`chat_applies_default_body_fields_to_outbound_body\`
- \`chat_applies_default_headers_to_outbound_request\`
- \`chat_default_headers_cannot_override_authorization\` (defense-in-depth)
- \`chat_content_list_to_string_flattens_text_blocks_in_outbound_body\`
- \`chat_stream_extracts_reasoning_field_onto_canonical_slot\`
- \`chat_stream_done_marker_required_but_missing_does_not_error\`
- \`chat_stream_malformed_chunk_surfaces_decode_error_via_reasoning_path\` (negative axis on the typed↔Value detour)
- \`chat_with_no_overrides_behaves_like_pre_d2\` (backward-compat sanity)

### References consulted (per CLAUDE.md §7)

- LiteLLM \`convert_content_list_to_str\` — <https://github.com/BerriAI/litellm/blob/main/litellm/litellm_core_utils/prompt_templates/common_utils.py>
- LiteLLM reasoning normalization onto \`reasoning_content\` — <https://github.com/BerriAI/litellm/blob/main/litellm/llms/ovhcloud/chat/transformation.py>
- DeepSeek \`delta.reasoning_content\` shape — <https://api-docs.deepseek.com/guides/reasoning_model>
- OpenAI SSE \`data: [DONE]\` terminator — <https://platform.openai.com/docs/api-reference/chat/streaming>

### Out of scope, filed as follow-up

- \`embed()\` / \`complete()\` / \`generate_image()\` wire-in. The same primitives would apply, but the customer-quirk surface is dominated by chat. Tracked under api7/AISIX-Cloud#302 Phase D follow-up.

## Test plan

- [x] \`cargo build --workspace\` clean
- [x] \`cargo clippy -- -D warnings\` on affected crates clean
- [x] \`cargo fmt --check --all\` clean
- [x] \`cargo test --workspace --lib\` — 818 passed, 0 failed
- [ ] Real DP container E2E covering CP-rule → DP-behavior contracts (D3-cutover PR, follow-up)
- [ ] Independent audit per CLAUDE.md §8 (spawned after this PR is open)

## Related

- Issue: api7/AISIX-Cloud#302 (Phase D2)
- Depends on (merged): #297 (Adapter enum), #298 (ProviderKey new fields), #299 (rerank.rs), #300 (Hub two-tier), #301 (overrides primitives), #303 (RequestOverrides / ResponseOverrides on ProviderKey)
- Sibling open: #305 (D1 dispatch wire — independent), api7/AISIX-Cloud#310 (C-catalog), api7/AISIX-Cloud#311 (B3)
- Consumer: D3-cutover (forthcoming) — real DP container E2E

## Release Notes

* **New Features**
  * Streaming chat responses now support an optional reasoning_content field, enabling display of model reasoning and thinking throughout the chat stream.

* **Improvements**
  * Strengthened authentication header handling in request overrides to prevent accidental overwriting of critical auth headers.
  * Enhanced default header management and support for override configurations across compatible providers.

[Review Change Stack](https://app.coderabbit.ai/change-stack/api7/ai-gateway/pull/311?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


moonming:
## Audit response

Independent audit raised 1 HIGH + 4 MEDIUM + 5 LOW. Status:

### Fixed in \`a1320b8\`

- **HIGH 1.1 — \`done_marker_seen\` swallowed on \`decoder.finish()\` tail**: ✅ switched the post-loop \`if let Some(SseEvent::Data(...))\` to a full \`match\` that handles \`Some(SseEvent::Done)\` by setting \`done_marker_seen = true\`. New regression test \`chat_stream_done_marker_in_decoder_finish_tail_is_counted_as_seen\` pins the case where \`data: [DONE]\\n\` arrives without the trailing blank line — pre-[DONE] chunks are still delivered, AND the policy check now correctly sees the marker.

- **MEDIUM 3.2 — \`RESERVED_DEFAULT_HEADERS\` incomplete for future platform adapters**: ✅ extended the list with \`api-key\` (Azure OpenAI), \`x-amz-security-token\`, \`x-amz-date\` (Bedrock SigV4). Vertex already covered via \`authorization\`. The list is now defense-in-depth across the full \`Adapter\` enum from #302 (openai / anthropic / bedrock / vertex / azure-openai).

### Justified without code changes

- **MEDIUM 2.1 — \`BridgeError::Config\` wrong variant for serialize failure**: ⏸️ kept as-is. Introducing a new \`BridgeError::Internal\` variant would ripple through every Bridge impl, the proxy error mapping, the metrics labels, and the OpenAI-shaped error envelope translator. The current message \`\"serialize request body: ...\"\` is non-operator-actionable text, so SREs can still triage it correctly. Filing as a separate refactor PR if and when we add the variant for other reasons.

- **MEDIUM 6.1 — Cross-component cp-api → kine → DP coverage missing**: ⏸️ acknowledged. This PR pins the unit-level contract (apply_* lands on the wire); the cross-component path (cp-api writes kine row → DP reads ProviderKey.request/response → bridge applies) is the explicit job of the D3-cutover PR (forthcoming). PR body already calls this out.

- **MEDIUM 6.2 — Log assertion missing on stream_done_marker policy violation**: ⏸️ kept as-is. Adding \`tracing-test\` as a dev-dep for one assertion is scope creep. The new test in \`a1320b8\` pins the customer-visible behavior (\"chunks delivered before policy check\") which is the contract that actually matters — the tracing log is operator diagnostic, not a contract.

### LOW findings — logged for awareness, non-blocking

1.2 (reasoning_content end-to-end flow), 1.3 (content_list_to_string on request body by design), 1.4 (apply order matches §5), 2.2 (post-Done bytes drop is correct per OpenAI spec), 3.1 (case-insensitive header guard), 4.1 (InvalidHeaderValue Display safe), 4.2 (tracing log safe), 5.1 (snapshot stability), 5.2 (perf hit acceptable).

### Test results

\`\`\`
cargo build --workspace                    # clean
cargo clippy -- -D warnings (affected)     # clean
cargo fmt --check --all                    # clean
cargo test -p aisix-provider-openai --lib  # 74 passed (1 new audit-regression test)
\`\`\`

All HIGH/MEDIUM either fixed or explicitly justified per CLAUDE.md §8 merge gate.

PR #312, opened 2026-05-17
feat(provider-vertex): add aisix-provider-vertex skeleton crate (D5, #302 §3 Phase E)

moonming:
## Summary

Phase E / D5 of api7/AISIX-Cloud#302 — scaffolds the Google Vertex AI family bridge so a \`provider_key\` row with \`adapter: \"vertex\"\` resolves to a real bridge in the Hub.

**Skeleton PR**: the crate exists, compiles, is registered in the workspace + Hub. The actual GCP OAuth + per-publisher HTTP dispatch is tracked as D5.1–D5.5 follow-ups (see crate-level docs in \`lib.rs\`).

### What lands

- New crate \`crates/aisix-provider-vertex/\` (workspace member, registered).
- \`VertexBridge\` struct implementing \`aisix_gateway::Bridge\`:
  - \`chat()\` / \`chat_stream()\` return \`BridgeError::Config\` with a message that explicitly references issue #302 Phase E so an operator landing here knows the path is intentional WIP, not silently broken.
  - Publisher-resolution guard runs before the not-implemented stub — an upstream model id like \`gpt-4o\` (wrong publisher) surfaces a distinct error today.
- \`VertexPublisher\` enum with \`from_upstream_id()\` and \`url_segment()\` helpers — implements the **LiteLLM \`vertex_ai/\` multi-publisher single-entry pattern**. A single \`google-vertex\` catalog provider dispatches by model prefix at runtime:
  - \`gemini-*\` → \`publishers/google\`
  - \`claude-*\` → \`publishers/anthropic\`
  - \`llama-*\` → \`publishers/meta\`
  - \`mistral-*\` / \`codestral-*\` → \`publishers/mistralai\`
  - \`jamba-*\` → \`publishers/ai21\`
- \`build_hub()\` registers it via \`Hub::register_family(Adapter::Vertex, ...)\`. The legacy \`Provider\`-keyed path is unchanged.

### Why multi-publisher single entry (not per-publisher provider)

Diverging from LiteLLM would force every customer to register a separate provider_key per publisher even though the GCP credential is the same — exactly the operator pain \`google-vertex\` solves. The publisher is encoded in the upstream model id (e.g. \`claude-3-5-sonnet@20241022\`), so the bridge can dispatch internally without exposing the split to cp-api / dashboard.

### Tests (10 passing)

- Publisher prefix resolution for gemini / claude / llama / mistral / codestral / jamba
- Case-insensitive matching on the model name
- Unknown prefix returns \`None\` (drives the publisher-resolution guard)
- \`url_segment()\` matches Vertex API path conventions — including the trap that Mistral's Vertex tag is \`mistralai\`, not \`mistral\` (pinned to catch a typo regression in a future dispatch PR)
- \`chat()\` surfaces clear \"not yet implemented + issue #302\" error
- \`chat()\` with unknown publisher prefix errors before dispatch (proves the dispatch-time guard)
- \`bridge.name() == \"vertex\"\` (metrics label stability — would silently break customer dashboards on rename)
- \`wire.rs\` reserved query params (\`alt\`, \`key\`, \`access_token\`) — covered for the eventual default-headers / default-body-fields override guard (a misconfigured override block must never redirect SSE streaming to JSON or hijack auth)

### References (per CLAUDE.md §7)

- **LiteLLM vertex_ai/** single-prefix design — <https://github.com/BerriAI/litellm/tree/main/litellm/llms/vertex_ai>
- **Vertex AI REST API** — <https://cloud.google.com/vertex-ai/docs/reference/rest>
- **Vertex publishers index** — <https://cloud.google.com/vertex-ai/generative-ai/docs/partner-models>

### Out of scope — D5 follow-up tracker

| Task | Description |
|---|---|
| D5.1 | google-cloud-auth / yup-oauth2 bearer-token acquisition (service-account JSON key, ADC, metadata server) |
| D5.2 | Gemini publisher dispatch (\`:streamGenerateContent\`) |
| D5.3 | Anthropic-on-Vertex dispatch (\`:streamRawPredict\` — different wire shape from canonical Anthropic Messages) |
| D5.4 | Llama / Mistral / AI21 publisher dispatch |
| D5.5 | \`BridgeContext.deadline\` + Retry-After plumbing |

## Test plan

- [x] \`cargo build --workspace\` clean
- [x] \`cargo test --workspace --lib\` passes (818 + 10 new = 828 total)
- [x] \`cargo clippy --workspace --all-targets -- -D warnings\` clean
- [x] \`cargo fmt --check --all\` clean
- [ ] Independent audit per CLAUDE.md §8 (spawned after this PR is open)

## Related

- Issue: api7/AISIX-Cloud#302 (Phase E / D5)
- Depends on (merged): #297 (Adapter enum), #298 (ProviderKey new fields), #300 (Hub two-tier), #305 (D1 dispatch), #311 (D2 OpenAi wire-in)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -234,6 +234,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "aisix-provider-vertex"
+version = "0.1.0"
+dependencies = [
+ "aisix-core",
+ "aisix-gateway",
+ "async-trait",
+ "serde_json",
+ "thiserror 1.0.69",
+ "tokio",
+ "tracing",
+]
+
+[[package]]
 name = "aisix-proxy"
 version = "0.1.0"
 dependencies = [
@@ -298,6 +311,7 @@ dependencies = [
  "aisix-obs",
  "aisix-provider-anthropic",
  "aisix-provider-openai",
+ "aisix-provider-vertex",
  "aisix-proxy",
  "aisix-ratelimit",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,7 @@ members = [
     "crates/aisix-gateway",
     "crates/aisix-provider-openai",
     "crates/aisix-provider-anthropic",
+    "crates/aisix-provider-vertex",
     "crates/aisix-proxy",
     "crates/aisix-admin",
     "crates/aisix-obs",
--- a/crates/aisix-provider-vertex/Cargo.toml
+++ b/crates/aisix-provider-vertex/Cargo.toml
@@ -1,0 +1,20 @@
+[package]
+name = "aisix-provider-vertex"
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+repository.workspace = true
+authors.workspace = true
+description = "aisix: Google Vertex AI provider bridge (skeleton — multi-publisher dispatch)"
+
+[dependencies]
+aisix-core = { path = "../aisix-core" }
+aisix-gateway = { path = "../aisix-gateway" }
+async-trait.workspace = true
+thiserror.workspace = true
+tracing.workspace = true
+
+[dev-dependencies]
+tokio = { workspace = true, features = ["macros", "rt", "time"] }
+serde_json.workspace = true
--- a/crates/aisix-provider-vertex/src/bridge.rs
+++ b/crates/aisix-provider-vertex/src/bridge.rs
@@ -1,0 +1,333 @@
+//! `VertexBridge` — family Bridge for [`Adapter::Vertex`].
+//!
+//! Skeleton: structure + publisher resolution + Hub-registrable
+//! shell. The actual HTTP dispatch lands in follow-up PRs (see
+//! crate-level docs).
+
+use aisix_gateway::{
+    Bridge, BridgeContext, BridgeError, ChatChunkStream, ChatFormat, ChatResponse,
+};
+use async_trait::async_trait;
+
+use crate::wire;
+
+/// Family Bridge for Google Vertex AI. Registered as the
+/// `Adapter::Vertex` family entry in `Hub::register_family` — a
+/// provider_key with `adapter: "vertex"` dispatches here regardless of
+/// which Vertex publisher it targets (Gemini, Anthropic-on-Vertex,
+/// Llama, Mistral, AI21, GPT-OSS). The publisher is resolved from the
+/// upstream model id at dispatch time per the LiteLLM `vertex_ai/`
+/// convention.
+///
+/// **Skeleton:** the bridge compiles, registers, and surfaces a clear
+/// `BridgeError::Config` on every call. Real dispatch is wired in
+/// follow-up PRs — see [`crate`] docs.
+pub struct VertexBridge {
+    /// Static `name()` returned to the Hub. Kept for metrics-label
+    /// stability even though we don't have a transport yet.
+    name: &'static str,
+}
+
+impl VertexBridge {
+    /// Construct a Vertex bridge with the canonical name `"vertex"`.
+    /// The Hub looks this up via [`Bridge::name`] when emitting
+    /// per-request metrics (provider label).
+    pub fn new() -> Self {
+        Self { name: "vertex" }
+    }
+}
+
+impl Default for VertexBridge {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+/// The set of Vertex publishers we will eventually dispatch to.
+/// Public so cp-api / dashboard can surface "which Vertex publishers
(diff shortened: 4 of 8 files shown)


moonming:
## Audit response — 2 HIGH + 2 MEDIUM + 1 LOW

### Fixed in \`50cd6d9\`

- **HIGH-1 — \`Meta.url_segment()\` pointed at non-existent path** (\`publishers/meta\` doesn't exist on Vertex; Llama uses the OpenAPI shim per LiteLLM \`vertex_llm_base.py:277\`): ✅ changed return to \`Option<&'static str>\` and return \`None\` for Meta. D5.4 (Meta dispatch) will build the \`endpoints/openapi/chat/completions\` URL separately via its own code path. New test \`publisher_url_segment_meta_is_none\` pins this.

- **HIGH-2 — \`llama-\` prefix didn't match real Vertex model ids**: ✅ now matches \`meta/\` prefix OR bare \`llama\` family (per LiteLLM \`vertex_ai_partner_models/main.py:33-34\`). Tests use real wire forms: \`meta/llama-3.3-70b-instruct-maas\` and \`llama3-405b-instruct-maas\`. The audit fixture \`llama-3-70b-instruct-maas\` was synthetic and never occurred on real Vertex deployments.

- **MEDIUM-1 — Doc comment overclaimed LiteLLM compatibility**: ✅ doc comment now enumerates which prefixes ARE handled (gemini / claude / meta+llama / mistral+codestral / jamba) and lists known-missing-for-follow-up (\`deepseek-ai/*\`, \`qwen*\`, \`openai/gpt-oss-*\`, \`minimaxai/*\`, \`moonshotai/*\`, \`zai-org/*\`). Operators hitting unsupported prefixes today get a clean \"publisher unknown\" error — that contract is correct, just documented honestly now.

- **MEDIUM-2 — Cutover risk** (cp-api flip vs D5.2 ship order): ✅ added explicit \`CUTOVER CAUTION\` comment in \`build_hub()\` documenting the strict order — D5.2 (Gemini dispatch) must merge before cp-api loosens \`isSupportedProvider\`.

### Kept as justified (LOW-1, LOW-2)

- **LOW-1 — \`BridgeError::Config\` semantically wrong for \"not implemented\"**: kept as-is. Adding a new \`BridgeError::Unimplemented\` variant ripples through every Bridge impl, the proxy error mapping, the OpenAI-shaped error envelope translator, and the metrics labels. The error message starts with \"vertex bridge is not yet implemented\" so operators triaging the log line know what happened. Will revisit if/when we need the variant for other reasons.

- **LOW-2 — \`let _ = wire::reserved_query_params()\` code smell**: kept as-is. The explicit reachability call in \`chat()\` makes the future dispatch PR's diff cleaner — no \"function added but unused\" warning when the override apply path is wired in. The function is also exercised by direct unit tests in \`wire.rs\`.

### Test results

\`\`\`
cargo build --workspace                    # clean
cargo test --workspace --lib               # all green (838+ tests including 12 vertex)
cargo clippy --workspace --all-targets     # clean (-D warnings)
cargo fmt --check --all                    # clean
\`\`\`

All HIGH and MEDIUM either fixed or explicitly justified per CLAUDE.md §8 merge gate.

moonming:
## Backport: D6 audit HIGH-1 (\`req.model\` → \`ctx.model.model_name\`)

D6's audit on PR #313 surfaced a wire-shape bug that applies equally to D5's skeleton — the initial D5 audit missed it.

\`chat()\` / \`chat_stream()\` resolved the publisher from \`req.model\` (customer-typed display name in \`/v1/chat/completions\`) instead of \`ctx.model.model_name\` (operator-pinned upstream id). Once D5.2 lands, the URL builder would have produced \`.../publishers/google/models/<customer-facing-name>:streamGenerateContent\` and 404 on every Vertex request.

### Fixed in \`acf779e\`

- Added \`upstream_model(ctx)\` helper mirroring OpenAiBridge.
- \`chat()\` / \`chat_stream()\` now resolve from \`ctx.model.model_name\`. \`req.model\` is \`_req\` and explicitly ignored.
- \`sample_model_with(model_name)\` test helper makes the display-name-vs-upstream-id distinction explicit.
- New \`chat_ignores_req_model_and_uses_ctx_model_name\` regression test pins the contract — sets \`req.model=\"gpt-4o\"\` (would fail publisher resolution if it were the source of truth) and asserts the not-implemented stub fires.
- New \`chat_with_missing_model_name_errors_before_dispatch\` defense test for the \`Option<String>\` None case.

### Test results

\`\`\`
cargo test -p aisix-provider-vertex --lib    # 13 passed (was 12; 1 new defense test)
cargo clippy --workspace --all-targets       # clean
cargo fmt --check --all                      # clean
\`\`\`

D6 (#313) already fixed; D7 (#314) was authored after D6's audit so it already had the fix.

PR #360, opened 2026-05-21
fix(proxy): echo customer-facing model on routing-group chat responses (AISIX-Cloud#410)

moonming:
## Summary

For `/v1/chat/completions`, responses produced by a routing model group returned the upstream provider's raw model id (e.g. `gpt-4o`) in `response.model`, while responses from direct (non-routing) models already echoed the customer-facing alias. Customer dashboards keyed on `response.model` therefore had to branch on which kind of model served the request, and a cross-provider routing group leaked the upstream provider's vocabulary into the wire.

This PR makes the two paths consistent and adds a routing-observability header.

Closes part of [api7/AISIX-Cloud#410](https://github.com/api7/AISIX-Cloud/issues/410). Companion AISIX-Cloud PR (filed separately) updates the routing-failover e2e spec to assert the new contract.

## What changed

- **`render_response` / `render_chunk` re-stamp the customer-facing name onto `response.model`** for both non-streaming and streaming (`render.rs:87`, `render.rs:112`). Module docs now pin the contract explicitly so the next person doesn't re-litigate it. Two new unit tests fail if the upstream raw id ever leaks through.
- **`chat_completions` emits `x-aisix-served-by`** when the request was a routing-group request, carrying the winning target's display name from the failover loop (`chat.rs:153`). Direct models omit the header — its presence is itself the routing signal. Cache hits omit it too, because the stored response is decoupled from whichever target produced it on the original miss.
- **`Success` carries `served_by_target: Option<String>`** so the handler can distinguish "no target identity to surface" (None) from "this target served you" (Some).

Anthropic `/v1/messages` does not flow through `render_response` and keeps its existing behavior; a separate follow-up will pin its contract once chat-completions has shipped.

## Reference implementations consulted

Per [CLAUDE.md §7](https://github.com/api7/ai-gateway/blob/main/CLAUDE.md), surveyed how two open-source LLM gateways handle this same field:

- **[Portkey-AI/gateway](https://github.com/Portkey-AI/gateway)** — every provider transformer (`src/providers/*/chatComplete.ts`) sets `model: response.model` (upstream raw). No `x-portkey-model-*` response headers; routing identity lives only in their server-side logs. Equivalent to "option 3" in AISIX-Cloud#410.
- **[BerriAI's open-source proxy](https://github.com/BerriAI/litellm)** — [PR #19943](https://github.com/BerriAI/litellm/pull/19943) re-stamps the client-requested alias onto `response.model`, with the deployment id available via a separate `x-*-model-id` response header. [Issue #22709](https://github.com/BerriAI/litellm/issues/22709) is an active disagreement over whether that choice broke downstream observability — maintainer admits *"we still need to decide what is the value that we actually want."* Equivalent to "option 1 + header."
- **OpenAI's own contract** — `"model": "gpt-4.1"` in → `"model": "gpt-4.1-2025-04-14"` out (resolved versioned id), neither pure alias nor pure raw ([API reference](https://platform.openai.com/docs/api-reference/chat/object)).

This PR picks **option 1 + header**: routing-group / alias name in `response.model`, target display name in the new `x-aisix-served-by` header. Rationale (per AISIX-Cloud#410 issue comment):

1. Direct-model behavior already echoes the customer-facing name (`dp-messages-anthropic-roundtrip-live.spec.ts:163` passes today); this PR closes the routing/direct inconsistency.
2. Failover-stable and provider-agnostic — customer dashboards don't flap, cross-provider mixes don't leak.
3. Header backstops observability — the lesson from the BerriAI proxy debate.

## Out of scope (follow-ups)

- Anthropic `/v1/messages` contract (filed against AISIX-Cloud#410 as a follow-up; same decision, separate code path in `messages.rs`).
- Other endpoints reaching `response.model`: `completions.rs`, `responses.rs`, `passthrough.rs` — none touched in this PR. Audit in a follow-up.

## Test plan

- [x] `cargo test -p aisix-proxy` — 273 unit tests pass, including two new contract tests in `render.rs` that pin `response.model` to the client-facing name even when the upstream returns a different id.
- [x] `cargo clippy -p aisix-proxy --all-targets -- -D warnings` clean
- [x] `cargo fmt --check` clean
- [ ] Companion AISIX-Cloud PR updates `dp-routing-failover-live.spec.ts` to assert `body.model === groupName` AND `response.headers.get('x-aisix-served-by') === healthyTargetName` after failover. Real-DP e2e validation happens in that PR.

## Independent audit

Per CLAUDE.md §8, an independent audit agent will review this PR with no shared context once it lands as a draft. All HIGH/MEDIUM findings will be addressed or explicitly justified before merge.

## Summary by CodeRabbit

* **New Features**
  * Added an x-aisix-served-by response header for routed requests to indicate which target served them; header is omitted for direct (non-routed) responses and cache hits, and omitted with a logged warning if the target name contains invalid header characters.

* **Bug Fixes**
  * Chat-completion outputs now consistently use the client-requested model name for both streaming chunks and non-streaming responses.

* **Documentation**
  * Added “Response Shape” docs and troubleshooting guidance describing model-field behavior, header presence, and interpretation of warnings and cache hits.

[Review Change Stack](https://app.coderabbit.ai/change-stack/api7/ai-gateway/pull/360?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/aisix-proxy/src/chat.rs
+++ b/crates/aisix-proxy/src/chat.rs
@@ -158,12 +158,31 @@ pub async fn chat_completions(
             // the request — see AISIX-Cloud#410. Only emitted when a
             // routing group was the entry point (direct models would
             // just echo `req.model`, which the body already carries).
+            //
+            // `HeaderValue::try_from` rejects CR/LF and non-visible
+            // ASCII (RFC 7230) — correct from a response-splitting
+            // standpoint, but if a routing target's `display_name`
+            // carries such bytes the header is silently absent and
+            // a customer debugging failover would see the same wire
+            // shape as a direct-model response. Surface the rejection
+            // in DP logs so operators can rename the offending target.
             if let Some(target) = success.served_by_target.as_deref() {
-                if let Ok(v) = axum::http::HeaderValue::try_from(target) {
-                    success
-                        .response
-                        .headers_mut()
-                        .insert("x-aisix-served-by", v);
+                match axum::http::HeaderValue::try_from(target) {
+                    Ok(v) => {
+                        success
+                            .response
+                            .headers_mut()
+                            .insert("x-aisix-served-by", v);
+                    }
+                    Err(err) => {
+                        tracing::warn!(
+                            target_display_name = %target,
+                            error = %err,
+                            "target display_name is not a valid HTTP header value; \
+                             omitting x-aisix-served-by — rename the target to use \
+                             only visible ASCII (no CR/LF, no non-ASCII characters)"
+                        );
+                    }
                 }
             }
             success.response
@@ -359,6 +378,68 @@ impl CacheStatus {
             CacheStatus::Miss => "miss",
             CacheStatus::Hit => "hit",
         }
+    }
+}
+
+/// Compute the value of [`Success::served_by_target`] for a request.
+///
+/// Centralises the "should we surface routing identity?" policy so
+/// the same rule applies on every dispatch branch (non-streaming
+/// success, streaming success, cache hit). The rule is:
+///
+/// - Routing-group request **and** an attempt won → `Some(target)`.
+/// - Anything else (direct model, no winner, cache hit) → `None`.
+///
+/// `None` is the wire signal "this was not a routing request" — the
+/// `x-aisix-served-by` response header is emitted only when the
+/// returned `Option` is `Some`, so its presence alone tells callers
+/// that failover routing fired. Direct-model responses must never
+/// carry the header.
+fn served_by_target_for_routing(
+    is_routing_request: bool,
+    served_target: Option<String>,
+) -> Option<String> {
+    if is_routing_request {
+        served_target
+    } else {
+        None
+    }
+}
+
+#[cfg(test)]
+mod served_by_target_tests {
+    use super::served_by_target_for_routing;
+
+    #[test]
+    fn direct_model_returns_none_even_when_a_target_was_captured() {
+        // A direct model still walks the attempt loop (with one
+        // candidate), so `served_target` may be `Some`. The routing
+        // flag is what gates header emission — direct models must
+        // never carry `x-aisix-served-by`.
+        assert_eq!(
+            served_by_target_for_routing(false, Some("gpt-4-direct".into())),
+            None,
+        );
+    }
+
+    #[test]
+    fn direct_model_with_no_target_is_none() {
+        assert_eq!(served_by_target_for_routing(false, None), None);
+    }
+
+    #[test]
+    fn routing_request_passes_winning_target_through() {
+        assert_eq!(
+            served_by_target_for_routing(true, Some("healthy-target".into())),
+            Some("healthy-target".into()),
+        );
+    }
+
+    #[test]
+    fn routing_request_with_no_winner_is_none() {
+        // All targets failed: header omitted. The caller is already
+        // surfacing an error response; routing identity is moot.
+        assert_eq!(served_by_target_for_routing(true, None), None);
     }
 }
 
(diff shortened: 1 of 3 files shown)


moonming:
## Audit reply (round 1)

Spawned a fresh independent audit agent against this PR per CLAUDE.md §8. Findings + responses:

**HIGH**: none.

**M1 — No in-tree assertion that `x-aisix-served-by` is omitted for direct models or cache hits.** Addressed in [5a2cb05](https://github.com/api7/ai-gateway/commit/5a2cb05):
- Extracted `served_by_target_for_routing(is_routing_request, served_target) -> Option<String>` so the same policy gates all three dispatch branches (non-streaming, streaming, cache hit).
- Four new unit tests pin the truth table: `(direct, Some) -> None`, `(direct, None) -> None`, `(routing, Some) -> Some`, `(routing, None) -> None`. A refactor flipping any branch surfaces as a unit-test failure.

**M2 — No in-tree assertion that `served_by_target` reflects the *winning* target after failover.** Not addressed in code, deferred to companion e2e:
- The auditor's concrete fix is to extract the attempt-chain loop (~100 lines in `chat.rs:864-961`) into a testable inner function with a mock `Bridge`. That refactor is reasonable but disproportionate to this PR's scope.
- The companion [api7/AISIX-Cloud#424](https://github.com/api7/AISIX-Cloud/pull/424) asserts `response.headers['x-aisix-served-by'] === healthyTargetName` after a 503-on-A failover. A refactor that hoisted the `chosen_target_display_name` capture above the `Ok(resp)` arm would capture target A and fail that e2e assertion.
- Filing M2's refactor as a follow-up issue rather than blocking this PR: the safety properties hold today (capture sits inside the success arm, sequential async loop, no concurrency), and the e2e gives the regression a backstop until the refactor lands.

**M3 — Display-name with non-visible-ASCII silently drops the routing signal.** Addressed in [5a2cb05](https://github.com/api7/ai-gateway/commit/5a2cb05):
- `HeaderValue::try_from` rejection (correct per RFC 7230 / response-splitting prevention) now emits a `tracing::warn!` carrying the offending target name and remediation hint. The rejection itself is preserved — header is still absent on invalid input.

**LOW**: the auditor noted slightly-misleading wording in the `served_by_target` struct field doc (the "set in that case below" bullet sits inside a "None in three cases" list). Not changed to keep this commit surgical to the actual findings; happy to clean up in a follow-up if reviewers want.

Independent re-audit of this amended commit will be spawned now.

moonming:
## Re-audit clean + CI lint failure is inherited

Independent re-audit of [5a2cb05](https://github.com/api7/ai-gateway/commit/5a2cb05) returned **0 HIGH / 0 MEDIUM**. Three LOW advisories (cosmetic: helper signature `Option<&str>` to avoid one clone, test-name length, log field/message wording consistency) — happy to address in follow-up if reviewers want.

**About the `lint (fmt + clippy)` CI failure**: it's inherited from main, not introduced by this PR. The fmt diff CI reports lives in `crates/aisix-etcd/src/loader.rs:445` — `provider_key_happy_path_accepts`, added in #362 — and `cargo fmt --all -- --check` (the exact CI command) reproduces the same diff on `origin/main` HEAD (`8d68d31`) untouched by this PR. Filed [#366](https://github.com/api7/ai-gateway/pull/366) as a 1-line surgical restore. Once #366 lands and this branch picks up main, the lint job here will go green.

Locally on this branch:
- `cargo fmt --all -- --check` clean (the loader.rs diff appears only when CI's merge-with-main introduces #362's regression)
- `cargo clippy -p aisix-proxy --all-targets -- -D warnings` clean
- `cargo test -p aisix-proxy` — 277/0 (273 baseline + 4 new contract tests)

PR #392, opened 2026-05-25
fix(platform-adapters): honor ProviderKey.api_base in Vertex + Azure bridges (#390, #391)

### LINKED ISSUE #390: Vertex bridge silently ignores ProviderKey.api_base in production builds
## Bug — Vertex bridge silently ignores `ProviderKey.api_base` in production builds

**Summary**: An operator who pastes `http://my-corporate-proxy/vertex` (or any custom hostname) into the BYO ProviderKey's `api_base` field for `adapter=vertex` will see their traffic still go to the canonical `https://<region>-aiplatform.googleapis.com`. The override is silently dropped — no error, no log warning, no UI feedback.

### Where

`crates/aisix-provider-vertex/src/bridge.rs`:

```rust
api_base_override: Option<String>,                          // line 54

pub(crate) fn with_api_base_override(mut self, url: impl Into<String>) -> Self {  // line 80
    self.api_base_override = Some(url.into());
    self
}

fn resolve_api_base(&self, region: &str) -> String {
    #[cfg(any(test, feature = "test-support"))]
    if let Some(b) = &self.api_base_override {
        return b.clone();
    }
    format!("https://{region}-aiplatform.googleapis.com")    // line 103
}
```

The `with_api_base_override` setter is `pub(crate)` — only callable from within the crate. Production registration in `crates/aisix-server/src/main.rs:795` uses `VertexBridge::new()` with no override. The bridge's `resolve_api_base()` never consults `ctx.provider_key.api_base`.

### Impact

- **Live e2e against `mock-vertex` is not possible.** The compose stack's `mock-vertex` service binds to `http://mock-vertex:8001` inside the bridge network, but a BYO PK with `api_base="http://mock-vertex:8001"` still hits `<region>-aiplatform.googleapis.com` (which 404s or times out depending on DNS).
- **Customer impact: BYO ProviderKey `api_base` is a documented field but has no effect for `adapter=vertex`.** A corporate-proxy or private-VPC operator setting it expects their traffic to be redirected — and is misled.
- **Audit failure mode: silent.** The DP doesn't log "ignoring api_base"; the operator sees their proxy receive zero traffic and traffic still landing on Google's endpoint.

### Compare to Bedrock

`crates/aisix-provider-bedrock/src/bridge.rs:597-603` reads `ctx.provider_key.api_base.as_deref()` and forwards it to the AWS SDK's `endpoint_url`:

```rust
let endpoint_url = {
    #[cfg(any(test, feature = "test-support"))]
    {
        self.endpoint_url_override
            .as_deref()
            .or(ctx.provider_key.api_base.as_deref())
    }
    #[cfg(not(any(test, feature = "test-support")))]
    {
        ctx.provider_key.api_base.as_deref()
    }
};
build_client(&creds, endpoint_url)
```

Bedrock honors `api_base` in production builds. Vertex should match that pattern.

### Suggested fix

Have `resolve_api_base()` consult `ctx.provider_key.api_base` as a production-grade override path. The test-only override seam can remain as a separate convenience (e.g. rename to `with_test_api_base_override` and gate on `#[cfg(test)]` only):

```rust
fn resolve_api_base(&self, region: &str, ctx_api_base: Option<&str>) -> String {
    #[cfg(test)]
    if let Some(b) = &self.api_base_override {
        return b.clone();
    }
    if let Some(b) = ctx_api_base.filter(|s| !s.is_empty()) {
        return b.to_string();
    }
    format!("https://{region}-aiplatform.googleapis.com")
}
```

Then callers pass `ctx.provider_key.api_base.as_deref()` from `chat_gemini` / `chat_gemini_stream`.

### Blocks

This blocks the per-adapter live e2e spec file `adapter-vertex-live.spec.ts` in AISIX-Cloud, which is the only way to assert the customer-visible contract end-to-end (request shape, OAuth header, response shape) against `mock-vertex` in CI.

### Surfaced via

AISIX-Cloud platform-adapter live-spec work (#302 Phase E follow-up). Discovered while writing `dashboard/tests/e2e/adapter-vertex-live.spec.ts` — the spec's `apiBase: 'http://mock-vertex:8001'` had no effect on the DP's outbound URL, and `mock-vertex`'s `/_internal/captured-requests` stayed empty.

### LINKED ISSUE #391: Azure-OpenAI bridge hardcodes .openai.azure.com host suffix, blocks BYO api_base override
## Bug — Azure-OpenAI bridge hardcodes `.openai.azure.com` host suffix, blocking BYO override

**Summary**: An operator setting `ProviderKey.api_base` for `adapter=azure-openai` is forced to use a hostname ending in `.openai.azure.com`. Any other host (corporate proxy, mock, private-VPC endpoint) fails resolution with a `Config` error before traffic ever leaves the DP. There is no production-grade override seam.

### Where

`crates/aisix-provider-azure-openai/src/bridge.rs::AzureUpstreamRef::resolve()` (line 205+):

```rust
let (host_resource, host_tail) = rest.split_once('.').ok_or_else(|| {
    BridgeError::Config(format!(
        "azure api_base {base:?} missing the .openai.azure.com suffix"
    ))
})?;
// ...
if host_tail_core != "openai.azure.com" {
    return Err(BridgeError::Config(format!(
        "azure api_base {base:?} host must end in .openai.azure.com \
         (got host suffix {host_tail_core:?})"
    )));
}
```

Test-only override seam at line 119:

```rust
#[cfg(test)]
pub(crate) fn with_url_override(mut self, url: impl Into<String>) -> Self {
    self.url_override = Some(url.into());
    self
}
```

The `url_override` field is `#[cfg(test)]` only and the setter is `pub(crate)`. No production code path bypasses the `.openai.azure.com` suffix check.

### Impact

- **Live e2e against `mock-llm`'s Azure deployment endpoint is not possible.** The compose stack's `mock-llm` service binds to `http://mock-llm:8000` inside the bridge network, and serves the Azure deployment URL shape `/openai/deployments/<deploy>/chat/completions`. A BYO PK with `api_base="http://mock-llm:8000"` is rejected at resolve time with `azure api_base "..." missing the .openai.azure.com suffix`.
- **Customer impact**: Operators running through corporate proxies (e.g. `https://azure-proxy.acme.internal/`) or private Azure Front Door routes cannot configure them via `api_base`. They need DNS / host-file hacks to make traffic round-trip.
- **Audit failure mode: explicit but blocking**. The `Config` error is at least visible — better than the silent Vertex case — but still wedges the BYO operator into a single host suffix.

### Compare to Bedrock

Bedrock honors arbitrary `api_base` via the SDK's `endpoint_url` parameter (no host-suffix check). See `crates/aisix-provider-bedrock/src/bridge.rs:597-603`.

### Suggested fix

Two-part approach:

1. **Default to `.openai.azure.com` resolution when no `api_base` is set**, exactly as today. This preserves the current "operator pastes just the resource name" UX path.
2. **When `api_base` IS set and the host doesn't end in `.openai.azure.com`, treat it as a production-grade override**: skip the resource-name extraction, use the host verbatim as the upstream base, and just append `/openai/deployments/<deployment>/chat/completions?api-version=<v>` to it.

```rust
pub fn resolve(deployment: &str, api_base: Option<&str>) -> Result<Self, BridgeError> {
    validate_url_token("deployment name", deployment)?;

    let base = api_base.unwrap_or_default().trim();
    if base.is_empty() {
        return Err(BridgeError::Config(
            "azure provider_key has no api_base — \
             expected https://<resource>.openai.azure.com or a bare resource name".into(),
        ));
    }

    // If base is a bare resource name or an Azure-canonical URL,
    // extract the resource as today. Otherwise treat as a verbatim
    // upstream override (corporate proxy / mock / private endpoint).
    if let Some(rest) = base.strip_prefix("https://").or_else(|| base.strip_prefix("http://")) {
        if let Some((host_resource, host_tail)) = rest.split_once('.') {
            let host_tail_core = host_tail.trim_end_matches('/').split_once('/').map(|(h,_)| h).unwrap_or(host_tail);
            if host_tail_core == "openai.azure.com" {
                return Ok(Self {
                    resource: host_resource.to_string(),
                    deployment: deployment.to_string(),
                    api_version: Self::DEFAULT_API_VERSION.to_string(),
                    upstream_override: None,
                });
            }
        }
        // Verbatim override path.
        return Ok(Self {
            resource: String::new(),
            deployment: deployment.to_string(),
            api_version: Self::DEFAULT_API_VERSION.to_string(),
            upstream_override: Some(base.to_string()),
        });
    }
    // Bare resource name shorthand — as today.
    validate_url_token("resource name", base)?;
    Ok(Self {
        resource: base.to_string(),
        deployment: deployment.to_string(),
        api_version: Self::DEFAULT_API_VERSION.to_string(),
        upstream_override: None,
    })
}

pub fn chat_completions_url(&self) -> String {
    let base = self.upstream_override.clone()
        .unwrap_or_else(|| format!("https://{}.openai.azure.com", self.resource));
    format!("{}/openai/deployments/{}/chat/completions?api-version={}",
            base, self.deployment, self.api_version)
}
```

### Also worth filing

Separately: the dashboard's `pk-secret` placeholder for `adapter=azure-openai` says `'azure-key-…'` which only covers the api-key path. The AAD path requires JSON `{tenant_id, client_id, client_secret}` — the placeholder should hint at both, or a separate UI affordance should let the operator pick the auth mode.

### Blocks

This blocks the per-adapter live e2e spec file `adapter-azure-live.spec.ts` in AISIX-Cloud.

### Surfaced via

AISIX-Cloud platform-adapter live-spec work (#302 Phase F follow-up). Discovered while writing `dashboard/tests/e2e/adapter-azure-live.spec.ts` — `createProviderKeyViaUI` with `byoAdapter: 'azure-openai'` and `apiBase: 'http://mock-llm:8000'` was rejected by cp-api → DP at PK projection time with the suffix error.

moonming:
## Summary

Two parallel production gaps where `ProviderKey.api_base` was silently dropped or rejected by the bridge, fixed in lockstep so the Bedrock parity story now covers all three platform adapters.

| Bridge | Pre-fix | Post-fix |
|:---|:---|:---|
| Bedrock | ✅ Already honored `api_base` via SDK `endpoint_url` (`bridge.rs:603`) | unchanged |
| **Vertex** | ❌ `resolve_api_base()` only consulted `#[cfg(test)]` `api_base_override`; production silently fell back to `<region>-aiplatform.googleapis.com` | ✅ Adds a `ctx_api_base` arg with precedence: cfg-test override > `ctx.provider_key.api_base` > canonical |
| **Azure** | ❌ `AzureUpstreamRef::resolve()` rejected any `api_base` whose host didn't end in `.openai.azure.com` | ✅ Adds an `upstream_override` field + verbatim-override branch for non-canonical hosts; canonical Azure path unchanged |

## Why now

Surfaced by AISIX-Cloud#491 — when writing per-adapter live e2e specs (`adapter-{azure,vertex,bedrock}-live.spec.ts`) against compose-stack mocks, the Bedrock spec worked because the bridge honored `api_base`, but Vertex + Azure could not redirect to `mock-vertex` / `mock-llm`. Filed as ai-gateway#390 + #391; this PR closes both.

## Defence-in-depth

The Azure override branch explicitly rejects `?` and `#` in `api_base` — the bridge appends `?api-version=…` so an operator-supplied query would open an api-version downgrade vector. The canonical Azure path keeps its existing strict resource-name validation.

The Vertex change drops nothing — the `#[cfg(test)] with_api_base_override` seam is preserved and takes precedence over `ctx.provider_key.api_base`, so existing tests keep working unchanged.

### Vertex (4 new unit tests in `bridge.rs::tests`)
- `resolve_api_base_honors_ctx_api_base_in_production_path` — pins the new production override path
- `resolve_api_base_trims_trailing_slash_on_ctx_override`
- `resolve_api_base_falls_back_to_canonical_on_empty_ctx_value` (empty string, whitespace, None)
- `resolve_api_base_cfg_test_override_takes_precedence_over_ctx` — proves wiremock-pinning tests stay intact

### Azure (5 unit-test changes in `bridge.rs::tests`)
- `resolve_accepts_verbatim_override_for_non_canonical_host` — replaces (renamed + flipped) the pre-fix `resolve_rejects_canonical_https_with_wrong_suffix` test, with an inline docstring citing #391
- `resolve_accepts_http_override_for_mock_endpoint` — pins the canonical mock-llm e2e shape
- `resolve_strips_trailing_slash_on_verbatim_override`
- `resolve_rejects_override_with_query_injection`
- `resolve_rejects_override_with_fragment`

All pre-existing tests (canonical Azure path, bare-resource shorthand, validators) pass unchanged.

## Test plan

- [x] `cargo test --workspace` — 1200+ tests pass, 0 failed
- [x] `cargo clippy -p aisix-provider-vertex -p aisix-provider-azure-openai --tests -- -D warnings` — clean
- [x] `cargo fmt` — applied
- [ ] CI on this PR
- [ ] Independent audit agent review
- [ ] After merge: cut the per-adapter live spec PRs on AISIX-Cloud for Vertex and Azure

## Closes

- Fixes #390 — Vertex bridge ignores `ProviderKey.api_base`
- Fixes #391 — Azure bridge hardcodes `.openai.azure.com` host suffix

## Unblocks

- AISIX-Cloud follow-up: `adapter-vertex-live.spec.ts` + `adapter-azure-live.spec.ts` (per-adapter live e2e specs, mirror of the merged `adapter-bedrock-live.spec.ts` in AISIX-Cloud#491)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed API base URL resolution for the Vertex provider so operator-supplied api_base overrides are respected in production.

* **New Features**
  * Azure OpenAI provider now supports operator-supplied upstream URL overrides (including HTTP), with trailing-slash trimming.

* **Validation & Tests**
  * Added validation to reject overrides containing query strings/fragments and added unit and end-to-end tests to cover override behavior.

* **Documentation**
  * Expanded google-vertex provider docs: api_base format, URL construction, streaming vs non-streaming endpoints, and OAuth token guidance.

[Review Change Stack](https://app.coderabbit.ai/change-stack/api7/ai-gateway/pull/392?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


moonming:
Audit findings addressed in fe13041.

**HIGH-1 (streaming-path E2E gap):** added `chat_gemini_stream_honors_provider_key_api_base_in_production_path` — mirrors the non-streaming wiremock test but drains the SSE stream via `bridge.chat_stream`. A regression that reverted just one of the two call sites would now fail.

**MEDIUM-1 (api_base SSRF / credential-embed surface):** before #390 the production code path hardcoded the canonical Google URL — there was no operator-supplied URL surface. The fix opened that surface; the original commit forgot to validate. `resolve_api_base` now returns `Result<String, BridgeError>` and rejects 4 classes of input as `BridgeError::Config`:

- non-http(s) scheme (`file:///etc/passwd`, `gs://...`, bare host)
- userinfo `@` (would leak credentials via access-log URLs + auth-header shadow)
- query string `?` (would silently merge with bridge's `?alt=sse` on streaming)
- fragment `#`

Mirrors the Azure-OpenAI sibling fix (#391) which rejects the same classes. Surfacing as Config errors (vs silently falling through to canonical) matches the principle of #390 itself: don't silently ignore operator config.

Five new unit tests cover the four rejection classes:
- `resolve_api_base_rejects_non_http_scheme` (5 cases incl. bare host)
- `resolve_api_base_rejects_embedded_userinfo`
- `resolve_api_base_rejects_query_string`
- `resolve_api_base_rejects_fragment`

LOW-1 and LOW-2 from the audit are non-blocking; no action.

Test run: 72/72 green (added 5 tests, was 67/67 pre-audit). `cargo clippy --all-targets -- -D warnings` clean.

moonming:
Re-audit LOW-1 addressed in 46f6997.

The userinfo-rejection branch's error message echoed the operator-supplied URL verbatim via `{b:?}` — so a rejected `https://user:secret@host` produced a Config error containing `user:secret@host`. Since the entire point of rejecting `@` is to keep operator credentials out of upstream traffic, the error shouldn't leak them via the audit/telemetry path either.

Added a small `redact_userinfo` helper that replaces `scheme://user:pass@host` with `scheme://<redacted>@host`. Preserves RFC 3986 path-`@` (e.g. `https://proxy/v1@my-namespace` stays unchanged because the `@` is past the first path slash).

74/74 Vertex tests green. The re-audit's LOW-2 note (`b.contains('@')` over-rejecting RFC 3986 path-`@`) is operationally moot for Vertex's bare-host-root convention; not changed.

PR #397, opened 2026-05-26
fix(embeddings): accept base64 vectors from upstream — #393 customer-blocker

### LINKED ISSUE #393: bug: /v1/embeddings returns 502 (upstream_decode_error) decoding base64 embedding responses (OpenAI SDK default)
## Summary

`/v1/embeddings` returns `502 upstream_decode_error` when the upstream returns **base64-encoded** vectors. The OpenAI SDKs request `encoding_format: base64` by default, so a plain `client.embeddings.create({ model, input })` through the DP always fails; `encoding_format: float` works. The upstream response is valid — the DP cannot decode it.

## Repro (source-blind; from AISIX-Cloud OpenAI real-chain manual testing, 2026-05-26)

1. Dashboard → Provider Keys → New → **OpenAI** (`api_base https://api.openai.com/v1`, real key).
2. Create a Model with upstream id `text-embedding-3-small`; issue an API key; start a real DP.
3. POST `/v1/embeddings` through the DP, `float` then `base64`:

```bash
curl -s -w "\nHTTP %{http_code}\n" $DP/v1/embeddings -H "Authorization: Bearer $AISIX_KEY" -H "Content-Type: application/json" -d '{"model":"<alias>","input":"hello","encoding_format":"float"}'
curl -s -w "\nHTTP %{http_code}\n" $DP/v1/embeddings -H "Authorization: Bearer $AISIX_KEY" -H "Content-Type: application/json" -d '{"model":"<alias>","input":"hello","encoding_format":"base64"}'
```

**Expected**: both `200`. `base64` is a valid, documented OpenAI format (the vector is returned as a base64 string), and the SDK requests it by default.

**Actual**: `float` → `200` (1536-dim vector); `base64` → `502 {"error":{"message":"upstream returned an unparseable body: error decoding response body","type":"upstream_decode_error"}}`. SDK default (`embeddings.create({ model, input })`, no `encoding_format`) → `502`.

## Confirmed by manual testing

Same model + DP via curl / python / OpenAI SDK: the failure tracks the `encoding_format` value, not the client (`float` 200, `base64` 502 in all three). DP access log tags the failed request `provider="unknown"` (vs `provider="openai"` on `float`) with ~450–600 ms latency — the upstream round-trip completed and the DP fails **decoding the response**. Hitting `api.openai.com` directly with `base64` works; upstream is healthy.

## Why this matters

The OpenAI SDK **default** embeddings path is unusable through the gateway — most callers never set `encoding_format`, so they get a 502. LiteLLM and Portkey both proxy base64 embeddings transparently. `encoding_format: "float"` is a client-side workaround, not a fix.

## Fix scope

Source-blind hypothesis (not yet audited): the OpenAI-provider embeddings response deserializer likely types `data[].embedding` as a float array, so a base64 **string** fails to deserialize → `upstream_decode_error` / 502, with provider classification falling back to `unknown`. Preferred fix: accept both shapes for `data[].embedding` (float array or base64 string) and round-trip verbatim, preserving the caller's `encoding_format`.

## Related

- AISIX-Cloud OpenAI real-chain coverage (pending PR) records `/v1/embeddings` as a soft probe behind this gap.

moonming:
## Summary

Closes #393. The gateway-internal type for the upstream embeddings response was `embedding: Vec<f32>` (strict-typed). When upstream returned a base64 string — the OpenAI SDK default — the deserializer rejected it and the gateway 502'd. Every vanilla `client.embeddings.create({ model, input })` call was broken.

## Fix

Add an untagged enum `EmbeddingVector { Float(Vec<f32>) | Base64(String) }` so the gateway accepts and round-trips both shapes per OpenAI's documented `string | array` union (<https://platform.openai.com/docs/api-reference/embeddings/object>). Pass-through, no internal translation — customers who chose base64 for payload-size reasons see the same bytes the upstream returned.

## Test coverage

- **Unit** (`crates/aisix-provider-openai/src/wire.rs`):
  - `embeddings_response_accepts_float_array`
  - `embeddings_response_accepts_base64_string` ← the #393 regression
  - `embeddings_response_serializes_back_to_same_shape`
- **E2E** (`tests/e2e/src/cases/openai-embeddings-base64-e2e.test.ts`):
  3 source-blind subtests:
  1. `encoding_format=float` → array round-trip
  2. `encoding_format=base64` → base64 string round-trip
  3. **SDK default** (no `encoding_format` argument) → base64 string round-trip (the customer-facing path #393 documents as broken)

## Why pass-through, not normalize-to-Vec<f32>

- Customers set `encoding_format: "base64"` for payload-size — re-encoding defeats the intent.
- The gateway forwards `encoding_format` verbatim to the upstream, so upstream-vs-client format mismatch never happens in practice.
- Matches the pass-through behavior of comparable open-source AI gateways and routing libraries (verified via internal triage of reference implementations cited in CLAUDE.md §7).
- Simpler — no base64 encode/decode hot path.

## Test plan

- [x] `cargo build --all` clean.
- [x] `cargo test -p aisix-provider-openai --lib wire::tests::embeddings` — 3/3 pass.
- [x] `pnpm test cases/openai-embeddings-base64-e2e` — 3/3 pass in 3.8s.
- [ ] CI green.

## Closes

- Closes #393

### DIFF
--- a/crates/aisix-gateway/src/chat.rs
+++ b/crates/aisix-gateway/src/chat.rs
@@ -368,12 +368,38 @@ pub struct ChatDelta {
 
 // ─── Embeddings ──────────────────────────────────────────────────────────────
 
+/// The vector returned on one embedding object. Untagged enum so JSON
+/// round-trips OpenAI's documented `string | array` shape on the
+/// `embedding` field (issue #393):
+///
+///   - `Float(vec![0.1, 0.2, ...])` → JSON array of numbers; what
+///     OpenAI returns when the request carries
+///     `encoding_format: "float"`.
+///   - `Base64("BASE64STRING")` → JSON string; what OpenAI returns
+///     when the request carries `encoding_format: "base64"` (the
+///     SDK default). Stored verbatim — the gateway is a pure
+///     pass-through for this field so callers who chose `base64`
+///     for payload-size reasons see the same bytes the upstream
+///     returned.
+///
+/// The gateway does NOT translate between the two formats. If a
+/// future caller needs cross-format translation, that belongs at
+/// the dispatcher, not the wire layer.
+///
+/// Reference: <https://platform.openai.com/docs/api-reference/embeddings/object>.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum EmbeddingVector {
+    Float(Vec<f32>),
+    Base64(String),
+}
+
 /// Single embedding object as returned by a provider.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct EmbeddingObject {
     pub index: u32,
     pub object: String,
-    pub embedding: Vec<f32>,
+    pub embedding: EmbeddingVector,
 }
 
 /// Normalised embedding request.
--- a/crates/aisix-gateway/src/lib.rs
+++ b/crates/aisix-gateway/src/lib.rs
@@ -32,7 +32,7 @@ pub use bridge::{
 };
 pub use chat::{
     ChatChunk, ChatDelta, ChatFormat, ChatMessage, ChatResponse, EmbeddingObject, EmbeddingRequest,
-    EmbeddingResponse, EmbeddingUsage, FinishReason, Role, UsageStats,
+    EmbeddingResponse, EmbeddingUsage, EmbeddingVector, FinishReason, Role, UsageStats,
 };
 pub use hub::Hub;
 pub use sse::{SseDecoder, SseEvent};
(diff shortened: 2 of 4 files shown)


moonming:
Audit findings addressed:

- **MEDIUM-1 (naming-rule LiteLLM mention)** — redacted from PR body.
- **LOW-1 (batch-input e2e)** — filed as ai-gateway#398, follow-up.
- **LOW-2 (null embedding still rejected)** — filed as ai-gateway#398 along with the batch case; hardening, not regression.

Other audit angles (correctness, pass-through semantics, no breaking changes, SDK default claim, source-blind discipline, reliability) all verified clean. Ready to merge once CI green.

Issue #336, opened 2026-05-18
/v1/messages returns OpenAI-shape error envelope; Anthropic SDKs expect {type:'error', error:{type, message}}


moonming:
## Repro (source-blind, from AISIX-Cloud #302 D3.3 E2E coverage)

1. Dashboard → Provider Keys → New → Anthropic. Save.
2. Create a Model on this PK with upstream id including `force-status-429` (mock-llm convention — same force-status semantics as the chat handler).
3. Issue API key, POST `/v1/messages` to the DP.

**Expected**: response body is Anthropic-shape `{type: "error", error: {type, message}}` per [Anthropic API error docs](https://docs.anthropic.com/en/api/errors). The Claude SDK reads `body.type === "error"` as the top-level discriminator — branching on `body.error.message` is keyed on this check.

**Actual**: response body is OpenAI-shape `{error: {message, type, code, param}}` regardless of the inbound endpoint. The top-level `type` field is absent, breaking Claude SDK error parsing.

## Confirmed by audit

- `crates/aisix-proxy/src/error.rs:1-179` — `ProxyError::IntoResponse` always serializes `ErrorEnvelope { error: ErrorBody { message, kind, param, code } }`. Module docstring explicitly says "OpenAI-compatible error envelope used by every proxy endpoint" — no Anthropic-shape branch.
- `messages.rs:260` wraps upstream errors as `ProxyError::Bridge(err)`, which falls through `IntoResponse` to OpenAI shape.

## Impact

Every Claude SDK customer routed through AISIX who hits any upstream error (rate limit, content filter, model not found, 5xx) sees an envelope shape Claude SDK can't parse. The SDK's error class hierarchy depends on `body.type === "error"` for the discriminator branch — without it, the SDK falls back to a generic exception with the entire body dumped to the message field, losing the structured error context that drives retry / fallback logic.

Note: `error.type` IS normalized to DP taxonomy (`upstream_error`) which is correct per ai-gateway#327 — but it's nested under the wrong envelope shape.

## Fix scope (DP side, not implementing here)

Two cleanest approaches:

(1) **Per-endpoint envelope** — branch `IntoResponse` on a request-scoped "inbound protocol" hint (set in `messages.rs` and `chat.rs` dispatchers). Anthropic shape:

\`\`\`json
{
  "type": "error",
  "error": {
    "type": "upstream_error",
    "message": "..."
  }
}
\`\`\`

[...]

Issue #159, opened 2026-05-09
bug: oversize request body resets connection (ECONNRESET) instead of returning 413 Content Too Large


moonming:
## Summary

When a caller sends a request body that exceeds the gateway's configured \`request_body_limit_bytes\` (10 MiB by harness default, mirroring production), the gateway closes the TCP connection mid-write rather than emitting a clean \`413 Content Too Large\` response.

The caller's HTTP client surfaces this as \`ECONNRESET\`, which is indistinguishable from a network failure or a gateway crash. Per [RFC 9110 §15.5.14](https://datatracker.ietf.org/doc/html/rfc9110#section-15.5.14), the correct response for a body that exceeds a server-imposed limit is \`413 Content Too Large\`.

## Repro

E2E test attached: drafted as part of \`tests/e2e/src/cases/body-edges-e2e.test.ts\` (held back from #151's C6 PR pending this fix).

\`\`\`ts
const filler = "x".repeat(10 * 1024 * 1024 + 512 * 1024); // 10.5 MiB
const body = JSON.stringify({
  model: "body-edges",
  messages: [{ role: "user", content: filler }],
});
const res = await fetch(\`\${app.proxyUrl}/v1/chat/completions\`, {
  method: "POST",
  headers: {
    authorization: \`Bearer \${CALLER_PLAINTEXT}\`,
    "content-type": "application/json",
  },
  body,
});
\`\`\`

Observed: \`fetch failed: write ECONNRESET\` — the gateway closed the socket mid-stream.

Expected:
\`\`\`
HTTP/1.1 413 Content Too Large
Content-Type: application/json
Content-Length: ...

{"error":{"message":"request body exceeds 10485760 bytes","type":"...","code":"..."}}
\`\`\`

## Why this matters

- A clean \`413\` tells the caller the request is invalid because of size — they shouldn't retry without changing the payload.
- \`ECONNRESET\` is ambiguous: callers commonly retry on connection failures (assuming transient network issues), making the gateway look like it's hanging or flaky.
- Operator debugging: distinguishing "client sent oversized body" from "upstream timed out" from "gateway crashed" is much harder when all three look the same on the wire.

## Proposed fix

If the body-size limit is enforced via Axum / Tower middleware (\`axum::extract::DefaultBodyLimit\` or similar), check Content-Length **before** beginning to consume the body. When Content-Length exceeds the limit, return \`413\` with an OpenAI-shape error envelope and close the connection cleanly.

[...]

PR #400, opened 2026-05-26
fix(chat): missing required field returns 400 not 422 (#324)

### LINKED ISSUE #324: bug(dp): missing-`model`-field surfaces as 422, not OpenAI-compat 400
## Summary

Phase 3 of the E2E matrix series (#332) added a scenario asserting that a valid-JSON chat request missing the required \`model\` field surfaces as **400 Bad Request** (matching OpenAI's wire contract for missing required fields). The DP returns 422 (Unprocessable Entity) instead.

## Repro

Phase 3 spec, scenario \"valid JSON but missing model field → 400 (canonical missing-field)\":

\`\`\`
expect: 400
received: 422
> 305 |       ).toBe(400);
\`\`\`

Request body:
\`\`\`json
{
  \"messages\": [{\"role\": \"user\", \"content\": \"hi\"}]
}
\`\`\`
(no \`model\` field)

## Customer impact

OpenAI's official API returns **400** for missing required fields. SDKs that branch on status code (e.g., \"400 = fix your input and retry\" vs \"422 = re-render the form\") see different semantics depending on which gateway / proxy they sit behind. A customer migrating from OpenAI direct to AISIX (or the inverse) silently gets a different error code surface, which breaks tooling pinned to one or the other.

422 is technically valid HTTP semantics — RFC 4918 §11.2 \"The server understands the content type of the request entity ... but was unable to process the contained instructions.\" Some frameworks (FastAPI, Spring Validation) default to 422 for schema-validation failures. So this isn't a bug in HTTP-spec terms, but it IS a wire-compat divergence from the OpenAI API contract the DP advertises.

## Expected behavior

Match OpenAI's wire contract: return **400** for missing required fields, **400** for malformed JSON, reserve **422** (if used at all) for semantic-but-unparseable cases the spec hasn't pinned. Same applies to other required fields (\`messages\`, etc.).

## Test impact

Phase 3 matrix scenario currently fails on the strict 400 assertion. To unblock Phase 3 merge, the assertion will be loosened to accept both 400 and 422 with a comment + issue link pointing here; once this bug is fixed, the assertion gets tightened to 400 only.

moonming:
## Summary

Closes #324. Missing required field (\`model\`, \`messages\`, etc.) on \`/v1/chat/completions\` surfaced as 422 instead of OpenAI's documented 400. SDKs branching on the status code saw different semantics between OpenAI-direct and gateway-fronted deployments.

## Fix

Mirror the messages.rs (#336) pattern: change handler signature to \`Result<Json<ChatFormat>, JsonRejection>\` and intercept the rejection in the handler, mapping to \`ProxyError::InvalidRequest\` (→ 400 + \`invalid_request_error\` per OpenAI's documented envelope).

\`BytesRejection\` keeps its existing 413/400 split (#159 + messages.rs audit MEDIUM-A/B); the fallback \`_\` arm maps JsonDataError / JsonSyntaxError / MissingJsonContentType / any future axum variant to 400 invalid_request_error.

## Coverage

- \`missing_model_field_returns_400_not_422\` — the issue's exact repro (Phase 3 matrix scenario \"valid JSON but missing model field\").
- \`missing_messages_field_returns_400_not_422\` — companion required-field case.
- \`malformed_json_returns_400\` — JsonSyntaxError variant pinned.

## Scoped out (file as follow-up)

\`embeddings.rs\` uses \`Json<EmbeddingRequestBody>\` (also strict-typed) so it ships the same bug. Same one-line fix applies. \`completions.rs\` / \`responses.rs\` / \`rerank.rs\` / \`images.rs\` / \`audio.rs\` use \`Json<Value>\` so they don't surface the bug. Will file as follow-up.

## Test plan

- [x] \`cargo test -p aisix-proxy --lib\` — 302/302 green locally including the 3 new tests.
- [ ] CI green.

## Closes

- Closes #324

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error responses for chat completions requests with invalid or malformed payloads.

* **Tests**
  * Added regression tests validating proper error handling for malformed JSON and missing required fields in chat requests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/api7/ai-gateway/pull/400?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/aisix-proxy/src/chat.rs
+++ b/crates/aisix-proxy/src/chat.rs
@@ -125,11 +125,47 @@ use crate::cooldown::decide_cooldown;
 pub async fn chat_completions(
     State(state): State<ProxyState>,
     auth: AuthenticatedKey,
-    Json(req): Json<ChatFormat>,
+    // Issue #324: catch the JSON-extractor rejection here so we can
+    // map it to OpenAI's documented 400 invalid_request_error wire
+    // shape. Axum's `Json<T>` extractor returns 422 on JsonDataError
+    // (valid-JSON-but-missing-required-field, e.g. no `model`),
+    // which diverges from OpenAI — every SDK that branches on 400
+    // vs 422 sees different semantics here than it does talking to
+    // api.openai.com. Same discriminate-then-map pattern as
+    // messages.rs (#336) which already handles this for /v1/messages.
+    body: Result<Json<ChatFormat>, axum::extract::rejection::JsonRejection>,
 ) -> Response {
     let started = Instant::now();
     let method = "POST";
     let path = "/v1/chat/completions";
+    let req = match body {
+        Ok(Json(r)) => r,
+        Err(rej) => {
+            use axum::extract::rejection::JsonRejection;
+            use axum::http::StatusCode;
+            // BytesRejection → distinguish 413 (PAYLOAD_TOO_LARGE,
+            // real per-extractor cap exceeded) from 400 (transport-
+            // side read failure). `JsonRejection` is `#[non_exhaustive]`
+            // so the fallback `_` arm catches today's JsonDataError
+            // (the #324 case) / JsonSyntaxError / MissingJsonContentType
+            // AND any future variant axum adds, defaulting to 400
+            // until each new variant gets an explicit policy decision.
+            return match rej {
+                JsonRejection::BytesRejection(inner)
+                    if inner.status() == StatusCode::PAYLOAD_TOO_LARGE =>
+                {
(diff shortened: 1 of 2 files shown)


moonming:
Audit MEDIUM addressed in 2edbffb. Both `missing_messages_field_returns_400_not_422` and `malformed_json_returns_400` now pin `error.type == "invalid_request_error"` on the envelope, symmetric with the original `missing_model_field_returns_400_not_422` test. All 3 JsonRejection-variant tests now pin status + envelope-shape.

PR #409, opened 2026-05-26
test(e2e): drive gateway through real Claude Code CLI

moonming:
## Summary

- Adds `tests/e2e/src/cases/claude-code-cli-compat-e2e.test.ts` — drives the gateway's Anthropic-native `/v1/messages` endpoint through the real `claude -p` CLI, with `ANTHROPIC_BASE_URL` pointed at the test gateway and a real non-Anthropic upstream (OpenAI or DeepSeek) behind it.
- Adds `.github/workflows/real-chain.yml` — PR-label-gated workflow modeled on AISIX-Cloud's `real-chain.yml`. Validates secrets, builds the binary, installs the Claude Code CLI via npm, runs only this new test.

## Why this matters

`openai-sdk-compat.test.ts` already proves the gateway speaks OpenAI Chat Completions to OpenAI-shape clients. This sibling proves it speaks Anthropic Messages to the wire shape an Anthropic-ecosystem client (Claude Code CLI, Claude Desktop, anthropic-sdk-{python,typescript}) actually emits — which the hand-rolled harness does not generate.

A `claude -p` invocation posts ~17KB to `/v1/messages`:
- structured `system` array with `cache_control: { type: "ephemeral" }` markers
- a tool catalogue of ~30 entries in Anthropic `input_schema` shape
- `anthropic-beta` header (`prompt-caching-scope-…,claude-code-…,advisor-tool-…`)
- `?beta=true` query suffix
- `x-api-key` auth header (Anthropic shape, not `Authorization: Bearer`)

A regression in any of:
- `/v1/messages?…` query-string tolerance
- `x-api-key` auth fallback
- `anthropic-beta` / `anthropic-version` header passthrough vs. strip
- `system: [{type:"text", text, cache_control?}]` block parsing
- `tools[].input_schema` → upstream `tools[].function.parameters` translation for non-Anthropic upstreams

would break Claude Code customers but pass every existing test.

## Gating

Test default-skips unless **all** of:
- `RUN_CLAUDE_CODE_REAL_CHAIN=1` (acknowledges real-upstream spend)
- `OPENAI_API_KEY` **or** `DEEPSEEK_API_KEY` (upstream credential)
- `claude` CLI on `PATH`

The new workflow gates on the `real chain test` PR label + the in-repo fork guard, matching AISIX-Cloud's pattern. PRs without the label do not trigger this workflow.

## References

- Anthropic Messages API spec: https://docs.anthropic.com/en/api/messages
- Anthropic Node SDK (the wire shape Claude Code uses): https://github.com/anthropics/anthropic-sdk-typescript
- Claude Code CLI setup (npm install path): https://code.claude.com/docs/en/setup
- \`--bare\` flag (strips OAuth/keychain so auth is strictly \`\$ANTHROPIC_API_KEY\`): \`claude --help\`
- Companion DP-side PR: https://github.com/api7/AISIX-Cloud/pull/508

## Test plan

- [ ] \`real-chain\` workflow turns green on this PR (label triggers it)
- [ ] On a separate PR that breaks \`x-api-key\` auth, this test fails before merge
- [ ] On a separate PR that breaks Anthropic→OpenAI tool translation, this test fails before merge
- [ ] Without \`RUN_CLAUDE_CODE_REAL_CHAIN=1\`, the default \`pnpm test\` matrix still skips this case cleanly

## Summary by CodeRabbit

* **Tests**
  * Added an opt-in end-to-end test that exercises the CLI against a configurable live upstream to validate request translation, authentication, and tool/function shaping.

* **Chores**
  * Added an automated workflow to run these compatibility checks on schedule, via manual dispatch, and on pull requests, with required API key validation and environment setup for reliable E2E runs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/api7/ai-gateway/pull/409?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/real-chain.yml
+++ b/.github/workflows/real-chain.yml
@@ -24,8 +24,11 @@ permissions:
   contents: read
 
 concurrency:
+  # Cancel superseded runs — every `labeled` event re-triggers the
+  # workflow and each run burns real upstream quota; without
+  # cancellation a labeling spree stacks costs fast.
   group: real-chain-${{ github.ref }}
-  cancel-in-progress: false
+  cancel-in-progress: true
 
 jobs:
   real-chain:
@@ -34,8 +37,11 @@ jobs:
     timeout-minutes: 30
     # Same fork-guard + label-gate shape as AISIX-Cloud's real-chain
     # workflow: secrets only flow on PRs from the same repo, and only
-    # when an opt-in label is set.
-    if: ${{ github.event_name != 'pull_request' || (github.event.pull_request.head.repo.full_name == github.repository && contains(github.event.pull_request.labels.*.name, 'real chain test')) }}
+    # when an opt-in label is set. The `action != 'labeled'` clause
+    # narrows the label trigger to the relevant label name — otherwise
+    # *any* label change on a PR that already carries the gating label
+    # re-runs the workflow.
+    if: ${{ github.event_name != 'pull_request' || (github.event.pull_request.head.repo.full_name == github.repository && contains(github.event.pull_request.labels.*.name, 'real chain test') && (github.event.action != 'labeled' || github.event.label.name == 'real chain test')) }}
     services:
       etcd:
         image: quay.io/coreos/etcd:v3.5.15
@@ -91,5 +97,11 @@ jobs:
           AISIX_BIN: ${{ github.workspace }}/target/debug/aisix
           AISIX_E2E_ETCD: http://127.0.0.1:2379
           RUN_CLAUDE_CODE_REAL_CHAIN: "1"
+          # Pin upstream deterministically — without REAL_CHAIN_UPSTREAM
+          # the test's pickUpstream() iterates UPSTREAMS in array order
+          # (OpenAI first), and if a future runner had both keys set
+          # the upstream would silently change across runs, making
+          # regression triage harder.
+          REAL_CHAIN_UPSTREAM: deepseek
           DEEPSEEK_API_KEY: ${{ secrets.DEEPSEEK_API_KEY }}
         run: pnpm exec vitest run src/cases/claude-code-cli-compat-e2e.test.ts
--- a/.github/workflows/real-chain.yml
+++ b/.github/workflows/real-chain.yml
@@ -1,0 +1,95 @@
+# Scheduled real-chain coverage.
+#
+# Drives the gateway's Anthropic-native /v1/messages endpoint through
+# the REAL Claude Code CLI (`claude -p`) against a real non-Anthropic
+# upstream (DeepSeek by default — the credential available in CI).
+# Catches regressions in cross-provider translation of the wire shape
+# that Anthropic-ecosystem clients actually emit but the in-repo
+# vitest mock-upstream tests do not generate.
+#
+# Gated separately from PR CI for the same reason as AISIX-Cloud's
+# real-chain.yml: spends real upstream quota.
+
+name: real chain
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled]
+  schedule:
+    # 08:00 Asia/Shanghai every day.
+    - cron: "0 0 * * *"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: real-chain-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  real-chain:
+    name: claude-code-cli real-chain
+    runs-on: ubicloud-standard-2
+    timeout-minutes: 30
+    # Same fork-guard + label-gate shape as AISIX-Cloud's real-chain
+    # workflow: secrets only flow on PRs from the same repo, and only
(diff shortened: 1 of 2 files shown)


moonming:
## Independent audit — Round 2

First-round HIGH findings (no wire-shape pinning, unverified CLI flags, skip-vs-throw in CI mode) are resolved. The recording reverse-proxy correctly captures the gateway's translated outbound request, the flag probe fails loud when required flags are missing, and prerequisite gaps in CI mode now throw instead of silently skipping.

### Round-2 audit BLOCK — addressed in commit \`ab68d50\`

**H1-r2** — \`pickUpstream()\` silently fell back to a different provider when \`REAL_CHAIN_UPSTREAM\`'s pinned key was missing, defeating the determinism the env var was added for in the previous round.

  Fixed: throws when the pin names an unknown provider or when the pin's required env var is absent.

**M4-r2** (recommended before merge) — the prompt round-trip assertion \`JSON.stringify(userMsgs).toLowerCase().includes(\"ready\")\` was brittle: false-passes if Claude Code's own catalogue contains \"ready\", false-fails with a misleading message if a future bridge moved the prompt into \`role=system\`.

  Fixed: assert \`role=user\` separately, then assert the literal \`PROMPT\` round-trips verbatim. Each assertion message pins the specific failure mode it catches.

**L1-r2** — dead \`etcdReachable\` local removed.

### Remaining Round-2 MEDIUMs — filed as follow-up

The audit raised four MEDIUM items that touch the recording-proxy implementation more invasively (buffer accumulation, word-boundary flag matching, relay timeout + client-disconnect propagation, workflow guard against accidental upstream change). Filed as **#410 — \"test/e2e: recording-proxy and CI hardening for the claude-code real-chain spec\"** with concrete code suggestions for each.

Per CLAUDE.md §8, MEDIUMs are explicitly justified via #410 rather than blocking this PR's merge.

### Round-3 audit incoming

A fresh independent audit on the new commit will follow.

moonming:
## Independent audit — Round 3 verdict: **MERGE**

All three fixes from commit \`ab68d50\` confirmed correct by the fresh audit:

- **H1-r2** (pickUpstream hard-fail on bad pin) — throw flow is sound; \`beforeAll\` is the right call site, module load + \`pnpm test\` skip cleanly, CI-mode misconfiguration fails with the descriptive error.
- **M4-r2** (literal PROMPT round-trip) — \`PROMPT\` is a constant flowing to both \`spawnSync\` argv and the assertion. Both string and typed-array content shapes are handled. False-pass paths (different field, base64 encoding, role demotion) all fail with the correct message.
- **L1-r2** (dead var removal) — surgical, no remaining references.

Deferral of the four Round-2 MEDIUMs to **#410** confirmed legitimate — none escalate to HIGH on closer inspection. The Round-3 audit surfaced one additional LOW (JSON-escape brittleness in user-content extraction); folded into #410.

No new regressions introduced by Round-2 or Round-3 commits.

Per CLAUDE.md §8, this PR is unblocked for human review and merge.

moonming:
## CI Round-3 — \`claude-code-cli real-chain\` workflow failed at \`Validate required secrets\`

\`\`\`
##[error]DEEPSEEK_API_KEY is required (the test gates on at least one upstream credential)
\`\`\`

\`gh secret list -R api7/ai-gateway\` returns empty; this repo doesn't have \`DEEPSEEK_API_KEY\` configured. The companion AISIX-Cloud repo does (\`ANTHROPIC_API_KEY\`, \`DEEPSEEK_API_KEY\`, \`OPENAI_API_KEY\`), which is why #508's real-chain step ran successfully.

### Required from a maintainer

Add one or both of these as repo-level secrets at https://github.com/api7/ai-gateway/settings/secrets/actions:

- \`DEEPSEEK_API_KEY\` (matches \`REAL_CHAIN_UPSTREAM=deepseek\` pinned in the workflow)
- Or \`OPENAI_API_KEY\` (if you'd rather pin \`REAL_CHAIN_UPSTREAM=openai\`)

The same secret can be the value from AISIX-Cloud's existing secret of the same name — they bill against the same upstream account.

Once the secret is in place, push an empty commit (or re-trigger via \`gh pr edit 409 --add-label \"real chain test\"\` after toggling the label) and the workflow will pick up the secret on the next run.

Everything else CI-side (lint, schema-drift, unit, build, e2e, coverage, code-rabbit) is green.

moonming:
撤回 — 这条 PR 跑 real-chain 需要 \`DEEPSEEK_API_KEY\` 之类的 secret，但 api7/ai-gateway 这个 repo 没配（secrets 只在 api7/AISIX-Cloud）。

覆盖同一个契约的 sibling PR api7/AISIX-Cloud#508 在 AISIX-Cloud 那边已经端到端跑通：真 \`claude -p\` → live AISIX DP → /v1/messages → 真 DeepSeek 上游 → Anthropic 响应回到 CLI。\`real-chain\` workflow 上一轮 ✅。

关联 follow-up issue #410 一起关。

PR #434, opened 2026-05-29
feat(azure-openai): production AAD authority_host override seam (closes #413)

### LINKED ISSUE #413: Azure AAD: no production-grade token endpoint override (blocks live e2e against mock)
## Bug — same class as #390/#391

\`crates/aisix-provider-azure-openai/src/aad_token_mint.rs:140-160\` declares \`token_endpoint_override: Option<String>\` BUT both the field and its setter \`with_token_endpoint_override\` are \`#[cfg(test)]\` only. Production builds always POST to the hardcoded:

\`\`\`
https://login.microsoftonline.com/{tenant_id}/oauth2/v2.0/token
\`\`\`

An operator deploying behind a corporate proxy, on Azure Government cloud (\`login.microsoftonline.us\`), Azure China (\`login.partner.microsoftonline.cn\`), or pointing at a mock for e2e cannot redirect AAD token minting.

## Customer impact

1. **Azure Government / Azure China customers blocked** — they use entirely different AAD endpoints (\`.us\` / \`.cn\` suffixes). The bridge hardcodes \`.com\`.
2. **AISIX-Cloud AAD live e2e blocked** — \`adapter-azure-live.spec.ts\` currently only exercises the api-key path because mock-llm has no way to receive AAD client_credentials requests.

## Compare to peers

- **Vertex bridge** — production override works via the SA JSON's \`token_uri\` field (industry-standard GCP SA shape). \`crates/aisix-provider-vertex/src/token_mint.rs:269\` uses \`sa.token_uri.clone()\` in production.
- **Bedrock bridge** — \`endpoint_url\` from \`ProviderKey.api_base\` flows to the AWS SDK in production (fixed via #392).
- **Azure bridge** — this gap.

## Suggested fix

Add an optional \`aad_token_endpoint\` field to the AAD secret JSON:

\`\`\`json
{
  "tenant_id": "abc-123",
  "client_id": "...",
  "client_secret": "...",
  "aad_token_endpoint": "https://login.microsoftonline.us/abc-123/oauth2/v2.0/token"
}
\`\`\`

Default: derive from \`tenant_id\` against \`login.microsoftonline.com\` (current behavior).

Implementation:
1. Add \`pub aad_token_endpoint: Option<String>\` to \`AadCredentials\` struct
2. In \`resolve_token_endpoint\`, prefer \`creds.aad_token_endpoint\` over the hardcoded URL
3. Add scheme/userinfo/query/fragment defensive checks (mirror \`api_base\` validation pattern from #392)
4. Cap impact: when set, the URL is used verbatim; \`tenant_id\` is no longer interpolated (operator pasted the full path)

~30-50 LOC change + tests + docs note. Surface area: one new optional JSON field, fully backward compatible.

## Blocks

- AISIX-Cloud P1 #4 (AAD + SA-OAuth live spec) — Vertex half can ship via SA's existing \`token_uri\` mechanism; Azure half waits on this.
- Future Azure Government / Azure China customer support.

## Surfaced via

P1 #4 of #302 follow-up audit (AISIX-Cloud), discovered during scoping of mock-llm \`/{tenant}/oauth2/v2.0/token\` endpoint addition.

moonming:
## Summary

The AAD token-endpoint resolution had only a `#[cfg(test)]` override seam — production always POSTed to the hardcoded `https://login.microsoftonline.com/{tenant}/oauth2/v2.0/token`. This blocked any live e2e of the AAD `client_credentials` Bearer path against a mock token endpoint (AISIX-Cloud could only cover the api-key auth path), and silently excluded every Azure national/sovereign cloud tenant.

Adds an optional `authority_host` field to `AadCredentials` (parsed from the ProviderKey secret JSON). `resolve_token_endpoint` precedence:

1. `#[cfg(test)]` `token_endpoint_override` — unit-test seam (unchanged)
2. `creds.authority_host` — production override; tenant + standard `/oauth2/v2.0/token` path interpolated onto it
3. default public-cloud `login.microsoftonline.com`

## Why this is a real production feature (not just a test seam)

Azure national clouds use distinct AAD authorities per [Microsoft's national-cloud endpoints table](https://learn.microsoft.com/en-us/entra/identity-platform/authentication-national-cloud):
- US Gov: `https://login.microsoftonline.us`
- China (21Vianet): `https://login.chinacloudapi.cn`

It's the same value the Azure SDK reads from `AZURE_AUTHORITY_HOST`.

## Reference-implementation alignment

Mirrors the established cross-bridge pattern in this repo:
- **Vertex** carries its token endpoint in the SA JSON's own `token_uri` field (`token_mint.rs` resolves `sa.token_uri` in production; the `#[cfg(test)]` override is unit-test-only).
- **Vertex/Bedrock** honor `ProviderKey.api_base` for the data plane (#392).

Validation mirrors the Vertex `resolve_api_base` rules (#392): bare http(s) origin only; reject userinfo (`@`) / query (`?`) / fragment (`#`) with a **fixed** message (no echo) so a pasted `user:pass@host` credential never surfaces in logs; trailing slash trimmed to avoid `//{tenant}`.

## Backward compatibility

`authority_host` is `#[serde(default)]` `Option<String>`, so every existing operator secret (`{tenant_id, client_id, client_secret}`) resolves to public-cloud unchanged. No cp-api change needed — the required-config validator checks for missing required keys, not forbidden extra keys.

## Test plan

- [x] `cargo test -p aisix-provider-azure-openai` — 72 pass (66 existing + 6 new)
- [x] `cargo clippy -p aisix-provider-azure-openai --all-targets -- -D warnings` clean
- [x] `cargo fmt` clean
- [x] `cargo check --workspace` clean
- 6 new unit tests: default public-cloud resolution; production `authority_host` interpolation (NO `cfg(test)` override — exercises the real path); trailing-slash trim; end-to-end mint through a wiremock `authority_host`; validation rejects scheme-less host; validation rejects userinfo-bearing host without leaking it.

Unblocks the AISIX-Cloud Azure AAD live e2e (the last auth-path gap across the 3 platform adapters, #302 Phase F / D6.6).

Closes #413.

### DIFF
--- a/crates/aisix-provider-azure-openai/src/aad_token_mint.rs
+++ b/crates/aisix-provider-azure-openai/src/aad_token_mint.rs
@@ -73,6 +73,22 @@ pub(crate) struct AadCredentials {
     pub client_id: String,
     /// Client secret value (NOT the secret id). Confidential.
     pub client_secret: String,
+    /// Optional AAD authority host override. Absent for the common
+    /// public-cloud case, which defaults to
+    /// `https://login.microsoftonline.com`. Set when the tenant lives
+    /// in an Azure national / sovereign cloud whose AAD authority
+    /// differs from public Azure:
+    ///   - US Government: `https://login.microsoftonline.us`
+    ///   - China (21Vianet): `https://login.chinacloudapi.cn`
+    ///
+    /// per Microsoft's national-cloud authentication endpoints table
+    /// <https://learn.microsoft.com/en-us/entra/identity-platform/authentication-national-cloud>
+    /// (the same value the Azure SDK reads from the
+    /// `AZURE_AUTHORITY_HOST` environment variable). Must be a bare
+    /// http(s) origin — the bridge interpolates the tenant and the
+    /// `/oauth2/v2.0/token` path itself.
+    #[serde(default)]
+    pub authority_host: Option<String>,
 }
 
 impl AadCredentials {
@@ -106,6 +122,32 @@ impl AadCredentials {
                 return Err(BridgeError::Config(format!(
                     "azure aad credentials.{name} {value:?} contains URL-control \
                      characters — reject `/`, `?`, `#`, whitespace, `..`"
+                )));
+            }
+        }
+        // Optional authority_host: when present it becomes the origin
+        // of the token-endpoint URL, so it must be a bare http(s)
+        // origin. Mirrors the Vertex `resolve_api_base` validation
+        // (PR #392) — reject userinfo / query / fragment first (fixed
+        // message, no echo, so pasted `user:pass@host` credentials
+        // never surface in logs), then enforce the scheme.
+        if let Some(host) = self
+            .authority_host
+            .as_deref()
+            .map(str::trim)
+            .filter(|s| !s.is_empty())
+        {
+            if host.contains('@') || host.contains('?') || host.contains('#') {
+                return Err(BridgeError::Config(
+                    "azure aad credentials.authority_host must be a bare origin — \
+                     reject userinfo (@), query (?), fragment (#)"
+                        .into(),
+                ));
+            }
+            if !(host.starts_with("https://") || host.starts_with("http://")) {
+                return Err(BridgeError::Config(format!(
+                    "azure aad credentials.authority_host must use http:// or https:// \
+                     scheme, got {host:?}"
                 )));
             }
         }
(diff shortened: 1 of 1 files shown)


moonming:
**Independent audit: clean — no HIGH, no MEDIUM.**

Verified by a fresh cold-read agent against the sibling Vertex pattern:

- **No `client_secret` redirect vector** — `validate()` rejects `@`/`?`/`#` then enforces `http(s)://` scheme, and runs *before* `resolve_token_endpoint`, so a malformed authority never reaches the POST that carries the secret.
- **Userinfo no-leak ordering holds** — the `@`/`?`/`#` rejection (fixed, no-echo message) is ordered *before* the scheme-error branch that echoes `{host:?}`, so a scheme-less `user:pass@host` never surfaces the userinfo. Pinned by `validate_rejects_authority_host_with_userinfo_without_echoing_it`.
- **Real production path is tested** — `resolve_token_endpoint_honors_authority_host_in_production` + the wiremock `mint_via_production_authority_host_posts_to_interpolated_path` both run with NO `#[cfg(test)]` override, so a regression that ignored `authority_host` fails.
- **Backward-compat confirmed** — `#[serde(default)] Option` ⇒ existing `{tenant_id, client_id, client_secret}` secrets resolve to public cloud unchanged.

**LOW-1 (not blocking):** `authority_host` with an embedded path (`https://host/evil`) is accepted verbatim (→ `https://host/evil/{tenant}/oauth2/v2.0/token`). It's a path prefix, not a host redirect, and the field is operator-supplied config (not request input). Vertex's `resolve_api_base` has the identical gap, so tightening only Azure here would create asymmetry — filed as a shared cross-provider follow-up: api7/ai-gateway#435. Agreed not to block this merge.

Merging once CI is green.

Issue #245, opened 2026-05-11
docs: create landing page, overview, feature matrix, and roadmap


moonming:
Parent issue: #240

## Summary
Create the top-level docs entrypoint and product-orientation pages for the new documentation set.

## Scope
- `docs/index.md`
- `docs/roadmap.md`
- `docs/overview/what-is-aisix-ai-gateway.md`
- `docs/overview/deployment-modes.md`
- `docs/overview/core-concepts.md`
- `docs/overview/feature-matrix.md`

## Acceptance Criteria
- The docs homepage routes readers by audience and deployment mode
- AISIX AI Gateway is positioned as the primary product
- AISIX Cloud is positioned as the managed extension
- The feature matrix distinguishes `Available`, `Limited`, `Preview`, and links to the roadmap
- Not-yet-implemented capabilities are captured only in `docs/roadmap.md`

moonming:
Closed by the merge of #249.

PR #436, opened 2026-05-29
fix(messages): parse Anthropic streaming usage tokens (#245 dp-blocker)

moonming:
## Summary

Fixes **api7/AISIX-Cloud#245** (dp-blocker). Parity with the OpenAI streaming fix (#225 / #196).

Pre-fix, the Anthropic `/v1/messages` passthrough **streaming** branch forwarded the upstream byte stream verbatim and emitted a UsageEvent with `prompt_tokens=0 completion_tokens=0`. Every streaming `/v1/messages` request billed as **zero** — bypassing budget caps and under-counting revenue across the entire Anthropic-SDK streaming surface.

## Fix

`build_anthropic_passthrough_stream` wraps the byte stream:

- buffers chunks, drains complete SSE frames (blank-line delimited), **reassembling frames split across byte-chunk boundaries** (reqwest's `bytes_stream()` gives no frame-alignment guarantee)
- parses each `data:` JSON to accumulate usage:
  - `message_start` → `input_tokens`, `cache_creation/read_input_tokens`, message `id`, `model`
  - `content_block_*` → time-to-first-token
  - `message_delta` → running `output_tokens` (max-wins), `stop_reason`
- yields original bytes **unchanged** — client sees the exact Anthropic SSE wire shape
- fires `emit_anthropic_usage_event` from a Drop guard (`AnthropicStreamGuard`) so the event ships on normal end-of-stream **and** client-disconnect mid-stream (mirrors chat.rs `CompleteOnDrop`)

**#419 cost-leak parity:** `AnthropicDeliveryCounter` counts byte chunks that reached the client; if none did, the Drop guard zeroes completion-side counters while keeping `prompt_tokens`.

The streaming branch sets `usage_handled_by_stream: true` so the handler doesn't double-emit.

## Test plan

- [x] `anthropic_passthrough_streaming_records_usage_from_sse_frames` — end-to-end: realistic Anthropic SSE (input_tokens=37, output_tokens=52, cache fields, stop_reason) → asserts UsageEvent carries real counts + bytes verbatim + TTFT > 0
- [x] `sse_frame_parser_reassembles_split_chunks` — frame split across two byte chunks parsed only after terminator arrives
- [x] `stream_guard_zeroes_completion_when_nothing_delivered` — #419 gate: completion zeroed when delivered==0 (prompt kept), preserved when delivered>0
- [x] All 30 pre-existing `messages::tests` pass — no regression
- [x] `cargo clippy -p aisix-proxy -- -D warnings` clean

## References

- Issue: api7/AISIX-Cloud#245
- Reference fix: #225 / #196 (OpenAI streaming), #419 (cost-leak gate)
- Anthropic streaming spec: <https://docs.anthropic.com/en/api/messages-streaming>
- CP-side e2e gate `e2e/cases/messages_streaming_billing_test.go` becomes worth writing now the DP gap is closed (noted in the issue).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed token usage and metrics tracking for streaming API responses. Now accurately extracts prompt tokens, completion tokens, cache information, and response timing from upstream sources.

* **Tests**
  * Added comprehensive unit tests validating streaming token usage extraction, SSE event frame parsing across byte chunks, and cost calculations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/api7/ai-gateway/pull/436?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/aisix-proxy/src/messages.rs
+++ b/crates/aisix-proxy/src/messages.rs
@@ -1031,6 +1031,12 @@ fn emit_anthropic_usage_event(
 // end-of-stream OR on client-disconnect (mirroring chat.rs's
 // `CompleteOnDrop`), so a streamed request always ships a UsageEvent.
 
+/// Upper bound on the in-flight SSE frame buffer (PR #436 audit
+/// MEDIUM-2). Real Anthropic SSE frames are a few KB at most; this
+/// ceiling only trips on a non-conformant upstream that never emits a
+/// frame terminator, guarding against per-request memory exhaustion.
+const MAX_SSE_FRAME_BUF_BYTES: usize = 1 << 20; // 1 MiB
+
 /// Accumulated usage observed across an Anthropic SSE stream.
 /// Sourced from `message_start` (input + cache tokens, id, model) and
 /// `message_delta` (running output_tokens, stop_reason). All fields
@@ -1099,12 +1105,21 @@ fn update_anthropic_usage(
             }
         }
         Some("message_delta") => {
-            if let Some(t) = json
-                .get("usage")
-                .and_then(|u| u.get("output_tokens"))
-                .and_then(Value::as_u64)
-            {
-                acc.completion_tokens = acc.completion_tokens.max(t as u32);
+            if let Some(v) = json.get("usage").and_then(|u| u.get("output_tokens")) {
+                if let Some(t) = v.as_u64() {
+                    acc.completion_tokens = acc.completion_tokens.max(t as u32);
+                } else {
+                    // PR #436 audit LOW-1: a `usage` object present but
+                    // with a non-numeric `output_tokens` leaves
+                    // completion_tokens at the message_start floor
+                    // (often 1) — a silent under-count. Surface it so a
+                    // wire-shape drift is visible to operators.
+                    tracing::debug!(
+                        output_tokens = %v,
+                        "anthropic stream: message_delta usage.output_tokens \
+                         is non-numeric; completion_tokens left at floor"
+                    );
+                }
             }
             if let Some(sr) = json
                 .get("delta")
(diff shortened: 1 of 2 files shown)


moonming:
## Audit response (commit 7f8ef07)

**MEDIUM-2 (unbounded frame buffer) — ✅ fixed.** Added a 1 MiB `MAX_SSE_FRAME_BUF_BYTES` ceiling. Real Anthropic SSE frames are a few KB; the cap only trips on a non-conformant terminator-less stream, where we drop the buffer (skip usage parse) rather than OOM. Bytes still forward verbatim. Genuine new risk vs chat.rs (which operates on pre-framed chunks) — good catch.

**LOW-1 (non-numeric output_tokens) — ✅ fixed.** Added `tracing::debug!` when a `message_delta` has a `usage` object but `output_tokens` fails to parse, so a wire-shape drift is visible instead of silently leaving completion at the floor.

**Test angle 8c — ✅ added.** `sse_frame_parser_tolerates_streams_without_usage` pins that a usage-free error stream drains cleanly to zeros without panicking.

**E2E — ✅ added.** `anthropic-streaming-usage-e2e.test.ts` drives a real streaming `/v1/messages` through the DP binary against a mock Anthropic streaming upstream and asserts `aisix_llm_input_tokens_total` / `aisix_llm_output_tokens_total` for `/v1/messages` are non-zero. Pre-#245 these stayed at 0. Verified locally against the aisix-e2e stack.

---

**HIGH-1 — NOT actioned; the premise is factually incorrect.**

The finding states *"`grep -rn \"DeliveryCounter\|#419\" crates/` returns nothing"* and that `chat.rs::CompleteOnDrop` *"fires `on_complete(c)` unconditionally — no atomic, no delivery count."* Both claims are false on this branch:

```
$ grep -rn "DeliveryCounter" crates/
crates/aisix-proxy/src/chat.rs:2296:    DeliveryCounter {            # construction
crates/aisix-proxy/src/chat.rs:2312:struct DeliveryCounter<T> {      # the struct
crates/aisix-proxy/src/chat.rs:2317:impl<T> Stream for DeliveryCounter<T> {
...
$ grep -n "delivered == 0\|completion_tokens = 0" crates/aisix-proxy/src/chat.rs
2039:            if delivered == 0 {
2040:                c.completion_tokens = 0;     # the #419 cost-leak gate
```

`CompleteOnDrop` carries `delivered: Arc<AtomicU32>` (chat.rs:2010) and its `Drop` zeroes the completion-side counters when `delivered == 0` (chat.rs:2039-2045) — the exact behaviour this PR mirrors. The #419 cost-leak gate is referenced throughout chat.rs (lines 1961, 2001, 2030, 2069, 2306, 2656, 2723…).

So the PR's "mirrors chat.rs's `DeliveryCounter` / #419 parity" framing is **accurate**, and removing the `AnthropicDeliveryCounter` + `delivered == 0` gate would *break* parity with the OpenAI path, not restore it. The delivery gate stays.

(I suspect the audit's grep ran outside the worktree, or matched the stale comment at chat.rs:2809 that references a *prior* PR's HIGH-1.)

**MEDIUM-1 (byte-chunk vs SSE-event delivery granularity)** — kept as-is for parity: chat.rs's `DeliveryCounter` counts delivered SSE *Events* (including role-only / non-output events), so a client that receives the first event then disconnects already has `delivered >= 1` and gets billed the accumulated usage there too. Byte-chunk granularity is the natural analog for a raw-byte passthrough and matches that spirit. Tightening to "gate on a parsed content/message delta" would make Anthropic *stricter* than OpenAI — a divergence, not a fix. Noting rather than changing.

PR #443, opened 2026-05-29
feat(vertex): wire Claude-on-Vertex non-stream chat via :rawPredict

moonming:
## Summary

Vertex AI's Model Garden hosts Anthropic Claude alongside Gemini. The publisher resolver already mapped `claude-*` → Anthropic, but dispatch returned "not yet implemented". This wires the **non-stream** chat path (rail A of the Vertex-publisher build-out).

Claude on Vertex speaks the **Anthropic Messages** wire, so the bridge **reuses the shared `aisix-provider-anthropic::wire` serializer + response decoder** — the same components the Bedrock `/invoke` path reuses. Vertex-specific shaping:

| Aspect | Value |
|---|---|
| URL | `…/publishers/anthropic/models/<model>:rawPredict` |
| body | Anthropic Messages JSON, **minus `model`** (URL carries it), **minus `stream`** (rawPredict is non-stream), **plus `anthropic_version: "vertex-2023-10-16"`** (required; ≠ Bedrock's `bedrock-2023-05-31`) |
| auth | existing GCP OAuth2 Bearer (SA→JWT→OAuth mint+cache, or pre-minted) — shared with Gemini |
| response | native Anthropic Messages → shared `response_into_chat_response`; alias restore stays at the proxy render layer |

## Reference-impl + upstream sources (§7)

- Google Vertex AI Claude: https://cloud.google.com/vertex-ai/generative-ai/docs/partner-models/use-claude
- Anthropic on Vertex AI SDK: https://docs.anthropic.com/en/api/claude-on-vertex-ai

These confirm the `:rawPredict` / `:streamRawPredict` split, the `vertex-2023-10-16` version string, and that `model` lives in the URL (not the body).

## Out of scope (deferred to a focused follow-up, D5.3)

**Claude streaming (`:streamRawPredict`)** — the `stream`-field-in-body vs URL-only contract needs a primary-source confirmation before shipping rather than a guess (the building blocks — `AnthropicStreamEvent`/`StreamState` — are already `pub` and reusable). `chat_stream` returns a clear, actionable error noting non-stream is wired. The remaining publishers (`meta`/`mistral`/`ai21`, D5.4) stay deferred.

## Test plan

- [x] `cargo test -p aisix-provider-vertex` — 76 pass (incl. new dispatch + streaming-deferred tests)
- [x] `cargo clippy --all-targets -- -D warnings` clean; `cargo fmt` clean; `cargo check --workspace` clean
- New wiremock dispatch test pins: the exact `:rawPredict` URL, Bearer auth, body shape (`anthropic_version=vertex-2023-10-16`, `model`/`stream` stripped, `messages` + `max_tokens` present), and Anthropic-envelope response decode.
- Updated 3 existing tests that used `claude-*` as the not-implemented example → Mistral (still deferred), since Claude is now wired.

Follow-ups after this lands: AISIX-Cloud mock-vertex `:rawPredict` endpoint + live spec; then Claude-streaming (D5.3); then Llama-via-openapi-shim (D5.4).

Part of api7/AISIX-Cloud#302 Phase E (AC.11 Vertex publisher coverage).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -285,6 +285,7 @@ version = "0.1.0"
 dependencies = [
  "aisix-core",
  "aisix-gateway",
+ "aisix-provider-anthropic",
  "async-stream",
  "async-trait",
  "bytes",
--- a/crates/aisix-provider-vertex/Cargo.toml
+++ b/crates/aisix-provider-vertex/Cargo.toml
@@ -11,6 +11,11 @@ description = "aisix: Google Vertex AI provider bridge (skeleton — multi-publi
 [dependencies]
 aisix-core = { path = "../aisix-core" }
 aisix-gateway = { path = "../aisix-gateway" }
+# Claude on Vertex reuses the Anthropic Messages request serializer +
+# response decoder (same crate the Bedrock /invoke path reuses) — the
+# Vertex `:rawPredict` body is Anthropic Messages JSON with a
+# Vertex-specific `anthropic_version`.
+aisix-provider-anthropic = { path = "../aisix-provider-anthropic" }
 async-trait.workspace = true
 async-stream = "0.3"
 bytes.workspace = true
--- a/crates/aisix-provider-vertex/src/bridge.rs
+++ b/crates/aisix-provider-vertex/src/bridge.rs
@@ -1644,6 +1644,12 @@ mod tests {
         validate_url_token("region", "europe-west4").unwrap();
         validate_url_token("upstream_id", "gemini-1.5-pro").unwrap();
         validate_url_token("upstream_id", "gemini-2.0-flash-exp").unwrap();
+        // Vertex Claude model ids carry an `@<version>` suffix
+        // (e.g. `claude-3-5-sonnet@20241022`). The `@` is a legitimate
+        // path char and MUST pass validation — a future tightening of
+        // the reject set that banned `@` would break real Claude-on-
+        // Vertex dispatch, so pin it here.
+        validate_url_token("upstream_id", "claude-3-5-sonnet@20241022").unwrap();
     }
 
     #[test]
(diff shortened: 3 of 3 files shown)


moonming:
**Independent audit: clean — no HIGH, no MEDIUM.**

A fresh cold-read agent verified the change against `crates/` and the Bedrock `chat_anthropic` sibling. Confirmed would-fail-on-regression:
- URL exact `…publishers/anthropic/models/<model>:rawPredict`; `anthropic_version="vertex-2023-10-16"` (≠ Bedrock's `bedrock-2023-05-31`);
- **strip `model`/`stream` is load-bearing** — `AnthropicRequest.model` (`&str`) always serializes and `.stream` (bool) has no `skip_serializing_if` (serializes as `"stream": false`), so both `obj.remove(...)` do real work and the test's `!contains_key` assertions genuinely catch a regression;
- shared `split_system → build_request(..., false) → strip → insert-version` flow identical to Bedrock, only the version const differs;
- dispatch routing + deferred-stream error + `validate_url_token` (project/region/upstream_id) + safe auth/error mapping (no SA/token leak) all verified.

Two LOW:
- **L1 (addressed, f430082)** — no test pinned that a real Vertex Claude id with `@<version>` (`claude-3-5-sonnet@20241022`) passes `validate_url_token`. Added the guard so a future tightening that banned `@` can't silently break real Claude dispatch.
- **L2 (justified, no change)** — the streaming-deferred error embeds the internal `#302 D5.3` tracking id in a customer-visible `BridgeError::Config`. This is a **pre-existing** pattern (the Gemini path does the same); changing only this one arm would be inconsistent. Better addressed as a separate cross-bridge cleanup of roadmap-id-in-error across all publishers.

Merging on green CI.

PR #462, opened 2026-06-01
feat(bedrock): wire RequestOverrides on the Bedrock bridge (#302 §5, #340)

### LINKED ISSUE #340: wire RequestOverrides on aisix-provider-bedrock bridge (mirror OpenAiBridge apply pipeline)
## Gap (same root cause as #335/#337/<vertex follow-up>)

Verified by independent audit of PR #338: `crates/aisix-provider-bedrock/src/bridge.rs` does NOT invoke any of `apply_param_renames` / `apply_param_constraints` / `apply_default_body_fields` / `apply_default_headers` from `aisix-provider-openai::overrides`.

Customer flow: dashboard creates a chat request, AISIX routes to a Bedrock upstream via Hub dispatch → BedrockBridge. The PK's `request.*` override block silently no-ops on outbound.

## Fix scope

Mirror the OpenAI bridge wiring. Reference site: `crates/aisix-provider-openai/src/bridge.rs:310-367`.

Special consideration for Bedrock: SigV4 signing happens AFTER the headers are finalized, so `default_headers` must be applied BEFORE the SigV4 signing pass. `RESERVED_DEFAULT_HEADERS` already lists `x-amz-security-token` and `x-amz-date` so an operator-supplied `default_headers` entry naming those will not interfere with SigV4.

## Test plan

Mirror the 5 wiremock-driven tests pattern (already shipped by #338 for Anthropic-passthrough):
- `bedrock_applies_param_renames`
- `bedrock_clamps_temperature_via_param_constraints`
- `bedrock_fills_default_body_fields_when_caller_omits`
- `bedrock_injects_default_headers`
- `bedrock_default_headers_cannot_overwrite_x_amz_date` (SigV4 timestamp guard)

## Related
- ai-gateway#335 (same gap on /v1/messages Anthropic-passthrough — fixed in #338)
- ai-gateway#337 (same)
- AISIX-Cloud#302 §5 RuntimeConfig spec

moonming:
## What

Wires the per-`ProviderKey` **RequestOverrides** pipeline (issue #302 §5 RuntimeConfig) into the Bedrock bridge. Until now `crates/aisix-provider-bedrock/src/bridge.rs` invoked none of the override primitives — a Bedrock PK's `request.*` / `response.*` block silently no-op'd on outbound. Closes #340.

## Contract pinned

**Body overrides** — on the `/invoke` (Anthropic) JSON path, `apply_body_overrides` mirrors `OpenAiBridge::prepare_outbound_body` step-for-step: `param_renames` → `param_constraints` → `default_body_fields` → `content_list_to_string` (response-gated). Applied **before** the Bedrock `model`/`stream` strip, so an override can't reintroduce a URL-borne `model` into the `/invoke` body.

**default_headers** — ride a pre-signing interceptor (`DefaultHeadersInterceptor`, `Intercept::modify_before_signing`) registered on the per-request SDK client. This is the key Bedrock-specific consideration from #340: SigV4 signing runs *after* the request is built, so injecting at `modify_before_signing` puts the headers **inside** the signed canonical request rather than appending them post-signature. Applies uniformly across `/invoke` and **both Converse paths** (the interceptor is on the client config, not the operation). SigV4-owned headers (`wire::reserved_sigv4_headers`: `authorization`, `x-amz-date`, `x-amz-content-sha256`, `x-amz-security-token`, `x-amz-target`, `x-amzn-bedrock-accept`) are filtered out at construction so an operator override can never perturb the signature.

### Scope note — Converse path body overrides
The Converse path (`chat_converse` / `chat_converse_stream`, used by non-Anthropic publishers) builds its request through the SDK's **typed** `InferenceConfiguration` builder and carries no JSON body, so the top-level-key body transforms (`param_renames` / `default_body_fields` / `content_list_to_string`) have nothing to act on there — the operator sets those params through the normal request fields. `default_headers` **do** apply on Converse (header-level, publisher-agnostic). This is documented on `apply_body_overrides`.

## Reference
- Reference apply order + reserved-header guard: `OpenAiBridge::prepare_outbound_body` + `build_request_headers` (`crates/aisix-provider-openai/src/bridge.rs`); the shared primitives in `crates/aisix-provider-openai/src/overrides.rs` (reused verbatim).
- Pre-signing header injection: the AWS SDK `Intercept::modify_before_signing` hook — the same hook the SDK's own `user_agent` / `recursion_detection` interceptors use to add signed headers (`aws-runtime`). Mainstream gateways that front SigV4 providers inject operator/custom headers into the signed request the same way and exclude the signer-owned headers; this mirrors that approach.
- Bedrock `anthropic_version` body pin: <https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-messages.html>.

## Tests
8 wiremock round-trips **through the real AWS SDK** (so SigV4 actually runs and the captured request is the signed wire):
- `bedrock_applies_param_renames` — `temperature` → renamed key, source gone
- `bedrock_clamps_temperature_via_param_constraints` — over-max clamped
- `bedrock_fills_default_body_fields_when_caller_omits` — absent key injected
- `bedrock_injects_default_headers` — custom header reaches the wire AND a valid `AWS4-HMAC-SHA256` Authorization still present (signing not broken)
- `bedrock_default_headers_cannot_overwrite_x_amz_date` — reserved timestamp survives; non-reserved companion still applies
- `bedrock_default_headers_apply_on_converse_path` — header injection works on the Converse path too
- `bedrock_overrides_run_before_model_strip` — `default_body_fields.model` stays stripped; genuine field lands; `anthropic_version` shaping still runs
- `bedrock_response_content_list_to_string_flattens_outbound_body` — response-gated array→string flatten

`cargo test -p aisix-provider-bedrock` → 53 passed. `cargo clippy --all-targets` clean.

## Summary by CodeRabbit

* **New Features**
  * Bedrock provider now supports request transformation capabilities, including parameter customization, constraints, and header management
  * Request overrides are consistently applied across all provider implementations
  * Added comprehensive test coverage for request override functionality

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -242,6 +242,7 @@ dependencies = [
  "aisix-core",
  "aisix-gateway",
  "aisix-provider-anthropic",
+ "aisix-provider-openai",
  "async-stream",
  "async-trait",
  "aws-config",
--- a/crates/aisix-provider-bedrock/Cargo.toml
+++ b/crates/aisix-provider-bedrock/Cargo.toml
@@ -19,6 +19,10 @@ aisix-gateway = { path = "../aisix-gateway" }
 # AI21) and all streaming go through the unified Converse API
 # instead — see `chat_converse` / `chat_converse_stream`.
 aisix-provider-anthropic = { path = "../aisix-provider-anthropic" }
+# Reuse the per-ProviderKey request-override primitives (param_renames /
+# param_constraints / default_body_fields / content_list_to_string) so every
+# adapter applies the same wire transforms cp-api configures (#302 §5 / #340).
+aisix-provider-openai = { path = "../aisix-provider-openai" }
 async-stream = "0.3"
 async-trait.workspace = true
 thiserror.workspace = true
--- a/crates/aisix-provider-bedrock/src/bridge.rs
+++ b/crates/aisix-provider-bedrock/src/bridge.rs
@@ -2730,12 +2730,23 @@ mod tests {
         );
         // The custom header must NOT have broken signing: a valid SigV4
         // Authorization header is still present on the captured request.
+        let auth = headers
+            .get("authorization")
+            .and_then(|v| v.to_str().ok())
+            .expect("a SigV4 Authorization header");
         assert!(
-            headers
-                .get("authorization")
-                .and_then(|v| v.to_str().ok())
-                .is_some_and(|a| a.starts_with("AWS4-HMAC-SHA256")),
-            "the request must still carry a valid SigV4 Authorization; headers={headers:?}",
+            auth.starts_with("AWS4-HMAC-SHA256"),
+            "the request must still carry a valid SigV4 Authorization; got {auth}",
+        );
+        // Prove the injected header was SIGNED — it appears in the SigV4
+        // `SignedHeaders=` list — rather than merely appended after signing.
+        // A post-signing injection would leave a valid-looking Authorization
+        // but omit `x-corp-trace` from SignedHeaders, so this assertion is
+        // what actually pins `modify_before_signing` ordering. (#462 audit LOW-1)
+        assert!(
+            auth.contains("SignedHeaders=") && auth.contains("x-corp-trace"),
+            "the injected default header must be covered by the SigV4 signature \
+             (present in SignedHeaders); auth={auth}",
         );
     }
 
--- a/crates/aisix-provider-bedrock/src/bridge.rs
+++ b/crates/aisix-provider-bedrock/src/bridge.rs
@@ -44,6 +44,20 @@ use aisix_provider_anthropic::wire::{
     build_request, response_into_chat_response, split_system, AnthropicResponse,
 };
 
+// Per-`ProviderKey` request override pipeline (#302 §5 / #340). The JSON-body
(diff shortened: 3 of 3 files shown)


moonming:
Independent pre-merge audit (CLAUDE.md §8) returned **CLEAR to merge** — correctness, reliability, security, and breaking-change angles all verified against AWS SDK source (confirmed `modify_before_signing` runs before `sign_request` in `aws-smithy-runtime`; SigV4 reserved-header filter is case-insensitive and complete; no dependency cycle; no-override path byte-identical). Two non-blocking items:

- **LOW-1** (signing assertion proved header *presence*, not that it was *signed*) → **fixed** in the follow-up commit: `bedrock_injects_default_headers` now asserts `x-corp-trace` appears in the SigV4 `SignedHeaders=` list, empirically pinning the pre-signing ordering.
- **MEDIUM-1** (product gap — `param_constraints` temperature clamp silently no-ops on the Converse path for non-Anthropic publishers) → **filed as #463**, agreed not to block this merge (it's an additive enhancement; nothing this PR ships is wrong, and the Converse scope limit is documented). The other three body transforms correctly don't apply to the typed Converse API.

53 tests pass; clippy clean.

PR #464, opened 2026-06-01
fix(vertex,azure): reject embedded path in authority/api_base overrides (#435)

### LINKED ISSUE #435: harden: reject embedded path in token-endpoint authority overrides (Azure authority_host + Vertex api_base)
## Context

Surfaced by the independent audit of #434 (LOW-1). Both platform-adapter bridges that accept an operator-supplied origin override for an auth/data endpoint validate scheme + reject userinfo/query/fragment, but **neither rejects an embedded path component**:

- **Azure** `AadCredentials.authority_host` (`crates/aisix-provider-azure-openai/src/aad_token_mint.rs`, added in #434) — `authority_host = "https://login.microsoftonline.us/evil"` yields `https://login.microsoftonline.us/evil/{tenant}/oauth2/v2.0/token`.
- **Vertex** `resolve_api_base` (`crates/aisix-provider-vertex/src/bridge.rs:143-166`) — same gap; an `api_base` with a trailing path segment is accepted verbatim.

## Severity

LOW. The field is **operator-supplied config from their own ProviderKey secret**, not attacker-controlled request input. The failure mode is an operator typo / mis-paste producing a 404 token-mint (or a wrong upstream path) rather than a clean `Config` error — not a host redirect (the host portion is whatever DNS name the operator chose). No `client_secret` exfiltration vector: a path prefix can't change the destination host.

## Suggested fix (both bridges, for symmetry)

After the existing scheme check, reject a path segment so only `scheme://host[:port]` is accepted:

```rust
let after_scheme = host.split_once("://").map(|(_, r)| r).unwrap_or(host);
if after_scheme.contains('/') {
    return Err(BridgeError::Config(
        "<provider> authority/api_base must be a bare origin with no path".into(),
    ));
}
```

Plus a unit test per bridge asserting `"https://host/evil"` is rejected.

## Why filed separately (not in #434)

Addressing it in #434 would tighten Azure but leave Vertex inconsistent. A single cross-provider PR keeps the two validation paths symmetric. Agreed in the #434 audit not to block that merge on this.

moonming:
## What

Closes #435. Both platform-adapter bridges accept an operator-supplied origin override for an auth/data endpoint and already reject scheme / userinfo / query / fragment, but **neither rejected an embedded path component**:

- **Vertex** `resolve_api_base` (`crates/aisix-provider-vertex/src/bridge.rs`) — an `api_base` with a trailing path segment was accepted verbatim, silently redirecting every upstream call onto the wrong path.
- **Azure** `AadCredentials::validate` (`crates/aisix-provider-azure-openai/src/aad_token_mint.rs`) — same gap; a path segment would redirect the token endpoint (e.g. `https://login.microsoftonline.us/evil/{tenant}/oauth2/v2.0/token`).

## Fix

After the existing scheme check on both, reject a path segment so only `scheme://host[:port]` is accepted:

```rust
let after_scheme = b.split_once("://").map(|(_, r)| r).unwrap_or(b).trim_end_matches('/');
if after_scheme.contains('/') { return Err(BridgeError::Config("... bare origin ... no path ...")); }
```

Two deliberate details:
- **Trailing slash stays tolerated** — the check `trim_end_matches('/')` before testing, because Vertex already trims a trailing slash on return and Azure's `resolve_token_endpoint` trims it at URL-build time (both have existing tests for that). So `https://host/` validates; `https://host/evil` does not.
- **Runs after the userinfo/query/fragment rejections** — those reject with a fixed, no-echo message; the new path check echoes `{value:?}`, but at that point the value provably has no `@`/`?`/`#`, so it can't leak pasted `user:pass@host` credentials.

## Severity
LOW (per #435). The field is operator-supplied config from their own ProviderKey secret, not attacker-controlled request input — the failure mode is an operator typo producing a clean `Config` error instead of a confusing 404 token-mint / wrong upstream path. The host portion is whatever DNS name the operator chose, so there's no host-redirect or `client_secret`-exfiltration vector; a path prefix can't change the destination host.

## Tests
Per bridge:
- rejection: `https://host/evil` → `Config` error mentioning "bare origin" / "no path"
- allow (false-positive guard): `https://host:8443` (the `:port` colon must not trip the `/` check); Azure also re-confirms a bare trailing slash validates
- existing trailing-slash-trim tests still pass

`cargo test -p aisix-provider-vertex -p aisix-provider-azure-openai` → 92 + 68 passed. `cargo clippy --all-targets` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced Azure AD credential authority host validation to ensure proper endpoint configuration and reject malformed paths.
  * Strengthened Vertex provider API base validation to enforce strict origin format requirements and reject embedded path components.

### DIFF
--- a/crates/aisix-provider-azure-openai/src/aad_token_mint.rs
+++ b/crates/aisix-provider-azure-openai/src/aad_token_mint.rs
@@ -154,14 +154,17 @@ impl AadCredentials {
             // is a valid origin. A real path segment would silently redirect
             // the token endpoint (e.g. `.../evil/{tenant}/oauth2/v2.0/token`).
             // A bare trailing slash is tolerated for symmetry with the Vertex
-            // `resolve_api_base` check. `host` has no `@`/`?`/`#` here, so
-            // echoing it is safe. Audit #434 LOW-1 / #435.
+            // `resolve_api_base` check. Backslashes are rejected too: the
+            // WHATWG URL parser the HTTP client uses normalizes `\` to `/` on
+            // http(s) URLs, so `host\evil` injects a path exactly like
+            // `host/evil`. `host` has no `@`/`?`/`#` here, so echoing it is
+            // safe. Audit #434 LOW-1 / #435 (+ #464 audit MEDIUM).
             let after_scheme = host
                 .split_once("://")
                 .map(|(_, rest)| rest)
                 .unwrap_or(host)
                 .trim_end_matches('/');
-            if after_scheme.contains('/') {
+            if after_scheme.contains('/') || after_scheme.contains('\\') {
                 return Err(BridgeError::Config(format!(
                     "azure aad credentials.authority_host must be a bare origin \
                      (scheme://host[:port]) with no path, got {host:?}"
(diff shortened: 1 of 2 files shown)


moonming:
Independent pre-merge audit (CLAUDE.md §8) returned **CLEAR with one MEDIUM**, now resolved:

- **MEDIUM** — backslash bypass: `https://host\evil` survived the `contains('/')` check, but the WHATWG `url` parser reqwest uses normalizes `\` → `/` on http(s), so it injected a `/evil` path — the exact silent-redirect outcome #435 blocks, contradicting the PR's "only scheme://host[:port]" wording. **Fixed** in the follow-up commit: both bridges now reject `\` alongside `/`, with a backslash rejection test each.

Cleared by the audit (verified against `url v2.5.8` / `reqwest v0.12.28`, not asserted): credential-leak ordering safe on both (the echoing path-check is unreachable for any `@`-bearing input — userinfo rejected+redacted first); no false rejection of any legit origin (bare host, host:port, IPv6 `[::1]`/`[2001:db8::1]`, http/https, trailing slash all accepted); `%2F` fails closed (URL parse error, no redirect); both bridges behaviorally symmetric; pre-existing trailing-slash tests still pass.

Vertex 93 + Azure 69 tests pass; clippy clean.

PR #468, opened 2026-06-01
fix(bedrock): apply param_constraints temperature clamp on the Converse path (#463)

### LINKED ISSUE #463: Bedrock: param_constraints (temperature clamp) silently no-ops on the Converse path (non-Anthropic publishers)
## Gap

Surfaced by the independent audit of #462 (MEDIUM-1).

#462 wired `RequestOverrides` on the Bedrock bridge. The body-override pipeline (`apply_body_overrides`: `param_renames` / `param_constraints` / `default_body_fields` / `content_list_to_string`) is applied on the **`/invoke` (Anthropic) JSON path** only. The **Converse path** (`chat_converse` / `chat_converse_stream`, used by every non-Anthropic publisher — Meta / Mistral / Cohere / Amazon Titan/Nova / AI21) builds its request through the SDK's typed `InferenceConfiguration` and carries no JSON body, so `apply_body_overrides` is never called there.

For the top-level-key transforms (`param_renames` / `default_body_fields` / `content_list_to_string`) that's correct — there's no JSON body to act on. **But `param_constraints.temperature_max` is a semantic clamp an operator can reasonably expect to hold across all publishers on a PK**, and the typed Converse builder has a `temperature` field it *could* clamp. Today:

- `param_constraints.temperature_max = 1.0` on a Bedrock PK **clamps** Anthropic-on-Bedrock (`/invoke`).
- The same PK used for a Meta-Llama request through Converse forwards `temperature` **unclamped** (`build_inference_config`, `crates/aisix-provider-bedrock/src/bridge.rs`, passes `req.temperature` verbatim).

A budget/guardrail-minded operator setting a temperature ceiling would be silently surprised it doesn't hold for non-Anthropic publishers on the same key.

## Suggested fix

Clamp `req.temperature` against `param_constraints` before `build_inference_config` on both Converse methods:

```rust
// in chat_converse / chat_converse_stream, before build_inference_config(req):
let mut req_clamped = req.clone();
if let Some(c) = ctx.provider_key.request.as_ref().and_then(|r| r.param_constraints.as_ref()) {
    if let Some(t) = req_clamped.temperature {
        req_clamped.temperature = Some(t.clamp(
            c.temperature_min.unwrap_or(f32::MIN),
            c.temperature_max.unwrap_or(f32::MAX),
        ));
    }
}
```

Plus a wiremock round-trip on a Converse (e.g. Meta) publisher asserting the captured `inferenceConfig.temperature` is clamped.

## Severity / scope

MEDIUM. Not a security issue; nothing #462 shipped is wrong (it documents the Converse scope limit). This is an additive enhancement to make `param_constraints` semantics uniform across publishers on a PK. Agreed in the #462 audit not to block that merge.

Tracks: AISIX-Cloud#302 §5 RuntimeConfig.

moonming:
## What

Closes #463. The Bedrock `/invoke` (Anthropic) path runs the body override pipeline (`apply_body_overrides` → `apply_param_constraints`), but the **Converse path** (non-Anthropic publishers: Meta / Mistral / Cohere / Titan / Nova / AI21) built its typed `inferenceConfig` from `req.temperature` **verbatim**. So a Bedrock-wide `param_constraints.temperature_{min,max}` ceiling — a determinism / cost / safety guardrail — held on Anthropic-on-Bedrock but **silently no-op'd** for every Converse publisher routed through the same ProviderKey. Silent guardrail failure is the worst kind.

## Fix

Thread `Option<&ParamConstraints>` into `build_inference_config` and clamp `temperature` there, so both `chat_converse` and `chat_converse_stream` share one clamp with **no `req` clone**.

Two correctness details (both were traps in the originally-suggested patch):
- **Panic-safety:** uses two independent `if` comparisons, **not `f32::clamp`** — `f32::clamp(min, max)` *panics* when an operator misconfigures `min > max`. This mirrors `apply_param_constraints` (`aisix-provider-openai::overrides`), which also avoids `clamp` for the same reason. Max-then-min order matches the `/invoke` path (min wins on a misconfigured `min > max`).
- **Type-correctness:** `ChatFormat.temperature` is `f32` but `ParamConstraints.temperature_{min,max}` are `f64`. The comparison is done in **f64-space** (exact) and the cast to `f32` happens only on assignment, so an `f64` bound not representable in `f32` still clamps at the right boundary.

### Scope (deliberately narrow)
- **Temperature only** — the sole field `ParamConstraints` defines (it's `deny_unknown_fields`; a test asserts `top_p_max` is *rejected*). `inferenceConfig.topP`/`maxTokens` could be clamped, but there's no constraint field to read; inventing one is out of scope.
- **`additionalModelRequestFields` intentionally NOT used** for the other override primitives. Verified against the [AWS Converse API reference](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html): that field carries each model's *native* parameter names (its own example passes `top_k`), whereas `default_body_fields` holds OpenAI-shaped, operator-authored keys. Broadcasting those to every publisher's `additionalModelRequestFields` would 400 or be silently ignored per publisher. Comparable multi-provider gateways forward provider-specific params to a single target provider rather than broadcasting them — so `param_renames` / `default_body_fields` / `content_list_to_string` correctly have no Converse target. The `apply_body_overrides` doc comment is updated to say so.

## Tests
3 Converse wiremock round-trips through the real AWS SDK (capture the outbound request body, assert `inferenceConfig.temperature`):
- `chat_converse_clamps_temperature_via_param_constraints` — `1.9` → clamped to `temperature_max` `1.0`
- `chat_converse_clamps_temperature_min_via_param_constraints` — `0.1` → clamped up to `temperature_min` `0.5`
- `chat_converse_temperature_clamp_does_not_panic_on_min_gt_max` — misconfigured `min=1.0 > max=0.0` must not panic (the `f32::clamp` form would); pins max-then-min order

`cargo test -p aisix-provider-bedrock` → 56 passed. `cargo clippy --all-targets` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Temperature constraints are now consistently enforced across all request types.
  * Improved robustness when handling constraint configurations.

* **Tests**
  * Added validation tests for temperature constraint enforcement across different request paths.

### DIFF
--- a/crates/aisix-provider-bedrock/src/bridge.rs
+++ b/crates/aisix-provider-bedrock/src/bridge.rs
@@ -49,7 +49,7 @@ use aisix_provider_anthropic::wire::{
 // `default_headers` ride a pre-signing interceptor (see
 // [`DefaultHeadersInterceptor`]) so they land inside the SigV4-signed
 // canonical request rather than being appended after the signature is computed.
-use aisix_core::RequestOverrides;
+use aisix_core::{ParamConstraints, RequestOverrides};
 use aisix_provider_openai::overrides::{
     apply_content_list_to_string, apply_default_body_fields, apply_param_constraints,
     apply_param_renames,
@@ -397,7 +397,9 @@ fn filtered_default_headers(
 /// Only the `invoke_model` (Anthropic) path carries a JSON body; the Converse
 /// path builds its request through the SDK's typed `InferenceConfiguration`
 /// builder, which has no JSON body for these top-level-key transforms to act
-/// on (the operator clamps temperature etc. via the normal request fields).
+/// on. (`param_constraints`'s temperature clamp IS still re-applied on the
+/// Converse path via `build_inference_config` — see #463 — but `param_renames`
+/// / `default_body_fields` / `content_list_to_string` have no Converse target.)
 fn apply_body_overrides(body: &mut serde_json::Value, ctx: &BridgeContext) {
     if let Some(r) = ctx.provider_key.request.as_ref() {
         apply_param_renames(body, &r.param_renames);
@@ -781,7 +783,16 @@ impl BedrockBridge {
         // forwards these via build_request → AnthropicRequest;
         // dropping them on the Converse path is a silent
         // behavioural regression for every non-Anthropic customer.
-        if let Some(cfg) = build_inference_config(req) {
+        // #463: re-apply the PK's param_constraints temperature clamp here —
+        // only the /invoke body path runs apply_param_constraints otherwise,
+        // so without this a Bedrock-wide temperature ceiling silently no-ops
+        // for every Converse (non-Anthropic) publisher.
+        let constraints = ctx
+            .provider_key
+            .request
+            .as_ref()
+            .and_then(|r| r.param_constraints.as_ref());
+        if let Some(cfg) = build_inference_config(req, constraints) {
             call = call.inference_config(cfg);
         }
 
@@ -842,8 +853,14 @@ impl BedrockBridge {
         for mb in message_blocks {
             call = call.messages(mb);
         }
-        // Mirror chat_converse on the stream path — same audit fix.
-        if let Some(cfg) = build_inference_config(req) {
+        // Mirror chat_converse on the stream path — same audit fix +
+        // the #463 param_constraints temperature clamp.
+        let constraints = ctx
+            .provider_key
+            .request
+            .as_ref()
+            .and_then(|r| r.param_constraints.as_ref());
+        if let Some(cfg) = build_inference_config(req, constraints) {
             call = call.inference_config(cfg);
         }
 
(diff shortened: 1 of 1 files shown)


moonming:
Independent pre-merge audit (CLAUDE.md §8) returned **CLEAR** — correctness, panic-safety, scope, and breaking-change posture all verified empirically (the auditor replicated the clamp and proved `f32::clamp` would panic on the no-panic test's exact inputs; confirmed no double-clamp, `/invoke` untouched, temperature-only scope correct, private fn). One MEDIUM + one LOW, both fixed in the follow-up commit:

- **MEDIUM-1** — `chat_converse_stream`'s clamp was wired but untested (all 3 tests drove non-stream `chat()`). **Fixed**: added `chat_converse_stream_clamps_temperature_via_param_constraints` (a `/converse-stream` round-trip asserting the clamp reaches the wire).
- **LOW** — the two identical `param_constraints` extraction blocks risked drift. **Fixed**: extracted `pk_param_constraints(ctx)` as the single source both Converse paths read.

57 tests pass; clippy clean.

PR #498, opened 2026-06-02
fix(dispatch): rate-limit compat-shim warn to once per PK (#377)

### LINKED ISSUE #377: [follow-up to #375] compat-shim tracing::warn! fires per-request — rate-limit or one-shot
Follow-up to [#375](https://github.com/api7/ai-gateway/pull/375) round-3 audit LOW-1.

## Gap

`crates/aisix-proxy/src/dispatch.rs::resolve_bridge` emits `tracing::warn!(target: \"aisix_proxy::dispatch\", ...)` whenever a pre-Phase-A PK row falls through the compat shim. The warn fires **per request**, not per PK — a heavily-loaded customer with an un-migrated PK gets thousands of identical lines.

Verified the warn sits in the cold branch (after two-tier miss), so happy-path overhead is zero. But for the legacy-PK case, log volume can mask other signals.

## Why this is LOW

- Round-3 audit explicitly called this non-blocking (\"operators expect log volume to surface legacy migration debt\").
- The signal is genuinely useful: an SRE seeing 1000 warns/hour knows there's at least one un-migrated PK; the volume IS the urgency hint.
- Rate-limiting is straightforward when needed.

## Fix sketch

Option A (one-shot, simplest):

```rust
use std::sync::atomic::{AtomicBool, Ordering};
static SHIM_WARNED: AtomicBool = AtomicBool::new(false);

// inside the `if` body in resolve_bridge:
if !SHIM_WARNED.swap(true, Ordering::Relaxed) {
    tracing::warn!(/* existing structured fields */);
}
```

One warn per process lifetime. Re-arms on restart so a fresh deploy resurfaces the signal if the migration debt still exists.

Option B (per-PK):

```rust
static SHIM_WARNED_PKS: DashSet<String> = ...; // pseudo
if SHIM_WARNED_PKS.insert(provider_key.display_name.clone()) {
    tracing::warn!(...);
}
```

One warn per unique PK, useful when operators want to enumerate every un-migrated row. Adds a small DashSet at module level.

Option C: `tracing-subscriber` rate-limiting (e.g. `RateLimit::default()`) at the subscriber layer. Cleanest if other parts of the codebase already use it; otherwise overkill for one site.

Recommend Option A unless an operator reports needing per-PK enumeration; cheap to implement, cheap to revert.

## When to close

When (a) cp-api has confirmed all production PKs have been backfilled with `provider` + `adapter`, or (b) an operator reports log noise from this specific warn. The shim itself is one-cycle; the warn outlives the shim only if we forget to delete both together.

jarvis9443:
Closes #377.

### Problem
`resolve_bridge`'s compat-shim warning (for pre-Phase-A PK rows with empty `provider` + `adapter: None`) fired **per request**. A heavily-loaded customer with a single un-migrated PK emitted thousands of identical warns, masking other log signals.

### Fix
De-dupe per PK display name via a process-local `DashSet` (option B from the issue): one warn per distinct un-migrated PK rather than per request. This keeps the migration-debt signal — an SRE still sees one line per un-migrated row, which is the useful part — while dropping the noise. The set is process-local, so a fresh deploy re-surfaces the warning if the debt still exists. Happy-path dispatch (two-tier hit) is untouched; this only affects the cold compat branch.

### Tests
The gate is extracted into `should_warn_shim(seen, pk)` and unit-tested: first sighting of a PK warns, repeats don't, a distinct PK warns again.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced duplicate warning messages emitted during proxy operations, improving logging clarity and decreasing log noise.

* **Tests**
  * Added test coverage to verify warning deduplication behavior.

### DIFF
--- a/crates/aisix-proxy/src/dispatch.rs
+++ b/crates/aisix-proxy/src/dispatch.rs
@@ -18,12 +18,31 @@
 //! plumbing flows naturally.
 
 use std::sync::Arc;
+use std::sync::LazyLock;
 
 use aisix_core::resource::ResourceEntry;
 use aisix_core::{AisixSnapshot, Model, ProviderKey};
 use aisix_gateway::{Bridge, Hub};
+use dashmap::DashSet;
 
 use crate::error::ProxyError;
+
+/// PK display names already warned about for compat-shim dispatch.
+/// Without this, a heavily-loaded customer with one un-migrated PK
+/// emits an identical warn on every request, drowning other signals
+/// (#377). De-duping per PK keeps the migration-debt signal (one line
+/// per un-migrated row) while dropping the per-request noise. The set
+/// is process-local, so a fresh deploy re-surfaces the warning if the
+/// debt still exists.
+static SHIM_WARNED_PKS: LazyLock<DashSet<String>> = LazyLock::new(DashSet::new);
+
+/// Gate for the compat-shim warn: returns `true` only the first time a
+/// given PK is seen. Extracted so the once-per-PK behavior is
+/// unit-testable without capturing tracing output. `DashSet::insert`
+/// returns `true` when the key was newly inserted.
+fn should_warn_shim(seen: &DashSet<String>, pk_display_name: &str) -> bool {
+    seen.insert(pk_display_name.to_string())
+}
 
 /// Resolve the Bridge to dispatch this request through.
 ///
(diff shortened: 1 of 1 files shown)


jarvis9443:
Superseded by a cleaner fix. Per discussion: the product is in active development with no released data, so there are no pre-Phase-A PK rows to migrate — rate-limiting the warn was treating a symptom of a case that can't occur. Replacing this with the full removal of the compat shim (no migration tech debt at first release): #501.

jarvis9443:
Correction: the replacement PR is #505 (not #501).

Issue #466, opened 2026-06-01
guardrails: streaming output moderation forwards content live before check_output (pre-P2 leak)


moonming:
## Summary

For **streaming** responses, output guardrails do not actually hold content back — the forbidden content reaches the client, and a `content_filter` error frame is appended *after* it.

`build_sse_stream` (`crates/aisix-proxy/src/chat.rs`) `yield`s each rendered content chunk **live inside the loop** (`:2188`), accumulating the text into `content_buffer` only for a post-loop scan. `check_output` runs **after** the loop (`:2225`) and, on `Block`, emits a `content_filter` frame — but every content chunk has already been delivered. The comment at `:2195-2202` describes a "buffer-then-check ... security control", but the live `yield` defeats it for streaming: the secret/forbidden text is on the wire before the check runs.

This affects every guardrail whose streamed-output policy is `StreamOutputPolicy::EndOfStreamCheck` — today that's `kind=keyword` and `kind=azure_content_safety` (Prompt Shield) on the output hook. It is **pre-existing** (predates P2), surfaced while implementing P2 streaming hold-back.

## P2 status

P2 (`azure_content_safety_text_moderation`) introduces real hold-back via `StreamOutputPolicy::Window` / `BufferFull` (ai-gateway#461 + its streaming increment), consumed by a `build_sse_stream` rewrite. P2 guardrails will hold streamed output until it scans clean.

`kind=keyword` and `kind=azure_content_safety` keep the current `EndOfStreamCheck` (live-forward) behavior — **this issue tracks deciding whether to give them a hold-back streaming policy too** (e.g. default keyword/prompt_shield output to `BufferFull`, or expose a per-guardrail setting). Trade-off: hold-back fixes the leak but adds streaming latency (TTFT) for those kinds.

## Pointers

- `crates/aisix-proxy/src/chat.rs:2188` — live `yield` of content chunks.
- `crates/aisix-proxy/src/chat.rs:2213-2248` — end-of-stream `check_output` + `content_filter` frame.
- `crates/aisix-guardrails/src/lib.rs` — `StreamOutputPolicy` + `Guardrail::stream_output_policy` (the new hold-back plumbing).

Severity: MEDIUM — a genuine streaming moderation bypass for keyword/prompt-shield output guardrails (non-streaming output is unaffected; it blocks correctly).

jarvis9443:
Proposed fix — split per affected guardrail kind, since the two have different root causes:

[...]

PR #501, opened 2026-06-03
fix(openai): normalise OpenRouter non-stream message.reasoning → reasoning_content (#648)

moonming:
## Summary

Fixes the **non-stream** half of the OpenRouter reasoning gap in AISIX-Cloud **#648**. OpenRouter (and some OpenAI-compatible aggregators) return a reasoning model's chain-of-thought at `message.reasoning` (string) — **not** the DeepSeek-canonical `message.reasoning_content`. `OpenAiResponseMessage` was a closed struct deserializing only `reasoning_content`, so `message.reasoning` was dropped at decode. A reasoning model that emits its whole answer as reasoning (e.g. `z-ai/glm-4.6` via OpenRouter) then reached the customer with **both** `content` AND `reasoning_content` empty.

**Evidence** (AISIX-Cloud real-chain e2e): the z-ai non-stream call returns `status=200, completion_tokens=230` — the upstream generated output — yet the customer saw empty `content`+`reasoning_content`. Consistent across runs; not model nondeterminism.

The **streaming** path already handles OpenRouter's `delta.reasoning` via the per-key `reasoning_field` override — but that lift is streaming/`delta`-only (`extract_reasoning_field` hard-rejects non-`delta` paths), so the non-stream response had no path to surface it.

## Change (auto-normalize — no per-key config required)
`crates/aisix-provider-openai/src/wire.rs`:
- Capture `message.reasoning` on `OpenAiResponseMessage` (was dropped at deserialize).
- In `response_into_chat_response`, lift it into the canonical `reasoning_content` `extra` slot **when the canonical `reasoning_content` is absent/empty**. DeepSeek-canonical `reasoning_content` takes precedence when both are present; an empty `reasoning` adds no noise field (same skip-empty rule as #466). Symmetric with the existing non-stream `reasoning_content` capture.

## Scope / blast radius (this is a global OpenAI-wire change, not OpenRouter-gated)
`OpenAiResponseMessage` is the **shared** OpenAI chat-completions response type for the whole workspace — it's also what `aisix-provider-azure-openai` and any future OpenAI-compatible bridge parse responses through (see the module doc on `pub` visibility). So this normalization is **not** keyed to OpenRouter: *any* upstream routed through the OpenAI wire that returns a non-empty `message.reasoning` (and no/empty `message.reasoning_content`) will now have it surfaced as canonical `reasoning_content`.

This is intended — `reasoning` is the de-facto field for OpenAI-compatible aggregators — and is safe because:
- it only **fills** the canonical slot when the canonical field is absent/empty (never overrides a real `reasoning_content`);
- it adds **no** field when `reasoning` is empty/absent (serde `default`), so OpenAI / Azure-OpenAI / DeepSeek-compat upstreams that never emit `reasoning` are byte-for-byte unaffected.

The blast radius is "OpenAI-compatible upstreams that emit `message.reasoning`", which today is OpenRouter-class aggregators; calling it out so reviewers don't read it as an OpenRouter-only branch.

## Test plan
- [x] 4 unit tests in `wire.rs`: (1) OpenRouter `reasoning`→canonical; (2) canonical-`reasoning_content`-wins precedence; (3) empty-`reasoning`-adds-no-field; (4) **empty-canonical-`reasoning_content`-falls-through-to-`reasoning`** — the discriminating guard exercising the real precedence logic (`reasoning_content.filter(!empty).or(reasoning)`). Verified tests (1) and (4) **FAIL** against the pre-fix canonical-only code and pass with the fix (genuine regression guards, not non-discriminating).
- [x] Existing 8 reasoning tests pass (no regression) — 12 reasoning tests green total.
- [x] `cargo clippy -p aisix-provider-openai` + `cargo fmt --check` clean.
- [x] Sibling crates that reuse these wire types (`aisix-provider-azure-openai`, `aisix-proxy`) build.
- [ ] CI green.

## Streaming parity (follow-up, out of scope here)
This PR makes the **non-stream** path auto-normalize with no operator config. Bringing the **streaming** path to the same no-config parity (default the `delta.reasoning` lift for the OpenRouter adapter so it works without a per-key `reasoning_field` override) is tracked in **#502**.

Closes the non-stream half of api7/AISIX-Cloud#648. Once this lands in the `:dev` image, the AISIX-Cloud `openrouter-longtail` real-chain z-ai non-stream leg goes green without the held-back workaround.

## Release Notes

* **New Features**
  * Added support for handling reasoning content in OpenAI API responses. Reasoning information is now properly captured and included in response data when available.

### DIFF
--- a/crates/aisix-provider-openai/src/wire.rs
+++ b/crates/aisix-provider-openai/src/wire.rs
@@ -176,6 +176,17 @@ pub struct OpenAiResponseMessage {
     /// `choices[0].message.reasoning_content`.
     #[serde(default)]
     pub reasoning_content: Option<String>,
+    /// OpenRouter (and some other OpenAI-compatible aggregators) put a
+    /// reasoning model's chain-of-thought at `message.reasoning` —
+    /// NOT the DeepSeek-canonical `message.reasoning_content` (#648).
+    /// Captured here so `response_into_chat_response` can normalise it
+    /// into the canonical `reasoning_content` slot; without this, an
+    /// OpenRouter reasoning model that returns its whole answer as
+    /// reasoning surfaces empty `content` AND empty `reasoning_content`
+    /// to the customer. Default so non-OpenRouter upstreams (which omit
+    /// the field) parse unaffected.
+    #[serde(default)]
+    pub reasoning: Option<String>,
 }
 
 // `#[serde(default)]` at the container level so an upstream that omits
@@ -245,13 +256,24 @@ pub fn response_into_chat_response(mut raw: OpenAiResponse) -> ChatResponse {
             // streaming path's `delta.reasoning_content`. Skip empty
             // strings so a model that returns `""` doesn't add a noise
             // field.
-            if let Some(reasoning) = c.message.reasoning_content {
-                if !reasoning.is_empty() {
-                    extra.insert(
-                        "reasoning_content".to_string(),
-                        serde_json::Value::String(reasoning),
-                    );
-                }
+            //
+            // #648: normalise OpenRouter's `message.reasoning` into the same
+            // canonical `reasoning_content` slot. The DeepSeek-canonical
+            // `reasoning_content` takes precedence when present; otherwise we
+            // fall back to OpenRouter's `reasoning`. Without this an
+            // OpenRouter reasoning model that emits its whole answer as
+            // reasoning (empty `content`) reaches the customer with BOTH
+            // fields empty.
+            let reasoning_text = c
+                .message
+                .reasoning_content
+                .filter(|s| !s.is_empty())
+                .or(c.message.reasoning.filter(|s| !s.is_empty()));
+            if let Some(reasoning) = reasoning_text {
+                extra.insert(
+                    "reasoning_content".to_string(),
+                    serde_json::Value::String(reasoning),
+                );
             }
             (
                 ChatMessage {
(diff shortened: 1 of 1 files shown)


moonming:
## Audit remediation (REQUEST-CHANGES → all 3 MEDIUM addressed)

**MEDIUM-1 — non-discriminating tests.** Added a 4th test `non_streaming_empty_canonical_falls_through_to_openrouter_reasoning` (commit `bf3f008`) covering the one branch with real precedence logic: empty canonical `reasoning_content:""` alongside a real `reasoning` must fall through to `reasoning`. Verified it (and `…normalises_to_reasoning_content`) **FAIL** against the pre-fix canonical-only code and pass with the fix — i.e. genuine regression guards. The two original "skip-empty"/"precedence" tests pass either way by design; they're kept as documentation of those rules.

**MEDIUM-2 — scope read as OpenRouter-only.** Added a **"Scope / blast radius"** section to the PR body: `OpenAiResponseMessage` is the shared OpenAI-wire response type (also used by `aisix-provider-azure-openai` and any OpenAI-compatible bridge), so the normalization applies to *any* upstream emitting `message.reasoning`, not just OpenRouter. Called out that it only **fills** the canonical slot when absent/empty (never overrides) and adds no field when `reasoning` is empty/absent → upstreams that never emit `reasoning` are byte-for-byte unaffected.

**MEDIUM-3 — streaming-parity not tracked.** Filed #502 (auto-normalize `delta.reasoning` on the streaming path so neither path needs a per-key `reasoning_field` override) and linked it from the PR body under "Streaming parity (follow-up)".

Local gate: `cargo test -p aisix-provider-openai` → **94 passed / 0 failed**; `cargo clippy --all-targets` + `cargo fmt --check` clean.

PR #509, opened 2026-06-03
fix(stream): stamp TTFT at first delivered token under hold-back (#467)

### LINKED ISSUE #467: build_sse_stream: ttft_ms misreports under streaming hold-back (first-upstream, not first-delivered)
## Summary

Under the P2 streaming hold-back policies (`StreamOutputPolicy::Window` / `BufferFull`, used by `azure_content_safety_text_moderation`), content is withheld from the wire until it scans clean — with the default `window_size = 10_000` chars, nothing is delivered until end-of-stream. But `ttft_ms` is stamped on the **first upstream chunk** carrying content (`crates/aisix-proxy/src/chat.rs`, in the `build_sse_stream` loop, the `first_chunk_seen` block), *unconditionally*.

So for a hold-back stream, `ttft_ms` records time-to-first-**upstream**-token, not time-to-first-**delivered**-token. A response that takes (say) 8 s to its first delivered byte can report `ttft_ms ≈ 200`. This telemetry feeds SLO dashboards / usage events, so the under-report is silent and misleading.

`EndOfStreamCheck` streams (keyword / prompt_shield / no guardrail) are unaffected — they forward live, so first-upstream == first-delivered.

Surfaced by the increment-2 self-audit of api7/ai-gateway#461 (finding M2). The security/correctness behavior of hold-back is unaffected; this is an observability bug.

## Fix

Stamp `ttft_ms` at first **delivery** when holding back: gate the existing first-chunk stamp on `!hold_back`, and stamp on the first event actually released in each `pending.drain(..)` site (mid-stream Window flush, BufferFull fail-open release, end-of-stream release):

```rust
// existing stamp — only the live-forward path:
if !hold_back && !first_chunk_seen && (chunk.delta.content.is_some() || chunk.delta.tool_calls.is_some()) {
    first_chunk_seen = true;
    guard.comp().ttft_ms = started.elapsed().as_millis().min(u32::MAX as u128) as u32;
}
// in each release loop, before the first drained event:
if !first_chunk_seen {
    first_chunk_seen = true;
    guard.comp().ttft_ms = started.elapsed().as_millis().min(u32::MAX as u128) as u32;
}
```

Severity: MEDIUM (observability under hold-back streams; not a correctness or security issue).

jarvis9443:
## Problem

`build_sse_stream` stamped `ttft_ms` on the first **upstream** chunk carrying content/tool_calls, unconditionally. Under a hold-back stream-output policy (`StreamOutputPolicy::Window` / `BufferFull`, used by the Azure text-moderation output guardrail) that content is withheld in `pending` until it scans clean, so `ttft_ms` recorded time-to-first-**upstream**-token (~ms) rather than time-to-first-**delivered**-token — which can be seconds later, or effectively end-of-stream for a full `BufferFull` hold. The metric misreported badly exactly on the path where it matters most.

## Fix

- Gate the upstream-arrival stamp to the live-forward path (`!hold_back`).
- Stamp at the three sites where held content first reaches the wire: window-clean release, `BufferFull` fail-open cap release, and the end-of-stream final release.

Live-forward streams (the default / no-guardrail path) are byte-for-byte unchanged — there, first-upstream == first-delivered. The `messages.rs` SSE builders have no hold-back path, so their existing first-upstream stamp is already correct.

## Tests

- Unit (`chat::ttft_holdback_tests`): drives the real `build_sse_stream` with a `BufferFull` guardrail that allows, an upstream that emits content immediately then idles 150ms before terminating, and asserts the captured `ttft_ms` reflects the end-of-stream release time (≥ ~150ms). Fails before (stamped at arrival → 0ms), passes after.

Fixes #467

### DIFF
--- a/crates/aisix-proxy/src/chat.rs
+++ b/crates/aisix-proxy/src/chat.rs
@@ -2032,7 +2032,17 @@ where
                     // Record TTFT on the first chunk carrying generated
                     // output (content or tool calls). Skip role-only
                     // chunks that OpenAI emits before actual tokens.
-                    if !first_chunk_seen
+                    //
+                    // #467: on the live-forward path the chunk is yielded to
+                    // the wire immediately below, so first-upstream == first-
+                    // delivered. On the hold-back path (Window / BufferFull)
+                    // the chunk is withheld in `pending` until its window (or
+                    // the whole response) scans clean, so stamping here would
+                    // misreport TTFT as time-to-first-*upstream*-token. Gate
+                    // this stamp to the live-forward path and stamp the
+                    // hold-back path at the release (drain) sites instead.
+                    if !hold_back
+                        && !first_chunk_seen
                         && (chunk.delta.content.is_some() || chunk.delta.tool_calls.is_some())
                     {
                         first_chunk_seen = true;
@@ -2177,6 +2187,15 @@ where
                                 // this window's events, then keep the
                                 // trailing overlap as scan context for the
                                 // next window (its events were already sent).
+                                // #467: this is the first content actually
+                                // reaching the wire on the hold-back path —
+                                // stamp TTFT here.
+                                if !first_chunk_seen && !pending.is_empty() {
+                                    first_chunk_seen = true;
+                                    guard.comp().ttft_ms =
+                                        started.elapsed().as_millis().min(u32::MAX as u128)
+                                            as u32;
+                                }
                                 for e in pending.drain(..) {
                                     yield Ok::<_, Infallible>(e);
                                 }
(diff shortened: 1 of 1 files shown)


jarvis9443:
Closing — after checking LiteLLM parity, the premise here is non-standard, so we're keeping the existing upstream-arrival TTFT.

LiteLLM defines TTFT as **upstream arrival**: `completion_start_time` is stamped the moment the first chunk is decoded from the provider, *inside the stream wrapper, before delivery to the client* (`litellm/litellm_core_utils/streaming_handler.py:2077`), and the Prometheus `litellm_llm_api_time_to_first_token_metric` measures `api_call_start_time → completion_start_time`. Output-side buffering/hold-back is layered above the stamp point (`proxy_server.py:7024`), so a guardrail that buffers the whole stream does NOT shift LiteLLM's TTFT — it always reflects upstream responsiveness, never the gateway's hold-back delay.

Our existing `ttft_ms` (stamped at the first upstream content chunk) already matches that. This PR would have made it client-delivery under hold-back, diverging from the convention and losing the upstream-responsiveness signal. Per the project's LiteLLM-parity rule, keeping ttft = upstream is correct. Closing #467 as works-as-intended.

Issue #502, opened 2026-06-03
Auto-normalize OpenRouter `delta.reasoning` on the streaming path (no per-key config) — parity with non-stream #501


moonming:
## Context / follow-up to #501

#501 fixed the **non-stream** half of the OpenRouter reasoning gap (api7/AISIX-Cloud#648): `OpenAiResponseMessage` now captures `message.reasoning` and `response_into_chat_response` auto-normalizes it into the canonical `reasoning_content` slot **with no operator configuration required**.

The **streaming** path is not yet at parity. `extract_reasoning_field` (`crates/aisix-provider-openai/src/overrides.rs`) lifts a vendor reasoning field onto `delta.reasoning_content`, but only when an operator has configured a per-key `reasoning_field` override (e.g. `reasoning_field: "delta.reasoning"`). With no override, an OpenRouter reasoning model that streams its chain-of-thought under `delta.reasoning` still reaches the customer with empty `delta.content` and no `delta.reasoning_content`.

This is an **asymmetry**:

| Path | OpenRouter `reasoning` handling |
|------|-------------------------------|
| non-stream (`message.reasoning`) | **auto** (no config) — #501 |
| streaming (`delta.reasoning`) | requires per-key `reasoning_field` override |

## Ask

Make the streaming path auto-normalize `delta.reasoning → delta.reasoning_content` for the OpenRouter adapter (and any OpenAI-compatible aggregator that uses the bare `reasoning` field) **without** requiring a per-key `reasoning_field` override — symmetric with the non-stream behavior #501 introduced.

### Design notes
- `extract_reasoning_field` already does the lift; it currently runs only when the per-key override supplies the source path. The change is to have the OpenRouter adapter default that source to `delta.reasoning` (the way #501 made the non-stream lift unconditional for `message.reasoning`).
- Keep the existing precedence + skip-empty rules: canonical `delta.reasoning_content` wins when both are present; an empty source adds no field.
- Non-OpenRouter upstreams that don't emit `reasoning` are unaffected (the source key is simply absent).

## Verification
- Mock-parity: extend the streaming reasoning mock to emit `delta.reasoning` (no `reasoning_content`) and assert the gateway surfaces `delta.reasoning_content` with **no** per-key override configured.
[...]

PR #517, opened 2026-06-04
fix(openai): surface gpt-4o-audio message.audio on the non-stream response (#684)

moonming:
## Summary

Fixes **api7/AISIX-Cloud#684** (surfaced by #533 PR-K real-chain). gpt-4o-audio chat models (`gpt-audio`) return generated audio at `message.audio` (`{ id, data, transcript, expires_at }`) when the request asks for the audio output modality (`modalities:["text","audio"]` + `audio:{voice,format}`). The customer got a **200 with non-zero completion tokens but NO audio**.

**Root cause** (same class as #466/#648/#501): the request-side `modalities`/`audio` params already pass through to the upstream via the OpenAI bridge's flatten `extra` (`ChatFormat.extra` → `OpenAiRequest.extra`), so OpenAI *does* generate audio — but `OpenAiResponseMessage` was a closed struct that did **not** model `audio`, so it was dropped at deserialize. Exactly how `message.reasoning` was dropped before #501.

## Change
`crates/aisix-provider-openai/src/wire.rs`:
- Add `#[serde(default)] pub audio: Option<serde_json::Value>` to `OpenAiResponseMessage` (raw value — we don't model the sub-shape).
- In `response_into_chat_response`, surface it verbatim into the canonical `message.audio` slot via `ChatMessage.extra` (the same mechanism #501 used for `reasoning_content`). Null/absent audio adds no field.

No request-side change needed — `modalities`/`audio` already forward (the `ChatFormat` doc explicitly notes `audio` for the 4o audio models lands in `extra`). Non-audio upstreams (which omit the field) are byte-for-byte unaffected.

## Test plan
- [x] 3 unit tests in `wire.rs`: audio→`extra` surfacing (**discriminating** — verified it FAILS against the pre-fix struct), absent-audio-adds-no-field, null-audio-adds-no-field.
- [x] Full crate suite: **102 passed**; `cargo clippy --all-targets` + `cargo fmt --check` clean.
- [ ] CI green.

Once this lands in `:dev`, the AISIX-Cloud PR-K (#677) gpt-4o-audio real-chain leg flips from failing to hard-asserting green (audio output round-trips through the DP).

## Release Notes

* **New Features**
  * OpenAI provider now supports audio output from the gpt-4o-audio model.
  * Audio data is properly surfaced when present and omitted when absent or null.

* **Tests**
  * Added comprehensive test coverage for audio field handling in various scenarios.

### DIFF
--- a/crates/aisix-provider-openai/src/wire.rs
+++ b/crates/aisix-provider-openai/src/wire.rs
@@ -187,6 +187,17 @@ pub struct OpenAiResponseMessage {
     /// the field) parse unaffected.
     #[serde(default)]
     pub reasoning: Option<String>,
+    /// gpt-4o-audio chat models return generated audio at `message.audio`
+    /// (`{ id, data, transcript, expires_at }`) when the request asks for the
+    /// audio output modality (`modalities:["text","audio"]` + `audio:{…}`).
+    /// Without modeling it here the field is dropped at deserialize — the same
+    /// closed-struct gap that dropped `reasoning_content`/`reasoning` (#684,
+    /// cf #466/#648). Captured as a raw value (we don't model the sub-shape:
+    /// `id`/`data`/`transcript`/`expires_at`) and surfaced verbatim to the
+    /// client via `ChatMessage.extra`. Default so non-audio upstreams (which
+    /// omit the field) parse unaffected.
+    #[serde(default)]
+    pub audio: Option<serde_json::Value>,
 }
 
 // `#[serde(default)]` at the container level so an upstream that omits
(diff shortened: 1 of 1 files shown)


moonming:
## Independent audit: APPROVE (no HIGH/MEDIUM)

Cold-context audit verified empirically:
- **Wire path**: `extra["audio"]` → `RenderedMessage.extra` (`#[serde(flatten)]`, render.rs) → `message.audio` top-level on the client wire (same path as `tool_calls`/`reasoning_content`). No borrow/move issue.
- **Request side complete**: `modalities`/`audio` forward via `ChatFormat.extra` (flatten, no `deny_unknown_fields`) → `OpenAiRequest.extra` → upstream. Not a half-fix.
- **Discriminating test confirmed** to fail without the lift; 102 crate tests + clippy/fmt clean.
- **Backward compatible**: additive field; Azure/Vertex OpenAI-compat legs unaffected (and now surface audio too — intended). No flatten key collision. No `litellm` string; no leakage.

LOW notes (non-blocking):
- **LOW-2 → filed #518**: the **streaming** path (`delta.audio`) still drops audio (pre-existing gap, parallel to #466→#502 for reasoning). Tracked separately; out of scope here.
- LOW-1: the inner `!audio.is_null()` guard is defensively dead for the literal-`null` wire shape (serde maps `null`→`None` at the outer `Option`); harmless, kept as intent-documenting.
- LOW-3: request-side `modalities`/`audio` serialization is covered transitively by the generic `extra`-flatten test + the PR-K e2e leg.

PR #522, opened 2026-06-05
test(e2e): pin live weight-edit re-takes effect on dispatch (#196 L1)

moonming:
## What

Closes the **L1** gap (config edit → observable behavior change). The sibling `weighted-routing-distribution-e2e` pins that a weighted model's **initial** weights are honored. This pins that a **live edit** to those weights propagates and the weighted scheduler **rebuilds** — a scheduler that cached its weight-wheel on first dispatch and ignored config updates would silently keep serving the old split.

## How (deterministic, no statistics)

weight 0 = excluded (per `routing-strategies-e2e` "weighted picks the positive-weight target"):
1. Create `wr-edit-virtual` weighted `[wr-edit-a: 100, wr-edit-b: 0]` → assert all `BATCH` dispatches hit **A**.
2. `PUT /admin/v1/models/:id` inverting to `[0, 100]`.
3. **Propagation signal**: a virtual probe returning `"served by B"` — *impossible* under the old `[100,0]` config, so it proves the edit is live **and the scheduler rebuilt** before counting.
4. Assert all `BATCH` dispatches now hit **B**.

If the scheduler never rebuilds on a config edit (the regression this targets), step 3's `waitConfigPropagation` times out — surfacing it loudly rather than passing silently.

## Note on a possible real finding

If CI shows step 3 timing out, that's a **real DP bug** (weighted scheduler doesn't rebuild on weight edit) — I'll file it + hold this test, not weaken the assertion. Expected: the DP rebuilds on the etcd-watch snapshot swap (the same path every config-change test relies on), so it should pass.

## Verification

`tsc --noEmit` clean for this file (borrowed `node_modules`; fresh worktree). Authoritative run is CI's isolated `e2e` job (etcd + built DP + in-process mock upstreams — no shared-stack contention).

Refs #196 L1, #127 L1.

### DIFF
--- a/tests/e2e/src/cases/weighted-routing-edit-e2e.test.ts
+++ b/tests/e2e/src/cases/weighted-routing-edit-e2e.test.ts
@@ -1,0 +1,213 @@
+import { createHash } from "node:crypto";
+import OpenAI from "openai";
+import { afterAll, beforeAll, describe, expect, test } from "vitest";
+import {
+  AdminClient,
+  EtcdClient,
+  spawnApp,
+  startOpenAiUpstream,
+  waitConfigPropagation,
+  type OpenAiUpstream,
+  type SpawnedApp,
+} from "../harness/index.js";
+
+// E2E: a LIVE edit to a weighted routing model's weights re-takes
+// effect on the dispatch path (#196 L1, ai-gateway #127 L1).
+//
+// The sibling weighted-routing-distribution-e2e pins that the INITIAL
+// weights are honored. The gap this closes: after the model is live
+// and serving, an operator PATCHes the weights via the admin API, and
+// the change must propagate through the etcd watch and the weighted
+// scheduler must REBUILD — a scheduler that cached its weight wheel on
+// first dispatch and never rebuilt on config update would keep serving
+// the old split, silently ignoring the operator's change.
+//
+// Design is deterministic (no statistics): weight 0 = excluded (see
+// routing-strategies-e2e "weighted picks the positive-weight target").
+//   - Start [wr-edit-a: 100, wr-edit-b: 0]  → every dispatch hits A.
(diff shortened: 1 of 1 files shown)


moonming:
## Independent audit (CLAUDE.md §8) — CLEAR

A cold third-party agent reviewed against the actual DP scheduler source. **No HIGH/MEDIUM.**

- **Weight-0 = exact exclusion** — `weighted_pick` (routing.rs) uses strict `pick < acc`; a 0-weight target is unreachable. Unit-pinned (`weighted_pick_zero_weight_target_in_middle_is_never_picked`, 2000 trials) + the sibling routing-strategies test. So "all A"/"all B" can't flake on selection.
- **No fall-forward to the 0-weight target** — the dispatch loop only advances on a *retryable* failure; the mock returns 200, which stops dispatch before the in-list 0-weight fallback is contacted. Single dispatch/request, `bDelta==0`/`aDelta==0` guaranteed.
- **PUT body schema-valid + can't 4xx** — `update_model` requires only `display_name`; `weight: minimum 0`; same shape as the create that already validated; `assert_unique_name(..., Some(id))` excludes self. Revision is bumped → etcd watch fires.
- **Propagation signal sound** — "served by B" is provably impossible under the old `[100,0]` config; no partial-snapshot window yields it early; the wait fully gates phase 2; baselines snapshotted after the probe so it isn't counted.
- **Catches the regression loudly** — a scheduler that ignored the edit → probe stays A → `waitConfigPropagation` throws (30s). No silent-pass path.
- **Cleanup + isolation** — upstreams + app closed; unique etcd prefix/ports/admin-key/display-names/caller-key → safe under `maxForks=2` alongside the sibling.

Two LOWs (readiness gate + error-swallowing) are established harness idioms, no change. Verdict: safe to merge.

Issue #271, opened 2026-05-14
docs: write a 'rotate provider key without client impact' tutorial, backed by an e2e test


moonming:
## Context

`docs/tutorials/rotate-provider-key-without-client-impact.md` was added as a 24-line skeleton in #249 but did not include executable steps or a verification section. It was removed in the docs-rebuild-batch-1 follow-up because no existing e2e test pins the "rotation under live traffic" contract — writing the tutorial without that anchor would risk drift between the documented steps and real product behavior.

## What we need

1. An e2e test in `tests/e2e/src/cases/provider-key-rotation-e2e.test.ts` that:
   - creates a \`ProviderKey\` PK1, a \`Model\` pointing at PK1, an \`ApiKey\` allowed to use that model
   - sends N chat requests, asserts all succeed
   - issues \`PUT /admin/v1/provider_keys/:id\` to update PK1's \`secret\` (and optionally \`api_base\`) — same \`id\`, different upstream credential
   - sends another N requests with the same caller bearer + same model alias
   - asserts: all requests succeed, revision on the resource bumped, the caller-facing key + model alias were never re-issued
2. Once the test lands, restore \`docs/tutorials/rotate-provider-key-without-client-impact.md\` to a Kong-cookbook-tier walkthrough modeled on the test (see \`docs/tutorials/build-a-virtual-model-with-failover.md\` for the shape).

## Why not write the tutorial against just the resource API

\`PUT /admin/v1/provider_keys/:id\` exists and the handler tests cover the update path, but "client traffic stays intact across rotation" is a cross-component invariant (admin write → etcd watch → proxy snapshot → in-flight dispatch). Only an e2e test exercises that full path.

## Reference

- Mechanism: \`crates/aisix-admin/src/provider_keys_handlers.rs::update_provider_key\`
- Propagation: \`crates/aisix-etcd/\` watch supervisor + \`aisix-server\` snapshot apply
- Verification model: same shape as \`cache-policy-e2e.test.ts\` and \`routing-strategies-e2e.test.ts\`

## Definition of done

- e2e test added and green in CI
- \`docs/tutorials/rotate-provider-key-without-client-impact.md\` restored, Kong-tier shape, citing the new e2e test as the contract anchor

moonming:
## Finding: "zero in-flight disruption" on provider_key rotation is architecturally guaranteed

While building an e2e for this (PR #523), an independent audit traced the DP rotation path against source and found there is **no in-flight disruption surface to test**:

[...]

Issue #220, opened 2026-05-10
bug: chat completions silently drop `message.tool_calls` on the non-streaming path too


moonming:
## Summary

Companion to [#202](https://github.com/api7/ai-gateway/issues/202) (streaming \`tool_calls\` deltas dropped). The bug is **wider** than that issue assumed:

> Non-streaming chat completions ALSO drop \`message.tool_calls\` end-to-end. \`crates/aisix-provider-openai/src/wire.rs:43\` is explicit — the gateway-internal types don't model \`tool_calls\` AT ALL: "doesn't model (\`tool_calls\`, \`refusal\`, \`audio\`, …)". So the non-streaming \`ChatResponse\` round-trip also strips them.

When the upstream returns:

\`\`\`json
{
  "choices": [{
    "message": {
      "role": "assistant",
      "content": null,
      "tool_calls": [{"id":"call_abc","type":"function","function":{"name":"get_weather","arguments":"{\"location\":\"Beijing\"}"}}]
    },
    "finish_reason": "tool_calls"
  }]
}
\`\`\`

the caller observes \`message.tool_calls === undefined\` on the response. The \`finish_reason\` may also be lost (verify; the gateway-emitted shape needs inspection).

## Repro

E2E test attached: \`tests/e2e/src/cases/tool-calls-non-streaming-e2e.test.ts\` at PR #216 (commit e68e5b8). Mock upstream emits a canned non-streaming chat completion with a single \`message.tool_calls\` entry. Caller fires \`client.chat.completions.create({tools: [...], tool_choice: "auto"})\`. The \`expect(toolCalls).toBeDefined()\` assertion fails — the SDK's \`resp.choices[0].message.tool_calls\` is undefined.

CI repro: <https://github.com/api7/ai-gateway/actions/runs/25628534002/job/75227943034>

\`\`\`
AssertionError: expected undefined to be defined
❯ src/cases/tool-calls-non-streaming-e2e.test.ts:219:23
\`\`\`

## Why this matters

Tool-using chat completions on the **non-streaming** path is the most common deployment shape today (every Python tutorial that demos function calling starts non-streaming). Silently dropping \`tool_calls\` means every framework that depends on it — instructor, langchain, openai-agents-py, custom tool routers — sees a 200 response with empty content and no tool invocation, breaking the application logic with no error trail to follow.

[...]

PR #523, opened 2026-06-05
test(e2e): provider_key rotation with zero in-flight disruption (#196 L3, #271)

moonming:
## What

A liveness/smoke pin over the provider_key **in-place rotation under load** path (#196 L3). Under sustained concurrency (8 workers, 80 requests), a mid-stream `PUT /admin/v1/provider_keys/:id` (same id + api_base, new secret) keeps every request serving and bumps the resource revision; the caller's api_key + model alias are untouched.

## Scope honesty (per the #523 audit)

The audit (HIGH) established that **"zero in-flight disruption" is an architectural guarantee, not a falsifiable property** in this DP: there is no per-provider_key upstream client/pool — the DP holds one shared client and reads `pk.secret`/`api_base` **per-request from an atomic ArcSwap snapshot**, with each in-flight request holding its own snapshot `Arc` to completion. So the original "≥1 failure on the swap" framing would have certified a hollow green.

Re-scoped accordingly: this is a **smoke/liveness pin** (catches a future regression that wedged dispatch or broke watch-apply on a PK PUT) + a revision-bump check, with the architecture documented in the header. It does **not** claim to probe a teardown race (there isn't one).

## The real remaining facet → #220 (NOT closed here)

Asserting the rotated secret actually reaches upstream (old rejected / new accepted) needs a credential-sensitive mock — that's the pre-existing #220. This PR refs #271/#220 but does not close them; see the #271 comment for the architectural finding.

## Verification

`tsc --noEmit` clean for this file. Isolated DP `e2e` CI is authoritative. `maxRetries:2` immunizes the strict all-succeed gate against a loopback transient (a real wedge fails all retries too).

Refs #196 L3, #127 L3, #271, #220.

## Summary by CodeRabbit

* **Tests**
  * Added an end-to-end test that verifies provider key rotation causes zero disruption to in-flight API requests under sustained concurrent load.
  * Confirms all requests complete successfully with no errors during rotation and total successes match expected count.
  * Validates that the provider key rotation is recorded (revision increases) and completes as expected.

### DIFF
--- a/tests/e2e/src/cases/provider-key-rotation-e2e.test.ts
+++ b/tests/e2e/src/cases/provider-key-rotation-e2e.test.ts
@@ -1,0 +1,194 @@
+import { createHash } from "node:crypto";
+import OpenAI from "openai";
+import { afterAll, beforeAll, describe, expect, test } from "vitest";
+import {
+  AdminClient,
+  EtcdClient,
+  spawnApp,
+  startOpenAiUpstream,
+  waitConfigPropagation,
+  type OpenAiUpstream,
+  type SpawnedApp,
+} from "../harness/index.js";
+
+// E2E: provider_key rotation with ZERO in-flight disruption
+// (#196 L3, ai-gateway #271).
+//
+// Customer story: an operator rotates a leaked upstream credential by
+// PUT-ing a new `secret` onto the existing provider_key (same id, same
+// api_base). Live traffic must keep flowing — every request issued
+// before, during, and after the rotation succeeds. A DP that rebuilt
+// its per-provider_key upstream client non-atomically (briefly losing
+// the key, or dropping in-flight requests on the snapshot swap) would
+// surface here as ≥1 failed request.
+//
+// What this pins: under sustained concurrency, an in-place secret
+// rotation (PUT /admin/v1/provider_keys/:id) causes zero failed
+// requests and bumps the resource revision. The caller's api_key and
+// the model alias are never touched, so a real client keeps using the
+// same bearer throughout.
+//
+// Out of scope (separate gap #220): asserting WHICH secret the DP
+// dialed upstream — the mock upstream ignores the credential, so the
+// rotation's effect on the wire isn't observable here. This test pins
+// the disruption-free property, which is L3's distinguishing claim.
+//
+// Reference: OpenAI Chat Completions shape the caller sees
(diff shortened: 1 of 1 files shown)


moonming:
## Independent audit (CLAUDE.md §8) — resolved

Verdict **BLOCK** on HIGH-1; addressed by re-scoping (latest commit). The auditor verified against DP source (`bridge.rs` one shared client; `dispatch.rs`/`chat.rs` per-request snapshot reads; `snapshot.rs` ArcSwap; `supervisor.rs` CAS watch-apply) that **there is no per-provider_key client/pool to tear down** — `pk.secret`/`api_base` are read per-request from an atomic snapshot, in-flight requests hold their own Arc. So "≥1 failure on hot-swap" is impossible by construction and the old framing certified a hollow green (the §7 "claims more than it verifies" gap).

**Fixed**: header + PR + test title re-scoped to a rotate-under-load **liveness/smoke pin + revision bump**, with the architecture documented; the real "rotated secret reaches upstream" facet stays as **#220** (credential-sensitive mock needed), not closed here. Also bumped `maxRetries 0→2` (audit LOW: immunize the strict gate against a loopback transient — a real wedge fails all retries too).

Auditor confirmed CLEAR on the mechanics: valid PUT body, correct revision bump (1→2), real in-flight overlap during the PUT, correct isolation/cleanup, `rotated` guard genuinely asserted. MEDIUM-1 (overlap density) only mattered under the dropped disruption-probe framing.

moonming:
## CI: my test passed; failure is an unrelated cache flake — re-running

The `e2e` job went red, but **this PR's test passed**: `✓ src/cases/provider-key-rotation-e2e.test.ts (1 test) 1792ms`.

The failure is in an unrelated file — `cache-policy-e2e.test.ts › second identical request is served from cache, upstream not re-hit` → `AssertionError: expected 3 to be 2` (the upstream was hit 3× instead of 2× — a cache-warm/timing flake, the cache write hadn't landed before the second request). This PR only **adds** `provider-key-rotation-e2e.test.ts`; it touches no caching code, and under `maxForks=2` each test runs its own `spawnApp` with a unique etcd prefix + its own DP/cache, so this PR cannot affect another test's cache-hit count. Pre-existing flaky-shaped failure (timing), not a regression here.

Re-enqueued the failed job. If `cache-policy-e2e` recurs it's a flaky cache test to harden separately (not this PR).

Issue #527, opened 2026-06-05
e2e: pre-existing `tsc --noEmit` errors in cache-scenarios / request-id-uuid tests


moonming:
`pnpm exec tsc --noEmit` in `tests/e2e` reports 4 pre-existing type errors, unrelated to current feature work. Recording here; not fixing now.

## Errors
```
src/cases/cache-scenarios-e2e.test.ts(96,36): error TS18048: 'app' is possibly 'undefined'.
src/cases/request-id-uuid-e2e.test.ts(124,20): error TS18048: 'app' is possibly 'undefined'.
src/cases/request-id-uuid-e2e.test.ts(128,20): error TS18048: 'app' is possibly 'undefined'.
src/cases/request-id-uuid-e2e.test.ts(132,20): error TS18048: 'app' is possibly 'undefined'.
```

## Notes
- `strictNullChecks` narrowing gaps: `app` (a `SpawnedApp | undefined` assigned in `beforeAll`) is used in the test body without a non-null guard. Most other e2e cases gate on `if (!app) { ctx.skip(); return; }` first.
- vitest runs via esbuild (no typecheck at runtime), so these do not fail the test run — only `tsc --noEmit` surfaces them. CI does not currently run `tsc --noEmit` on the e2e harness.
- Suggested fix: add the missing `if (!etcdReachable || !app) { ctx.skip(); return; }` guard (or a local non-null assertion) in those two files.

Surfaced while adding the `aliyun_sls` exporter e2e (which typechecks clean); deferred by decision.

jarvis9443:
`tsc --noEmit` in `tests/e2e` reported 12 errors, mostly narrowing lost inside async closures. They're fixed, and CI now runs the typecheck so they can't come back. Fixed in https://github.com/api7/aisix/pull/1233.

PR #528, opened 2026-06-05
feat(obs): pluggable observability-sink framework + Aliyun SLS exporter

moonming:
## Summary

Introduces a pluggable observability-sink framework and lands its first real vendor — **Aliyun SLS** — so the DP can ship per-request telemetry (and, later, full prompt/response) to customer log/analytics backends without a bespoke connector per vendor.

Driven by the pre-sales requirement on api7/AISIX-Cloud#687 (customer wants AI Gateway logs in Aliyun SLS) and the generic-integration design in api7/AISIX-Cloud#692.

### Framework (one implementation per *family*, not per vendor)
- `ObservabilitySink` trait + `SinkCapabilities` matrix (idempotency / ordering / batch-unit / partial-batch / streaming) so the shared pipeline adapts to a sink's wire contract as **data, not a fork**.
- `SinkPipeline` — bounded mpsc → batch → retry/backoff → drop-with-metric, one per exporter.
- `ExporterPipelines` manager — lazy `get_or_create` (immediate consistency) + `retain` GC.
- The existing OTLP/HTTP fan-out is migrated onto this pipeline (`OtlpSink`), behaviour-preserving (the `usage-client-source` e2e is the regression gate).

### Aliyun SLS exporter
- `AliyunSlsSink` (http_batch family): one protobuf `LogGroup` per batch → lz4 **block** compression → SLS signature **v1 (HMAC-SHA1)** → `POST /logstores/<ls>/shards/lb`. Built on the official `aliyun-log-sdk-protobuf` / `aliyun-log-sdk-sign` sub-crates over the existing **rustls** reqwest — **no OpenSSL / native-tls** is pulled in.
- `ExporterKind::AliyunSls` carries a **`credential_ref`, never the AccessKey** — the plaintext key never rides the kine path. The DP resolves the ref from its local env (`SLS_CRED_<REF>_AK_{ID,SECRET}`, deliberately **not** the `AISIX_` prefix the config loader claims). BYOK variants slot into the same resolution point later.
- `fan_out` generalized to dispatch per kind over one shared `ExporterPipelines` + client; **all 7 proxy call sites unchanged**.

## Key design decisions
- **No plaintext credentials in etcd** — `credential_ref` + local env resolution; `rejects_plaintext_credentials_in_config` pins it at the serde + schema layers.
- **Field mapping serializes `UsageEvent`** (single-sources the schema + its `skip_serializing_if` rules) rather than hand-listing 35 fields; opt-in prompt/response land as flat fields, absent on the default metadata-only path so prompts cannot leak.
- **Error classification** mirrors `OtlpSink` (5xx/408/429 → transient) **plus** SLS `errorCode`-aware promotion, so a 4xx quota/busy/clock-skew is retried, not dropped (a log sink must not drop under throttling).
- Endpoint handling: a bare region host → `https://<project>.<host>`; a scheme-qualified loopback (the e2e mock) is used verbatim.

## Test plan
- **Unit**: `aisix-core` 204 passed, `aisix-obs` 75 passed (sink encode/sign/lz4 round-trip decode, error classification incl. quota→transient, credential resolver, schema validators). clippy `-D warnings` clean; `schemas/` regenerated (schema-drift gate).
- **L2 mock e2e** (`aliyun-sls-exporter-e2e`): a dashboard-configured `aliyun_sls` exporter makes the **real `aisix` binary** post a signed lz4-protobuf PutLogs to a mock SLS receiver on a real chat — asserts logstore path, SLS headers, lz4 framing, and that `Authorization` carries the **env-resolved** AccessKey id (secret never on the wire). ✅ passing.
- **OTLP regression gate** (`usage-client-source-e2e`): ✅ 2/2 — confirms the per-kind dispatch generalization preserved the OTLP path.
- **Real-SLS smoke** (`sls_smoke`, `#[ignore]` + env-gated): validates the v1 signature / lz4 / protobuf against the **live** SLS endpoint (the thing a mock cannot check). Runs in the new `sls-smoke` CI job with the SLS secrets injected; fork PRs skip cleanly. PutLogs→2xx is the hard assertion (Permanent error fails fast, Transient retries); GetLogs readback is best-effort with diagnostics pending its first real run.

## Deliberately out of scope (follow-ups)
- **Content capture** (full prompt/response into SLS — the customer's primary ask): the sink already emits content when a record carries it; the hot-path capture + size-cap + per-exporter gating is the next milestone. Current fan-out is metadata-only (parity with OTLP).
- **CP-side `aliyun_sls` support** (AISIX-Cloud / Go) — the DP accepts the kind via admin API + kine now; the dashboard/CP flow is separate per #687 ("DP first, then CP").
- Type rename `OtlpHttpFanOut → ExporterFanOut` and wiring the fan-out GC/shutdown loop — mechanical follow-ups kept out to avoid churn.
- 4 pre-existing `tsc --noEmit` errors in unrelated e2e files tracked in #527 (not touched here; my additions typecheck clean).

Refs: api7/AISIX-Cloud#687, api7/AISIX-Cloud#692

## Summary by CodeRabbit

* **New Features**
  * Added Aliyun SLS as an observability exporter backend with secure credential references and pluggable sink support.
  * Introduced resilient delivery pipelines with batching, retry, backpressure, and per-sink health/stats.

* **Tests**
  * Added CI smoke test and end-to-end test validating Aliyun SLS export behavior and headers.

* **Chores**
  * Updated CI to run the new SLS verification.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -184,6 +184,33 @@ jobs:
           if-no-files-found: ignore
           retention-days: 7
 
+  sls-smoke:
+    name: aliyun sls smoke (real)
+    runs-on: ubicloud-standard-2
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: arduino/setup-protoc@v3.0.0
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - uses: Swatinem/rust-cache@v2
+      # Real Aliyun SLS PutLogs validation — the v1 signature / lz4 block /
+      # protobuf a mock receiver cannot check. The test is `#[ignore]` +
+      # env-gated, so it only does real work when the SLS secrets are present;
+      # fork PRs (where GitHub withholds secrets) see empty vars and skip
+      # cleanly. NOTE: `AISIX_E2E_SLS_*` is read directly by the test — this is
+      # a `cargo test` process that never loads `Config`, so the `AISIX_`
+      # prefix is safe here (unlike the spawned binary in the e2e job, whose
+      # credentials must use the non-`AISIX_` `SLS_CRED_*` form).
+      - name: real-SLS smoke
+        env:
+          AISIX_E2E_SLS_AK_ID: ${{ secrets.ALI_SLS_AK_ID }}
+          AISIX_E2E_SLS_AK_SECRET: ${{ secrets.ALI_SLS_SK }}
+          AISIX_E2E_SLS_ENDPOINT: ap-southeast-3.log.aliyuncs.com
+          AISIX_E2E_SLS_PROJECT: aisix-e2e-obs
+          AISIX_E2E_SLS_LOGSTORE: request-events
+        run: cargo test -p aisix-obs --lib -- --ignored sls_smoke --nocapture
+
   coverage-gate:
     name: coverage >= 90%
     needs: [rust-unit, e2e]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -179,9 +179,13 @@ name = "aisix-obs"
 version = "0.1.0"
 dependencies = [
  "aisix-core",
+ "aliyun-log-sdk-protobuf",
+ "aliyun-log-sdk-sign",
  "anyhow",
  "async-trait",
  "base64 0.22.1",
+ "http 1.4.0",
+ "lz4_flex",
  "metrics",
  "metrics-exporter-prometheus",
  "opentelemetry",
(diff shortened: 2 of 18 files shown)


moonming:
## Independent audit (CLAUDE.md §7) — results + dispositions

A fresh general-purpose agent reviewed this PR cold (no shared context), verifying the wire/secret/regression claims against the actual Aliyun SLS SDK source, not just the description. **No HIGH findings.** Wire-signing order, lz4-block choice, `credential_ref` isolation, secret hygiene, and OTLP regression-safety all verified at source level; unit tests + clippy `-D warnings` green.

### MEDIUM
- **M1 — `max_batch_bytes` advertised but enforced nowhere → resolved in code (a8b40ae).** The capability claimed a 3 MB ceiling, but the pipeline batches by record count and the sink encodes the whole batch into one PutLogs. Harmless on today's metadata-only path (~100 × ~1 KB), but a silent `Permanent`-drop once content capture makes a batch exceed SLS's body limit. Fixed by downgrading the SLS capability to its honest current state (`batch_unit: Records`, `max_batch_bytes: None` — parity with `OtlpSink`). The real fix — byte-aware chunking in `append_batch` (per the pipeline's sink-self-chunks design) + reconciling the contradictory capability docs — is filed as **#529 and blocks the content-capture milestone**.

### LOW (non-blocking)
- **L1 — `is_transient_error_code` omits a `RequestTimeout` body code.** Accepted, no change: an HTTP 408 is already promoted by the status arm, and a body `errorCode:"RequestTimeout"` on a non-408 status is unattested in the SLS error reference — adding an unverified code conflicts with our research-discipline rule. Will revisit if a real run surfaces it.
- **L2 — description says "7 proxy call sites", there are actually 8** (`completions, chat, messages, responses, embeddings, rerank, audio, images`). All 8 are identical in shape and **unchanged** (not in the diff); `fan_out`'s signature is preserved. Cosmetic doc inaccuracy, corrected here.
- **L3 — SLS log `__time__` is ingest-`now()`, not `occurred_at`.** Accepted, no change: this is the defensively-correct choice (avoids SLS time-skew rejection on delayed retries); the real event time survives as the `occurred_at` content field. The auditor concurred it's intentional, not a bug.

Bottom line from the audit: safe to merge after M1, which is now addressed. Remaining items are LOW and justified above.

Issue #529, opened 2026-06-05
obs sink: byte-aware chunking for SLS PutLogs (blocks content capture)


moonming:
Surfaced by the independent audit of #528 (finding M1).

## Problem
The sink framework's byte-ceiling is **advertised but enforced nowhere**:
- `SinkCapabilities.max_batch_bytes` is set by sinks but read by no code. `SinkPipeline` batches purely by record count (`max_batch`, default 100) + flush interval.
- The docs contradict each other on *who* enforces it:
  - `capabilities.rs` (`max_batch_bytes`): "The pipeline never hands the sink a batch larger than this."
  - `pipeline.rs` module doc: "the sink owns chunking a batch down to its own per-request byte limit inside `append_batch`."
  - Reality: neither happens.

`AliyunSlsSink::append_batch` encodes **all** records into a single `LogGroup` → one PutLogs. SLS caps the request body (single LogGroup, low single-digit MB). On today's **metadata-only** path this is fine (≈100 records × ~0.5–2 KB ≪ limit). But the next milestone — full prompt/response **content capture** (`SinkContent`) — makes a 100-record batch easily exceed the limit, and SLS then returns `400 PostBodyInvalid`, which classifies as `SinkError::Permanent` → the whole batch is **dropped, never retried**. A log sink silently dropping under load is exactly the failure the sink's error-classification is meant to avoid.

## Fix (do this before content capture ships)
1. Implement byte-aware chunking in `AliyunSlsSink::append_batch` (per the `pipeline.rs` design — the sink self-chunks, since only the sink knows the encoded size): pack records into multiple PutLogs requests, each kept under a conservative raw-size ceiling (~3 MB), always sending at least one record per request. Fast-path the common case (whole batch fits → one encode + one POST); only split when oversize.
2. Reconcile the contradictory capability docs (`capabilities.rs` vs `pipeline.rs`) to a single source of truth: the sink self-limits.
3. Add a unit test that drives a synthetic >ceiling batch (e.g. one large `SinkContent`) and asserts it fans out across multiple signed PutLogs requests.
4. Restore the SLS capability to `batch_unit: Both` + `max_batch_bytes: Some(...)` once (1) lands.

## Interim (in #528)
The SLS capability was downgraded to the honest current state — `batch_unit: Records`, `max_batch_bytes: None` (parity with `OtlpSink`) — so it no longer advertises an unenforced ceiling. This issue **blocks** the content-capture milestone.

Refs: #528, api7/AISIX-Cloud#687

PR #531, opened 2026-06-06
feat(obs): object_store sink — one S3/GCS/Azure exporter on the shared pipeline

moonming:
## What

Adds **`ExporterKind::ObjectStore`** — a single `ObjectStoreSink` that delivers batched AI-gateway telemetry as NDJSON objects to **S3, GCS, and Azure Blob** (plus S3-compatible MinIO / Cloudflare R2), built **once** on the shared observability-sink pipeline (#528 / AISIX-Cloud#692). The only per-provider code is `build_object_store` — the `object_store` crate's builders own SigV4 / GCS OAuth / Azure shared-key signing, so we hand-roll **zero** signers.

Refs AISIX-Cloud#689 (the S3 / object-storage observability sink).

## Why this shape (reference implementations — CLAUDE.md §7)

Surveyed how the field does multi-cloud blob export (official docs + source):
- **A major open-source LLM proxy**: four *separate* per-provider logging integrations (boto3 / hand-rolled SigV4 / hand-rolled GCS REST / ADLS) — the "write it three times" anti-pattern.
- **Langfuse**: clean seam (one `StorageService` interface + factory) but still three cloud SDKs — the JS ecosystem has no unified blob lib.
- **Cloudflare Logpush**: cleanest *contract* — one URI whose scheme picks the provider.
- **Rust data-infra** (InfluxDB 3.0, DataFusion, Lance, delta-rs) standardizes on the **`object_store`** crate: one `ObjectStore` trait over S3/GCS/Azure. InfluxDB 3.0 serves all three clouds through it with **no separate SDKs** — the existence proof.

Divergence: in Rust we collapse the per-provider duplication into **one** code path (what JS/Python could not). Chosen `object_store` over `opendal` because it's first-party with the `parquet`/Arrow stack (the Parquet phase becomes free) and is the Arrow-ecosystem standard. rustls only, **no OpenSSL** (`tls-webpki-roots`).

## Design

- **Batched NDJSON (+gzip), one object per flush.**
- **Content-addressed keys** `…/dt=YYYY-MM-DD/hh=HH/<sha256(body)[..32]>.ndjson[.gz]`: a pipeline retry of the same batch maps to the same key, so a downstream **Snowpipe / Auto Loader dedups by filename and never double-loads** (AISIX-Cloud#689 §8) — achieved without needing offset/FileSeq-marker plumbing yet.
- **Credentials never on the kine path**: `credential_ref` resolved DP-side from env (`OBJSTORE_CRED_<REF>_*`), mirroring the SLS sink. A missing credential or un-buildable backend yields a `BrokenSink` that reports the reason via delivery health — never a silent drop.
- Loader JSON-schema validation (provider enum, required fields, https-or-loopback endpoint) + regenerated `schemas/resources/observability_exporter.schema.json`; fan-out arm + config fingerprint.

## Verified

- `cargo test -p aisix-core` (213) + `-p aisix-obs` (89) green; `clippy -D warnings` (all targets) clean; `cargo fmt --check` clean; full-workspace `cargo check` green.
- **No OpenSSL / native-tls** in the dep graph after adding `object_store` (verified via `cargo tree`).
- No schema drift (only `observability_exporter.schema.json` regenerated).
- **Real-backend e2e: S3 round-trip verified live against MinIO** — real SigV4 + PUT + GET + gzip round-trip + content-addressed retry-dedup (one object after re-delivery) + cleanup. Run via `crates/aisix-obs/tests/object-store-emulators.compose.yml`.
- The GCS smoke caught + fixed a real bug (`with_url` misuse for the emulator endpoint).

## Test plan

- Unit (hermetic): encode/key/dedup/error-map/credential-resolve/schema against `object_store::memory::InMemory`.
- Smoke (`#[ignore]` + env-gated, real emulators): `objstore_smoke_{s3,azure,gcs}_roundtrip`. **S3/MinIO greens locally.** Azure/Azurite is ready (needs the container pre-created). GCS round-trip needs **real GCS** — fake-gcs's XML API doesn't round-trip `object_store`'s percent-encoded object names (documented in the test + compose).

## Deferred (not in this PR)

Parquet output; restart-stable `FileSequence` markers; org/env partitioning; size/age rollover + multipart tuning; CP (Go) `cpapi` support; dashboard multi-kind form; exporter-credential field-level encryption (the pre-existing plaintext-in-etcd dependency).

## Summary by CodeRabbit

* **New Features**
  * Added an object-storage exporter sink for S3, GCS, and Azure Blob with NDJSON output, optional gzip compression (default gzip), deterministic content-addressed keys, and Hive-style dt/hh partitioning.
  * Support for credential references (env-driven) and configurable region/endpoint (allows localhost/emulator HTTP).
  * Schema and config validation updated to include the object_store kind.

* **Tests**
  * Added unit tests and ignored emulator smoke tests plus a Docker Compose setup for local S3/GCS/Azure emulators.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -184,10 +184,15 @@ dependencies = [
  "anyhow",
  "async-trait",
  "base64 0.22.1",
+ "chrono",
+ "flate2",
+ "futures",
+ "hex",
  "http 1.4.0",
  "lz4_flex",
  "metrics",
  "metrics-exporter-prometheus",
+ "object_store",
  "opentelemetry",
  "opentelemetry-otlp",
  "opentelemetry-semantic-conventions",
@@ -196,6 +201,7 @@ dependencies = [
  "reqwest",
  "serde",
  "serde_json",
+ "sha2 0.10.9",
  "thiserror 1.0.69",
  "tokio",
  "tracing",
@@ -1293,6 +1299,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
 
 [[package]]
+name = "chacha20"
+version = "0.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6f8d983286843e49675a4b7a2d174efe136dc93a18d69130dd18198a6c167601"
+dependencies = [
+ "cfg-if",
+ "cpufeatures 0.3.0",
+ "rand_core 0.10.1",
+]
+
+[[package]]
 name = "chrono"
 version = "0.4.44"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2045,6 +2062,7 @@ dependencies = [
  "cfg-if",
  "libc",
  "r-efi 6.0.0",
+ "rand_core 0.10.1",
  "wasip2",
  "wasip3",
 ]
@@ -2807,6 +2825,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0e7465ac9959cc2b1404e8e2367b43684a6d13790fe23056cc8c6c5a6b7bcb94"
 
 [[package]]
+name = "md-5"
+version = "0.10.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d89e7ee0cfbedfc4da3340218492196241d89eefb6dab27de5df917a6d2e78cf"
+dependencies = [
+ "cfg-if",
+ "digest 0.10.7",
+]
+
+[[package]]
 name = "md5"
 version = "0.7.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


moonming:
## Independent audit (cold, no shared context) — resolved

An independent agent reviewed this PR against the 6 angles. Result: **0 HIGH, 2 MEDIUM, 3 LOW**. Both MEDIUMs are fixed in code; LOWs handled/justified.

**MEDIUM-1 — `capabilities()` overstated idempotency (`FileSequence` vs at-least-once).** Fixed: now declares `IdempotencyScheme::None`. Content-addressed keys dedup an *in-process* retry by filename, but are **not** a restart-stable monotonic sequence (a re-batch after a DP restart yields a new key). Module doc + capability comments corrected to scope the "no double-load" guarantee to in-process retry; restart-stable `FileSequence` markers stay a tracked follow-up.

**MEDIUM-2 — unbounded in-RAM encode+gzip once content capture is enabled.** Fixed by documenting + tracking it (mirrors the SLS sink's #529 note): `capabilities()` now explains the body is count-bounded today and byte-aware chunking is required before full prompt/response content rides these records (rollover/chunking follow-up). No change on the default metadata-only path (stays small).

**LOW-2 — region doc said "required" but `object_store` defaults to us-east-1.** Fixed: softened to "recommended; defaults to us-east-1" (a wrong region surfaces as a permanent delivery-health error, not a silent drop).

**LOW-1 / LOW-3 — justified, not blocking.** LOW-1 (endpoint allows any https host) is defense-in-depth only — `object_store` issues *signed* requests and the endpoint is operator-supplied at the same trust level as the bucket; the plaintext-SSRF vector was tested and is correctly rejected by the schema regex. LOW-3 (GCS `endpoint` silently ignored) is cosmetic — the GCS override rides the service-account `gcs_base_url`, per the code comment.

**Follow-ups (deferred cut-2 items):** (a) run the Azure/Azurite smoke in CI (compose already has the container init); (b) byte-aware rollover/chunking for content capture; (c) restart-stable `FileSequence` markers.

Audit also confirmed: no-plaintext-secret holds at **both** serde (`deny_unknown_fields`) and loader JSON-schema (`additionalProperties:false`) layers; rustls-only / no OpenSSL; flat kind-tagged wire = non-breaking; metadata-only default cannot leak content; emulator env = well-known local defaults only (no real secret).

PR #534, opened 2026-06-07
feat(obs): per-attempt UsageEvent emission for Model-Group routing (#655)

jarvis9443:
## Problem

A Model-Group request that retries or fails over made multiple upstream attempts, but the DP emitted a single aggregated `UsageEvent` per request — the per-attempt detail (which target was tried, why it failed, how long it took) was flattened into a `routing_attempts` JSON blob and `ttft_ms`/`latency_ms` covered only the winning call. This diverges from how LiteLLM logs routing (one log record per call, linked by a trace id) and makes per-attempt latency/error analysis impossible downstream.

## What changed

Each upstream attempt — the initial try, each same-target retry, and each fallback — now emits its own `UsageEvent`, all sharing the request's `request_id` (the trace/group key) and ordered by a new `attempt_index`. Failed attempts carry zero tokens plus a bounded `error_class`/`error_message`; the winning attempt carries the real tokens/cost. `status_code`/`latency_ms`/`ttft_ms` are scoped to the attempt.

- New `attempt` module holds the shared `AttemptRecord`/`RoutingTelemetry`/`AttemptInfo` types and the error-class helpers, so `/v1/chat/completions`, `/v1/messages`, and `/v1/responses` classify and emit attempts identically and can't drift.
- All three Model-Group endpoints emit per-attempt events; the `/v1/responses` error path previously emitted nothing and now surfaces failed attempts.
- The one-line-per-request access log is unchanged (it keeps a served-by + attempt/fallback summary); per-attempt detail lives only in telemetry, matching LiteLLM's transport-log vs. structured-log split.
- The OTLP span exports the per-attempt attributes (`aisix.attempt_index`/`_kind`/`_model`/`error_class`/`error_message`).

New `UsageEvent` fields: `attempt_index`, `attempt_kind` (`initial`/`retry`/`fallback`), `attempt_model`, `error_class`, `error_message`. Removed: `served_by_model`, `routing_attempt_count`, `routing_fallback_count`, `routing_attempts` (and the `RoutingAttemptEvent` type).

## Compatibility

Breaking wire-shape change paired with the cp-api ingestion change (api7/AISIX-Cloud#655). Per the project's cross-plane policy this DP change lands first so a new `dev` image is available; the brief window where the not-yet-updated cp-api can't parse the new shape is acceptable and not worked around.

## Tests

- Per-attempt fallback integration tests for the chat, messages, and responses families (failed initial + successful fallback share one `request_id`, with the right per-attempt fields).
- OTLP unit test for the new span attributes.
- DP standalone E2E (`tests/e2e/per-attempt-telemetry-e2e`) drives a primary-502 → secondary-200 Model Group and asserts two attempt spans sharing the request id.

## Summary by CodeRabbit

* **New Features**
  * Emit per-upstream-attempt telemetry for routing/failover: each attempt reports index, kind (initial/retry/fallback), target model, error classification/message, and latency. OTLP spans now include attempt-level attributes. Access logs now surface served model plus attempt and fallback counts.

* **Tests**
  * Added E2E and unit tests validating per-attempt telemetry, OTLP span attributes, and updated access-log/usage emission behavior.

### DIFF
--- a/crates/aisix-obs/src/access_log.rs
+++ b/crates/aisix-obs/src/access_log.rs
@@ -20,10 +20,15 @@ pub struct AccessLog<'a> {
     pub completion_tokens: Option<u64>,
     pub total_tokens: Option<u64>,
     pub request_id: &'a str,
+    /// Routing target that ultimately served the request (the winning
+    /// attempt's display name). `None` for direct models / cache hits.
     pub served_by_model: Option<&'a str>,
+    /// Total upstream attempts made (initial + retries + fallbacks).
     pub routing_attempt_count: Option<u32>,
+    /// How many attempts moved to a different target. Per #655 the
+    /// per-attempt detail lives in telemetry (per-attempt UsageEvents),
+    /// not in this one-line-per-request access log.
     pub routing_fallback_count: Option<u32>,
-    pub routing_attempts: Option<&'a str>,
 }
 
 impl AccessLog<'_> {
@@ -47,7 +52,6 @@ impl AccessLog<'_> {
             served_by_model = self.served_by_model,
             routing_attempt_count = self.routing_attempt_count,
             routing_fallback_count = self.routing_fallback_count,
-            routing_attempts = self.routing_attempts,
             "proxy request completed",
         );
     }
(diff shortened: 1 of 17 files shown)


jarvis9443:
Thanks for the review.

- **Duplicated `ms_since` + `attempt_error_from_proxy`**: good catch — hoisted both into `crate::attempt` and removed the copies in `messages.rs`/`responses.rs` (af305c8).

- **`MessagesDispatchError::pre_dispatch` "never called"**: it is called — the `From<ProxyError>` impl delegates to it (`fn from(err) { Self::pre_dispatch(err) }`). That's why `clippy -D warnings` is clean (an actually-unused private method would fail the build). Keeping it as the single documented place for the "no recorded attempts" construction, so no change here.

PR #549, opened 2026-06-08
feat(obs): object_store cloud_identity auth — keyless via the DP's own cloud identity (S3 / GCS)

moonming:
## What

Adds a keyless **`cloud_identity`** auth mode to the `object_store` observability sink. Today the sink resolves `credential_ref` → `OBJSTORE_CRED_<SLUG>_*` static-key env vars on the DP. This adds a second mode where the DP authenticates with its **own attached cloud identity** — no static keys anywhere:

- `auth_mode: credential_ref` (default, unchanged).
- `auth_mode: cloud_identity` — **S3** via the AWS default credential chain (`AmazonS3Builder::from_env` → env / EKS IRSA / ECS task role / EC2 instance profile); **GCS** via Application Default Credentials (GKE Workload Identity / GCE metadata) by building with no service-account key.

`credential_ref` is now `#[serde(default)]` (omittable for cloud_identity). `build_object_store_sink` branches on `auth_mode`; the keyless builders live in a new `build_object_store_ambient`.

**Azure cloud_identity is intentionally out of scope in this cut** — Azure managed identity still needs a non-secret account name the keyless config doesn't carry. `build_object_store_ambient` returns a clear permanent error for `azure_blob`, and the loader schema pins `cloud_identity` to `s3` / `gcs` (tracked for a follow-up).

## Prior art

A **customer-run** data mover authenticating to object storage via its host's **ambient cloud identity** (instance role / IRSA / Workload Identity / metadata SA), with static keys as the opt-in fallback, is the standard pattern across self-hosted log/telemetry shippers, tiered-storage brokers, and backup tooling — they resolve through the cloud SDK's default credential chain (AWS) / ADC (GCP), and several recommend the role over static keys. A primary-source survey across API gateways, message middleware, and databases found ambient identity **universal for AWS** but **spottier for GCP/Azure** — which is exactly why this cut covers S3 + GCS and defers Azure.

## Test plan

- `build_object_store_ambient`: dispatches S3 (`from_env`, builds offline) + GCS (ADC, builds offline); **rejects Azure** with a clear permanent error.
- serde: `auth_mode` defaults to `credential_ref` (back-compat); `cloud_identity` round-trips with `credential_ref` omitted.
- loader schema (`validate_observability_exporter`): `cloud_identity` (s3/gcs) validates **without** `credential_ref`; `cloud_identity + azure_blob` is rejected; `credential_ref` is still required outside cloud_identity.
- `cargo fmt` + `cargo clippy -p aisix-core -p aisix-obs --all-targets -- -D warnings` + `cargo test -p aisix-core -p aisix-obs` → **99 passed, 0 failed**. `dump-schema` regenerated `schemas/resources/observability_exporter.schema.json` (no drift).

## Notes

- **Not emulator-testable**: CI has no real IAM role, so cloud_identity's actual auth can't run against MinIO / Azurite / fake-gcs — the builders are unit-tested for dispatch; real-cloud verification is manual (the same inherent limit as the existing GCS smoke).
- cp-api (accept + validate `auth_mode`, reject `cloud_identity + azure_blob` at create) and the dashboard (auth-mode selector) land in **separate PRs**; this is the DP foundation.

Refs: object_store sink line (DP #531; cp-api `api7/AISIX-Cloud#706`; dashboard `api7/AISIX-Cloud#714`). Design: `api7/AISIX-Cloud#724` (P2 keyless cloud-identity).

### DIFF
--- a/crates/aisix-core/src/models/observability_exporter.rs
+++ b/crates/aisix-core/src/models/observability_exporter.rs
@@ -181,9 +181,16 @@ pub struct ObjectStoreConfig {
     #[serde(default)]
     pub compression: ObjectStoreCompression,
 
+    /// How the DP authenticates to the bucket. Default `credential_ref`.
+    #[serde(default)]
+    pub auth_mode: ObjectStoreAuthMode,
+
     /// Opaque pointer to the cloud credentials, resolved locally by the DP
     /// at delivery time. The plaintext key MUST NOT live in etcd/kine — the
     /// control plane stores only this reference, never the secret itself.
+    /// Required when `auth_mode = credential_ref`; unused (and may be empty)
+    /// when `auth_mode = cloud_identity`.
+    #[serde(default)]
     pub credential_ref: String,
 }
 
@@ -209,6 +216,23 @@ pub enum ObjectStoreCompression {
     Gzip,
     /// No compression — raw NDJSON.
     None,
+}
+
+/// How the DP obtains credentials for the object-storage bucket.
+#[derive(
+    Debug, Clone, Copy, Default, Serialize, Deserialize, schemars::JsonSchema, PartialEq, Eq, Hash,
+)]
+#[serde(rename_all = "snake_case")]
+pub enum ObjectStoreAuthMode {
+    /// Resolve `credential_ref` to static keys from the DP's local env
+    /// (`OBJSTORE_CRED_<SLUG>_<FIELD>`). The default.
+    #[default]
+    CredentialRef,
+    /// Use the DP host's own attached cloud identity — EC2 instance role /
+    /// EKS IRSA / ECS task role (S3), or GKE Workload Identity / GCE metadata
+    /// (GCS) — via the cloud SDK's default credential chain, with no static
+    /// keys anywhere. Supported for S3 and GCS only.
+    CloudIdentity,
 }
 
 /// Top-level `ObservabilityExporter` resource. `deny_unknown_fields`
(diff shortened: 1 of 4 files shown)


moonming:
## Independent pre-merge audit — response

**VERDICT: CLEAR** (no HIGH/MEDIUM). The audit verified the crate-behavior claims against the **object_store 0.13.2 source** (`from_env` resolves the full AWS default chain incl. IMDS/IRSA/ECS as the final fallback; GCS `new()` with no key falls back to ADC / GCE metadata — both `build()` offline in CI), independently validated the schema's `if/then/else` over 10 cases (the `required:["auth_mode"]` guard is present, so back-compat holds and `credential_ref` is required outside cloud_identity), confirmed ambient creds are gated to **explicit opt-in at the serde type level** (default `credential_ref`), and that `dump-schema` regenerates **byte-identical** (no drift; the dumped file is consumed via `include_str!` and the change is additive). 28 object_store tests pass; clippy `-D warnings` clean.

3 LOW findings, all non-blocking (defense-in-depth or inherent) → tracked in **552**:
- DP fail-fast on `cloud_identity + endpoint` (S3) — cp-api gates it at create; DP hardening folded into the cp-api PR + the issue.
- `credential_ref` empty-string vs omitted under cloud_identity — normal path omits the field; add `skip_serializing_if` + tighten the doc; cp-api will omit it for cloud_identity.
- Real-cloud e2e verification before GA — inherent (no IAM role in CI; `build()` is lazy); manual real-cloud check tracked for GA.

CLEAR + LOWs tracked → ready to merge once CI is green.

Issue #546, opened 2026-06-08
Output guardrails on /v1/responses don't scan tool-call arguments (asymmetric with chat #448)


moonming:
## Summary
The output-guardrail PR (#544) scans the visible assistant text (`output[].content[].text`) on `/v1/responses`. It does **not** scan tool-call material in `output[]` — `function_call` / `custom_tool_call` items carry model-generated text in their `name` + `arguments` fields.

On `/v1/chat/completions`, `ChatResponse::guardrail_output_text` serializes `tool_calls` (name + arguments) into the scanned text — this was the #448 fix, closing exactly this hole. So a blocked literal emitted inside a Responses `function_call.arguments` bypasses the output guardrail on `/v1/responses` even though the chat equivalent is caught.

## Scope
- [ ] `responses_output_text` (and the streaming `responses_sse_output_text`) should also collect `function_call`/`custom_tool_call` `name` + `arguments` from `output[]`, so tool-call output is scanned like chat's #448.

## Severity
LOW — the visible-text echo concern #719 demonstrates is covered; this is the tool-call channel, asymmetric with chat. Surfaced by the independent audit of #544. Refs #719, #544, and the chat-side #448.

moonming:
Fixed by #553 (merged to main, commit bc369e1): the /v1/responses output guardrail now scans tool-call arguments (function_call name/arguments, mcp_call, custom_tool input), matching chat.rs #448. Closing as fixed.

Issue #448, opened 2026-05-29
security: enforce guardrails through a unified request policy pipeline


juzhiyuan:
## Summary

Several request modes and response fields bypass the normal guardrail pipeline. The fix should define one policy path for every prompt-bearing endpoint and every client-visible output field, including streaming, cache hits, reasoning output, tool-call arguments, routing targets, `/v1/messages`, and `/v1/messages/count_tokens`.

## Source

- Imported from Codex Security scan results shared in Slock `#Dev:bc985f48` (`chatgpt.com.har`).
- This issue groups findings with the same engineering boundary to avoid fragmented fixes.
- Triage priority: **P0**.

## Covered scanner findings

- [ ] **#2** `Non-streaming reasoning_content bypasses output guardrails` — `reasoning_content` is rendered to clients but output guardrails inspect only `message.content`.
- [ ] **#3** `Vertex Claude tool calls bypass output guardrails` — Vertex Claude `tool_calls` can carry model output that bypasses content/DLP checks.
- [ ] **#4** `Prompt Shield skips trailing text in oversized tokens` — Prompt Shield evaluates only the benign prefix of an oversized token while the full prompt reaches the model.
- [ ] **#5** `Routing bypasses model-scoped guardrails` — Guardrails are resolved for the virtual router model and not recomputed for the concrete target model.
- [ ] **#6** `count_tokens forwards prompts without input guardrails` — `/v1/messages/count_tokens` sends prompts/tools/images upstream without input checks.
- [ ] **#13** `Streaming reasoning_content bypasses output guardrails` — Streaming `delta.reasoning_content` is emitted but not included in the output-guardrail buffer.
- [ ] **#18** `Tool-call outputs bypass output guardrails` — OpenAI-compatible `tool_calls` bypass output guardrails in streaming and non-streaming paths.
- [ ] **#21** `Tool-call responses bypass output guardrails` — Anthropic `tool_use` arguments bypass output guardrails.
- [ ] **#22** `Anthropic /v1/messages bypasses quotas and guardrails` — Cross-provider `/v1/messages` dispatch bypasses the normal policy pipeline.
- [ ] **#28** `Streaming and cache hits bypass output guardrails` — Streaming responses and non-streaming cache hits can bypass output policy.

## Expected fix direction

- [ ] Route every prompt-bearing endpoint through input guardrails before upstream dispatch, including utility endpoints such as count_tokens.
[...]

Issue #153, opened 2026-05-09
bug: output guardrail error message echoes the matched forbidden literal back to caller


moonming:
## Summary

When an output guardrail with `kind: "keyword"` and `hook_point: "output"` blocks an upstream response containing a forbidden literal, the gateway's caller-visible error message includes the matched literal verbatim — defeating the purpose of an output guardrail (which exists precisely to keep forbidden content from reaching the caller).

## Repro

E2E test attached: drafted as `tests/e2e/src/cases/guardrail-output-e2e.test.ts` (held back from #151 batch 2 PR pending this fix).

Setup:
- Mock upstream returns assistant message containing the literal `"leakedsecret"`
- Output keyword guardrail configured with `patterns: [{kind:"literal", value:"leakedsecret"}]`
- Caller sends an innocent prompt

Observed:
\`\`\`json
{
  "error": {
    "message": "content blocked by policy: output blocked by literal \"leakedsecret\"",
    "type": "content_filter"
  }
}
\`\`\`

The forbidden literal is right there in `error.message`. Anyone who can trigger the guardrail can extract the policy's blocklist by inspecting error responses.

Expected:
\`\`\`json
{
  "error": {
    "message": "response blocked by content policy",
    "type": "content_filter"
  }
}
\`\`\`

Or any redacted equivalent that does NOT include the matched pattern.

## Severity

**MEDIUM-HIGH** for output guardrails specifically:
- For input guardrails, leaking the matched pattern back is mostly cosmetic — the caller sent it, they already know.
- For output guardrails, leaking the matched pattern is a real information leak — the whole point is to prevent the model's forbidden output from reaching the caller. Echoing it in the error envelope is a partial bypass.

## Proposed fix

Strip / redact the matched value from the error rendered to clients. Internal logs may keep the matched pattern for operator debugging, but the wire-level `error.message` and any other caller-visible field must not carry it.

## Test attached

The held-back test asserts:
\`\`\`ts
const errorBlob = JSON.stringify(caught.error ?? {});
expect(errorBlob).not.toContain(FORBIDDEN_WORD);
\`\`\`

Will be added back to the e2e suite once this is fixed.

PR #553, opened 2026-06-08
fix(guardrails): scan tool-call args on /v1/responses output (#546)

moonming:
## What
The `/v1/responses` output guardrail (added in #544) scanned only `output_text`. A `function_call`/`custom_tool_call` item carries model output under **top-level** `name`/`arguments`/`input` — **not** under `content[]` — so a blocked literal in tool-call arguments **bypassed** the output guardrail. The chat surface already scans tool-call output (`ChatResponse::guardrail_output_text`, the #448 fix); this restores parity.

## How
- **Non-streaming + the `response.completed` path**: `responses_output_text` now also collects each `output[]` item's `name` + `arguments` (function calls) + `input` (custom tool calls), in addition to message `content[].text`.
- **Streaming terminals**: `responses_sse_output_text` now matches `response.completed` **and** `response.incomplete` / `response.failed` — all three carry the full `output[]` (incl. tool-call items), and incomplete/failed fire routinely (e.g. `max_output_tokens` truncation).
- **Truncated streams (no terminal object)**: also accumulate `response.function_call_arguments.delta` (tool-call args stream via their own event, not `output_text.delta`), concatenated without a separator so a literal split across deltas still reassembles.

Reasoning items stay excluded (they carry `summary`, not `content`/`arguments`). Corrected the doc-comment that falsely claimed tool-call args were out of scope.

## Independent audit (CLAUDE.md §8) — findings + resolution
The first audit found a **HIGH**: the streaming fix initially only matched `response.completed`, so tool-call args still leaked (200) on `response.incomplete`/`response.failed` terminals and on truncated streams — the exact bypass class this PR exists to close, triggered by an ordinary `max_output_tokens` completion. **Fixed in code** (commit `0909f5d`): match incomplete/failed terminals + accumulate `function_call_arguments.delta`. Two gating tests added (they fail before the fix, pass after). Verified clean: non-streaming bypass closed, `arguments` scanned as a raw string (substring match hits the literal), reasoning excluded, no over-scan of message items, #153 redaction intact, buffer-cap still fails closed.

## Test plan
`function_call` with a blocked literal in `arguments` → 422 `content_filter`: non-streaming; streaming via `response.completed`; streaming via `response.incomplete` terminal; and a delta-only stream with the literal split across two `function_call_arguments.delta` events (no terminal). All hold the args back (never in the response body). fmt + clippy clean; output-guardrail suite green.

## Related
The **chat.rs streaming** `tool_calls` leak (same #448 class on the chat surface — streaming buffers only `delta.content`, not `delta.tool_calls`; pre-existing) is fixed in a separate PR. Refs #546, #719, #448.

### DIFF
--- a/crates/aisix-proxy/src/responses.rs
+++ b/crates/aisix-proxy/src/responses.rs
@@ -808,26 +808,46 @@ fn extract_response_usage(body: &Value) -> Option<ResponseUsage> {
     })
 }
 
-/// Collect the assistant's visible output text from a Responses-API
-/// response object for output-guardrail scanning (#719): the `text` of
-/// every `output_text` content part across all message items in
-/// `output[]`. Reasoning items and tool-call arguments are not included
-/// (reasoning is out of output-guardrail scope, matching the chat surface).
+/// Collect the assistant's output text from a Responses-API response object
+/// for output-guardrail scanning (#719/#546):
+/// - the `text` of every `output_text` content part of message items, and
+/// - the `name` + `arguments` (function calls) / `input` (custom tool calls)
+///   of tool-call items — these are **top-level** item fields, not under
+///   `content[]`, so without scanning them a blocked literal placed in a
+///   tool-call's arguments would bypass the output guardrail. The chat
+///   surface scans tool-call output too (`ChatResponse::guardrail_output_text`,
+///   the #448 fix); this keeps the surfaces symmetric.
+///
+/// Reasoning items are intentionally excluded (out of output-guardrail
+/// scope, matching the chat surface) — they carry `summary`, not `content`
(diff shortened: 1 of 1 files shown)


moonming:
Re-audit of the HIGH fix returned **MERGE** (verified fail-before/pass-after on the incomplete-terminal + delta-only tests, no regressions, clippy/fmt clean). It flagged one LOW-MEDIUM residual — MCP/custom-tool arg deltas unscanned only on truncated streams with no terminal object — which I've now **closed in code** (commit folds `response.mcp_call_arguments.delta` / `response.custom_tool_call_input.delta` into the same delta arm; MCP test added). No known residual remains on the streaming tool-call surface.

Issue #552, opened 2026-06-08
object_store cloud_identity: DP fail-fast on endpoint, credential_ref serde, + real-cloud e2e (follow-ups from #549)


moonming:
## Context
#549 added the keyless `cloud_identity` auth mode (S3 / GCS) to the DP object_store sink. Its independent audit returned CLEAR and surfaced 3 LOW follow-ups (none blocking #549):

## 1. DP fail-fast on `cloud_identity` + custom `endpoint` (S3)
`build_object_store_ambient` honors a custom `endpoint` for cloud_identity S3, but ambient AWS credentials require the metadata service (169.254.169.254), which a non-AWS S3-compatible host (MinIO / R2 / OSS) doesn't provide — so `build()` succeeds (lazy) but every delivery fails soft at runtime. cp-api will reject `cloud_identity + endpoint` at create time (done in the cp-api PR), so this is belt-and-suspenders; the DP should still fail fast with a clear permanent error:
```rust
ObjectStoreProvider::S3 => {
    if endpoint.is_some() {
        return Err(SinkError::Permanent(
            "object_store: cloud_identity for s3 does not support a custom endpoint \
             (ambient AWS credentials require the metadata service); use credential_ref".into(),
        ));
    }
    ...
}
```

## 2. `credential_ref` empty-string vs omitted under cloud_identity
The `ObjectStoreConfig::credential_ref` doc says it "may be empty when auth_mode = cloud_identity", but the loader schema's `minLength: 1` rejects an explicit `credential_ref: ""`. The normal/tested path omits the field entirely (fine). Make the two consistent: add `#[serde(default, skip_serializing_if = "String::is_empty")]` so an empty ref is never serialized onto the wire, and tighten the doc to "omitted (not empty)". Also ensure cp-api **omits** `credential_ref` for cloud_identity (cp-api PR).

## 3. Real-cloud e2e verification before GA
cloud_identity's actual authentication has **no** CI coverage — there is no real IAM role in CI, and `object_store::build()` is lazy, so the passing unit tests prove only that the builder *wired* (S3 via `from_env`, GCS via no-key ADC), not that auth *works*. This is an inherent cloud-platform constraint (same as the existing GCS smoke). Before advertising cloud_identity GA, do a manual real-cloud check that objects actually land: S3 via EKS IRSA / EC2 instance role; GCS via GKE Workload Identity.

## Refs
- `api7/ai-gateway#549` (DP cloud_identity) — surfaced by its audit.
[...]

PR #568, opened 2026-06-09
fix(object_store): harden cloud_identity — fail-fast on S3 endpoint, omit empty credential_ref

moonming:
## What

Two hardening follow-ups surfaced by #549's audit and tracked in #552 (items 1 & 2). Both are small, fully unit-tested DP changes.

### 1. DP fail-fast on `cloud_identity` + custom `endpoint` (S3)

`build_object_store_ambient` honored a custom `endpoint` for cloud_identity S3. But ambient AWS credentials are sourced from the instance metadata service (IMDS / IRSA / ECS task role), which only the provider's native endpoint exposes — a custom endpoint means an S3-compatible host (MinIO / R2 / OSS) with no cloud IAM identity. So `build()` succeeded (lazy) and **every delivery failed soft** at runtime.

Now the S3 arm returns a clear `SinkError::Permanent` when an endpoint is set, pointing the operator back to `credential_ref`. cp-api already rejects this combination at create time (AISIX-Cloud#736); this is the DP-side backstop.

### 2. `credential_ref` omitted, not empty, under cloud_identity

`ObjectStoreConfig::credential_ref` now carries `#[serde(default, skip_serializing_if = "String::is_empty")]`, so an empty reference never serializes onto the wire. This keeps the keyless encoding consistent with the omitted-field path the loader accepts (the loader's `minLength: 1` would reject an explicit `credential_ref: ""`, but accepts the field omitted). The doc comment is tightened to "omitted (not empty)", and the generated schema is regenerated (it drops the now-moot `default: ""`).

## Tests

- `build_object_store_ambient_s3_rejects_endpoint` — the S3 ambient builder returns a `Permanent` error (mentioning `endpoint` + `credential_ref`) when a custom endpoint is supplied.
- `object_store_auth_mode_defaults_and_cloud_identity` — extended to assert a cloud_identity config serializes **without** `credential_ref` (`v.get("credential_ref").is_none()`).

Local verification: `cargo fmt`, `cargo clippy -p aisix-obs -p aisix-core --all-targets` (clean), `cargo test -p aisix-obs -p aisix-core` (111 pass + the 2 new), and `cargo run -p aisix-core --bin dump-schema` + committed the regenerated schema (no remaining drift).

## Scope

Pure DP hardening + tests; no wire-protocol break (an empty `credential_ref` was never a valid keyless input — it was already rejected by the loader's `minLength: 1`; this only stops the DP from *emitting* one). Real-cloud e2e for `cloud_identity` (#552 item 3) lands separately.

Refs: #549 (DP cloud_identity), #552 (this issue), AISIX-Cloud#736 (cp-api reject), AISIX-Cloud#724.

## Summary by CodeRabbit

* **Bug Fixes**
  * S3 cloud identity authentication now rejects custom endpoints and provides a clear error message directing users to use credential references for S3-compatible hosts.

* **Documentation**
  * Updated observability exporter schema documentation to clarify credential field behavior based on authentication mode configuration.

### DIFF
--- a/crates/aisix-core/src/models/observability_exporter.rs
+++ b/crates/aisix-core/src/models/observability_exporter.rs
@@ -284,9 +284,10 @@ pub struct ObjectStoreConfig {
     /// Opaque pointer to the cloud credentials, resolved locally by the DP
     /// at delivery time. The plaintext key MUST NOT live in etcd/kine — the
     /// control plane stores only this reference, never the secret itself.
-    /// Required when `auth_mode = credential_ref`; unused (and may be empty)
-    /// when `auth_mode = cloud_identity`.
-    #[serde(default)]
+    /// Required when `auth_mode = credential_ref`; omitted (not empty) when
+    /// `auth_mode = cloud_identity` — `skip_serializing_if` keeps an empty
+    /// reference off the wire so the two encodings never diverge.
+    #[serde(default, skip_serializing_if = "String::is_empty")]
     pub credential_ref: String,
 }
 
@@ -629,6 +630,14 @@ mod tests {
             }
             other => panic!("expected object_store, got {other:?}"),
         }
+        // An empty credential_ref is omitted on the wire (skip_serializing_if),
+        // so a keyless config never carries an empty key reference.
+        let v = serde_json::to_value(&e).unwrap();
+        assert_eq!(v["auth_mode"], "cloud_identity");
+        assert!(
+            v.get("credential_ref").is_none(),
+            "empty credential_ref must be omitted under cloud_identity, got {v}"
+        );
     }
 
     #[test]
--- a/crates/aisix-obs/src/sink/object_store.rs
+++ b/crates/aisix-obs/src/sink/object_store.rs
@@ -280,7 +280,9 @@ pub fn build_object_store(
 /// Build the backend using the DP host's OWN attached cloud identity, with no
 /// static keys ("cloud identity" auth). **S3** uses the AWS default credential
 /// chain via [`object_store::aws::AmazonS3Builder::from_env`] (env / EKS IRSA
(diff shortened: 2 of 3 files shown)


moonming:
## Independent audit (repo §8) — **CLEAR**

A fresh agent reviewed this PR cold at head SHA `1095883`, verifying every claim against source (not the description) — including a manual diff of the regenerated schema against `origin/main` and the green `schema drift` CI job.

**No HIGH / MEDIUM / LOW findings.**

- **Correctness** — the S3 arm returns `Permanent` (naming `endpoint` + `credential_ref`) *before* building; happy path (`endpoint=None`) still builds; GCS/Azure arms byte-unchanged. `skip_serializing_if` affects only serialization-when-empty; deserialization still defaults `""`; non-empty refs still serialize.
- **Schema** — the only change is `credential_ref` losing `default: ""` (+ description); zero other drift (confirmed by manual diff + the passing `schema drift` job); `grep '"default": ""'` is empty in the new schema.
- **No breaking change** — the only Rust consumer reads the in-memory `String` (always present, independent of serialization); the hand-written loader schema already accepts an *omitted* `credential_ref` under cloud_identity (base `required` = provider/bucket/prefix; the cloud_identity branch doesn't require it) and rejects an explicit `""` via the shared `minLength: 1` — so emitting the field omitted is the correct encoding. Needs no regeneration.
- **Tests pin the contract** — reverting the fail-fast fails `build_object_store_ambient_s3_rejects_endpoint`; removing `skip_serializing_if` fails `object_store_auth_mode_defaults_and_cloud_identity` (`credential_ref` re-emitted); the pre-existing `object_store_round_trips_flat` catches a regression on the credential_ref-MODE side (non-empty ref must still serialize).
- **No leakage** — the error string names only field/provider names, nothing sensitive.

**Merge gate:** no HIGH/MEDIUM outstanding.

PR #569, opened 2026-06-09
test(object_store): real-cloud credential_ref smoke CI for S3 + GCS

moonming:
## What

A **gated CI workflow** that runs the existing `objstore_smoke_*` round-trips against **real Amazon S3 and Google Cloud Storage**, plus the small test changes needed to target native cloud (not just S3-compatible emulators). This is the credential_ref (static-key) half of #552 item 3 — real-cloud verification that the object_store write path works against the actual cloud APIs, which the emulator smoke in `ci.yml` cannot do (no real S3 auth; fake-gcs-server can't round-trip the real GCS PUT path).

## Changes

**`crates/aisix-obs/src/sink/object_store.rs`** (test-only, inside `mod smoke`):
- `objstore_smoke_s3_roundtrip` — `endpoint` is now **optional**: omit it for native AWS S3 (virtual-hosted + region); set it for an S3-compatible host (MinIO / R2, path-style). Region comes from `AISIX_E2E_OBJSTORE_S3_REGION` (default `us-east-1`). Requires only bucket + key + secret.
- `objstore_smoke_gcs_roundtrip` — drop the **unused** `_ENDPOINT` requirement. `build_object_store` intentionally ignores `endpoint` for GCS (an emulator base URL rides the service-account JSON's `gcs_base_url`), so native GCS needs only bucket + service-account key.

**`.github/workflows/objstore-real-cloud.yml`** (new):
- Gated — never runs on a normal push. Triggers: `workflow_dispatch`, nightly `schedule`, and a same-repo PR labeled `objstore real cloud`.
- Runs `cargo test -p aisix-obs -- --ignored objstore_smoke_s3 objstore_smoke_gcs` with credentials from repository secrets.
- Each provider's test **self-skips** (the test's own `eprintln`, plus a `::warning` from a preflight step) when its secrets are absent, so an unconfigured run is a green no-op — it never breaks CI before the secrets exist.

**`crates/aisix-obs/tests/real-cloud.env.example`** (new): a local-run template documenting the env vars + the least-privilege IAM policy (`s3:PutObject/GetObject/DeleteObject` + `s3:ListBucket`) and GCS role (`roles/storage.objectAdmin`). The filled-in `real-cloud.env` is gitignored.

## Required repository secrets (operator-configured)

| Secret | Provider |
|---|---|
| `AISIX_E2E_OBJSTORE_S3_BUCKET`, `_REGION`, `_ACCESS_KEY_ID`, `_SECRET_ACCESS_KEY` (+ optional `_ENDPOINT` for S3-compatible) | S3 |
| `AISIX_E2E_OBJSTORE_GCS_BUCKET`, `_SERVICE_ACCOUNT` (full SA key JSON) | GCS |

Scope each identity to the test bucket only (policy in the workflow header + `real-cloud.env.example`).

## What the smoke asserts (observable contract)

`smoke_roundtrip`: write a batch → exactly one object under the prefix → re-deliver the same batch → still exactly one (content-addressed key is idempotent) → clean up. Real auth + real PUT/LIST/DELETE against the cloud.

## Verification

`cargo fmt`, `cargo clippy -p aisix-obs --all-targets` (clean), `cargo test -p aisix-obs -- --ignored objstore_smoke_s3 objstore_smoke_gcs` with no secrets set → both self-skip and pass (compile + skip path verified). Workflow YAML parses. The real round-trips run once the secrets are configured.

## Notes

- No emulator regression: the existing `object-store-emulators.env` path still works — S3 still honors `_ENDPOINT` when set; GCS never used `_ENDPOINT` (the SA JSON's `gcs_base_url` drives the emulator).
- Scope: credential_ref only. The keyless `cloud_identity` real-cloud path (OIDC / Workload Identity Federation) lands in a follow-up PR — it needs a runner with an attached cloud identity, not static keys.

Refs: #552 (item 3), #549, #568 (DP hardening).

## Summary by CodeRabbit

* **Tests**
  * Added GitHub Actions workflow for real cloud object-store smoke tests against S3 and GCS.
  * Updated S3 and GCS smoke test configurations to support optional endpoint configuration and service account authentication.

* **Chores**
  * Added environment template file and gitignore configuration for cloud storage credentials management.

### DIFF
--- a/.github/workflows/objstore-real-cloud.yml
+++ b/.github/workflows/objstore-real-cloud.yml
@@ -1,0 +1,79 @@
+# Real-cloud object_store smoke e2e — credential_ref (static-key) write path
+# against REAL Amazon S3 and Google Cloud Storage.
+#
+# Runs the `objstore_smoke_s3` / `objstore_smoke_gcs` round-trips (write a batch
+# → assert exactly one object under the prefix → re-deliver the same batch →
+# assert no duplicate key → clean up) against the actual cloud APIs. These
+# complement the emulator smoke in ci.yml's unit job, which cannot exercise real
+# S3 auth or (for GCS) the real PUT path that fake-gcs-server can't round-trip.
+#
+# Gated: never runs on a normal push. Triggered manually (workflow_dispatch),
+# nightly (schedule), or on a same-repo PR labeled `objstore real cloud`. Each
+# provider's test self-skips (logged) when its secrets are absent, so an
+# unconfigured run is a green no-op.
+#
+# ── Required repository secrets ──
+#   Settings → Secrets and variables → Actions → New repository secret
+#
+#   Amazon S3 (native AWS):
+#     AISIX_E2E_OBJSTORE_S3_BUCKET             test bucket name
+#     AISIX_E2E_OBJSTORE_S3_REGION             the bucket's region, e.g. us-east-1
+#     AISIX_E2E_OBJSTORE_S3_ACCESS_KEY_ID      IAM user access key id
+#     AISIX_E2E_OBJSTORE_S3_SECRET_ACCESS_KEY  IAM user secret access key
+#     (For an S3-compatible host — MinIO / Cloudflare R2 — also set
+#      AISIX_E2E_OBJSTORE_S3_ENDPOINT.)
+#   Google Cloud Storage:
+#     AISIX_E2E_OBJSTORE_GCS_BUCKET            test bucket name
+#     AISIX_E2E_OBJSTORE_GCS_SERVICE_ACCOUNT   the full service-account key JSON
+#                                              (paste the downloaded key as one secret)
+#
+# Least privilege — scope the identity to the test bucket only:
+#   AWS:  s3:PutObject, s3:GetObject, s3:DeleteObject on arn:aws:s3:::<bucket>/*
+#         + s3:ListBucket on arn:aws:s3:::<bucket>
+#   GCS:  roles/storage.objectAdmin on the bucket
+#
+# See crates/aisix-obs/tests/real-cloud.env.example for a local-run template.
+name: objstore-real-cloud
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "30 0 * * *" # 00:30 UTC (08:30 Asia/Shanghai)
+  pull_request:
+    types: [labeled, synchronize, reopened]
+    branches: [main]
+
+permissions:
+  contents: read
+
+concurrency:
+  group: objstore-real-cloud-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  smoke:
+    name: objstore real-cloud smoke (S3 + GCS)
+    # Label-gate on PRs; always run on manual dispatch / schedule.
+    if: >-
+      github.event_name != 'pull_request' ||
+      contains(github.event.pull_request.labels.*.name, 'objstore real cloud')
+    runs-on: ubicloud-standard-2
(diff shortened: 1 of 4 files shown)


moonming:
## Independent audit (repo §8) — MEDIUM + 2 LOW, **all resolved**

A fresh agent reviewed this PR cold at the head SHA, independently verifying the two load-bearing claims (GCS ignores `endpoint`; no emulator regression) and the security posture against source. No HIGH; no committed secret; no real-cloud calls on a normal push; no leakage; no production-code change. Findings + resolution:

**MEDIUM — missing explicit same-repo / fork guard on the PR trigger** → **fixed** (`5c5edf8`).
The `pull_request` trigger injected the six cloud-credential secrets as job env gated only by the label. Safe today only because plain `pull_request` withholds fork secrets by GitHub default — but a future settings flip (or a switch to `pull_request_target`) would expose all six credentials to fork-controlled `cargo test` code. Now the PR gate also requires `github.event.pull_request.head.repo.full_name == github.repository`, matching the description's "same-repo" claim and the repo's own `docker-image.yml` convention. Defense-in-depth, independent of the GitHub default.

**LOW — silently-green unconfigured run** → **fixed** (`5c5edf8`).
A deliberate `workflow_dispatch`/`schedule` run now **fails** (not skip-green) when neither provider's secrets are configured — a green check that tested nothing is a silenced coverage failure. PR runs may legitimately have no secrets (unlabeled / fork), so they still self-skip.

**LOW — cross-run interference** → no change (justified).
Keys are content-addressed under a per-run UUID prefix and each run cleans up its own prefix; `list_under` assertions are prefix-scoped. Concurrent runs cannot collide. Noted for completeness.

**Verified correct (no change):** GCS arm of `build_object_store` never references `endpoint` (so dropping `_ENDPOINT` is a true no-op for real GCS + the emulator, which rides the SA JSON's `gcs_base_url`); S3 both arms correct (native virtual-hosted vs S3-compatible path-style); `mod smoke` tests are `#[ignore]` so the normal unit job stays unaffected; `real-cloud.env` is gitignored while `.example` is tracked; no credential in any log line; setup steps match `ci.yml`.

**Merge gate:** the MEDIUM is fixed in code; both LOWs addressed/justified. No blocking findings outstanding.

moonming:
## Real-cloud verification — S3 + GCS both green ✅

The gated `objstore-real-cloud` workflow has been run end-to-end against **real Amazon S3 and Google Cloud Storage** (repository secrets configured), in addition to a local round-trip per provider.

**CI run (workflow `objstore-real-cloud.yml`, both providers configured):**
```
S3: configured
GCS: configured
sink::object_store::smoke::objstore_smoke_s3_roundtrip ... ok      # real AWS S3
sink::object_store::smoke::objstore_smoke_gcs_roundtrip ... ok     # real GCS
test result: ok. 2 passed; 0 failed
```

**Local round-trips** (same tests, real buckets): S3 `1 passed / 8.74s`, GCS `1 passed / 8.1s`.

Each `smoke_roundtrip` is the full observable contract against the live API: write a batch → exactly one object under the prefix → re-deliver the same batch → still exactly one (content-addressed idempotent key) → clean up — real auth + PUT/LIST/DELETE.

Also confirmed the gating works as designed: an earlier run with only GCS configured logged `::warning::S3 secrets absent` and self-skipped S3 while GCS ran real — no false-green for an unconfigured provider.

This closes the credential_ref half of #552 item 3 (real-cloud verification). The keyless `cloud_identity` path lands separately (it needs a runner with an attached cloud identity / OIDC, not static keys).

PR #572, opened 2026-06-09
test(object_store): keyless cloud_identity real-cloud smoke CI (OIDC)

moonming:
## What

The **keyless `cloud_identity`** half of #552 item 3: prove the data plane's ambient-credential builders write to **real S3 and GCS with no static keys**, the way a cloud-deployed DP authenticates via an EC2 instance role / EKS IRSA / GKE Workload Identity. The static-key `credential_ref` half landed in #569.

## Changes

**`crates/aisix-obs/src/sink/object_store.rs`** (test-only, `mod smoke`):
- `objstore_smoke_s3_cloud_identity` — keyless round-trip via `build_object_store_ambient(S3, …)`; AWS `from_env` sources the ambient chain (no static keys). Gated on `AISIX_E2E_OBJSTORE_CLOUDID_S3_BUCKET` (+ optional `_REGION`).
- `objstore_smoke_gcs_cloud_identity` — keyless round-trip via `build_object_store_ambient(Gcs, …)`; Application Default Credentials with no service-account key. Gated on `AISIX_E2E_OBJSTORE_CLOUDID_GCS_BUCKET`.
- Both reuse `smoke_roundtrip` (write → exactly one object → re-deliver → still one → clean up).

**`.github/workflows/objstore-cloud-identity.yml`** (new):
- `permissions: id-token: write` — the job mints a GitHub OIDC token to **assume an AWS IAM role** (`aws-actions/configure-aws-credentials`) and to **impersonate a GCP service account via Workload Identity Federation** (`google-github-actions/auth`). No static cloud keys touch CI.
- Gated: `workflow_dispatch` / nightly / same-repo PR labeled `objstore real cloud`. Each provider's auth step + smoke self-skips when its OIDC identifier secret is absent; a deliberate run with neither configured fails (not green) so it can't pass while testing nothing.

## OIDC setup (provisioned)

| Cloud | Identity | Secret (identifier, not a key) |
|---|---|---|
| AWS | IAM role `aisix-objstore-smoke-oidc`, trust `repo:api7/ai-gateway:*`, least-priv `s3:PutObject/GetObject/DeleteObject` + `ListBucket` on the test bucket | `AISIX_E2E_OIDC_AWS_ROLE_ARN` |
| GCP | Workload Identity Pool `github-actions` + provider `github` (restricted to `api7/ai-gateway`), impersonating the existing `objectAdmin` SA | `AISIX_E2E_OIDC_GCP_WIF_PROVIDER`, `AISIX_E2E_OIDC_GCP_SA` |

Buckets are reused from #569's secrets. **No static cloud key is stored anywhere** — only OIDC role/provider identifiers.

## Verification

`cargo fmt`, `cargo clippy -p aisix-obs --all-targets` (clean), and the two smokes self-skip + pass with no `CLOUDID` env set. The real keyless round-trips run in CI once this PR is labeled `objstore real cloud` (evidence will be posted on the PR).

Refs: #552 (item 3), #549 (DP cloud_identity), #569 (credential_ref real-cloud), #568 (DP hardening).

## Summary by CodeRabbit

* **Tests**
  * Added ignored, environment-gated end-to-end smoke tests for cloud identity (keyless) authentication against AWS S3 and Google Cloud Storage; tests self-skip when required environment variables are unset.

* **Chores**
  * Added a CI workflow to run the cloud-identity E2E smoke tests (manual, nightly, and PR triggers). Uses GitHub OIDC for keyless AWS access and gate conditions to avoid running when configuration is missing.

### DIFF
--- a/.github/workflows/objstore-cloud-identity.yml
+++ b/.github/workflows/objstore-cloud-identity.yml
@@ -1,24 +1,27 @@
 # Real-cloud object_store smoke e2e — KEYLESS cloud_identity path.
 #
-# Proves the data plane's ambient-credential builders (`build_object_store_ambient`)
-# write to real S3 / GCS with NO static keys: the runner assumes an identity via
-# GitHub OIDC (an AWS IAM role / a GCP Workload Identity Federation provider), and
-# the object_store ambient chain (AWS `from_env` / GCP Application Default
-# Credentials) picks it up — exactly what a cloud-deployed DP does via an EC2
-# instance role / EKS IRSA / GKE Workload Identity. The static-key credential_ref
-# path is covered separately by `objstore-real-cloud.yml`.
+# Proves `build_object_store_ambient` writes to real S3 with NO static keys: the
+# runner mints a GitHub OIDC token to assume an AWS IAM role
+# (aws-actions/configure-aws-credentials), and the object_store ambient chain
+# (AWS `from_env`) picks it up — exactly what a cloud-deployed DP does via an EC2
+# instance role / EKS IRSA. The static-key credential_ref path is covered by
+# objstore-real-cloud.yml.
 #
-# Gated: never on a normal push. Triggered manually, nightly, or on a same-repo
-# PR labeled `objstore real cloud`. Each provider self-skips when its OIDC is
-# unconfigured, so an unconfigured run is a green no-op.
+# Scope note — GCS keyless is NOT exercised here. On a non-GCE GitHub runner,
+# object_store's GCS Application Default Credentials resolution targets the GCE
+# metadata server (169.254.169.254), which the runner doesn't have; it does not
+# consume the Workload Identity Federation credentials file. GCS cloud_identity
+# works on a real GKE/GCE deployment (the metadata path that
+# build_object_store_ambient(Gcs, …) uses in production); its real-runner
+# round-trip is tracked in #573. objstore_smoke_gcs_cloud_identity self-skips here.
 #
-# Required repository secrets — these are IDENTIFIERS, not static cloud keys:
-#   AISIX_E2E_OIDC_AWS_ROLE_ARN     IAM role to assume (trusts repo:api7/ai-gateway via GitHub OIDC)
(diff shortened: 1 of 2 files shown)


moonming:
## Real-cloud verification + audit follow-ups

### Result of the first keyless run
- **AWS S3 keyless: verified end-to-end in CI** ✅ — `objstore_smoke_s3_cloud_identity ... ok`. The runner minted a GitHub OIDC token, assumed the IAM role, and `AmazonS3Builder::from_env()` used the role's ambient credentials for a real keyless round-trip. No static keys.
- **GCS keyless: not exercisable on a non-GCE GitHub runner** — `object_store`'s GCS ADC went to the GCE metadata service (`http://169.254.169.254/computeMetadata/...`) instead of the Workload Identity Federation credentials file, and a GitHub runner has no metadata server. This is an environment/tooling limitation, **not a product bug**: GCS `cloud_identity` works on a real GKE/GCE deployment (the metadata path the ambient builder uses in production). The GCS smoke now self-skips here and stays in the cargo filter for a real GKE/GCE run; verification tracked in **#573**. The GCS ambient *builder* is still unit-tested, and GCS credential_ref is green in `objstore-real-cloud.yml` (#569).

### Audit (repo §8) — MEDIUM + LOW
- **MEDIUM (fixed, `a0c7a4f`)** — the AWS auth step required `aws-region` but was gated only on the role ARN, so a role-set/region-empty config would hard-fail the whole job. Now gated on **role ARN AND region**, and a deliberate dispatch/schedule run fails fast (not silently green) when either is missing.
- **LOW (justified)** — the same-repo blast-radius depends on the IAM role trust + GCP WIF provider staying pinned to `repo:api7/ai-gateway` (they are: role trust `repo:api7/ai-gateway:*`, WIF provider attribute-condition `assertion.repository == 'api7/ai-gateway'`). The workflow itself does nothing to weaken this; noted so the trust is never broadened org-wide. No code change.

Security posture confirmed by the audit: a fork PR cannot mint the OIDC token (same-repo + label gate, plain `pull_request` withholds fork secrets/id-token), the builders are genuinely keyless (no `creds` param, no static keys in CI), and no secret value is echoed.

PR #577, opened 2026-06-09
test(object_store): real-cloud Azure Blob credential_ref smoke

moonming:
## What

Adds **real Azure Blob** to the credential_ref real-cloud smoke. The gated `objstore-real-cloud` workflow already ran S3 + GCS (#569); Azure was out of that scope. Now all three providers run against the real cloud APIs.

## Changes

- **`crates/aisix-obs/src/sink/object_store.rs`** (`mod smoke`): `objstore_smoke_azure_roundtrip` — `endpoint` is now **optional**. Omit it for a real Azure account (the `MicrosoftAzureBuilder` derives `https://<account>.blob.core.windows.net` from the account name); set it for the Azurite emulator. Requires only `account` + `access_key` + `container`. (Mirrors the native-cloud tweak #569 made for S3.)
- **`.github/workflows/objstore-real-cloud.yml`**: the gated job now runs `objstore_smoke_azure` too. Adds the three `AISIX_E2E_OBJSTORE_AZURE_{ACCOUNT,ACCESS_KEY,CONTAINER}` secrets, an Azure preflight `::warning`, and counts Azure in the "deliberate run with nothing configured → fail" guard. Azure self-skips when its secrets are absent.
- **`crates/aisix-obs/tests/real-cloud.env.example`**: Azure block added.

## Verification

Provisioned a dedicated real Azure storage account + container and ran the smoke locally with no `endpoint` (derived from the account):

```
test sink::object_store::smoke::objstore_smoke_azure_roundtrip ... ok   (1 passed, 3.66s)
```

That is the full observable contract against live Azure — real shared-key auth + PUT/LIST/DELETE (write a batch → exactly one object → re-deliver → still one → clean up). `cargo fmt`, `cargo clippy -p aisix-obs --all-targets` clean; the smoke self-skips + passes with no Azure env set. The real round-trip runs in CI once this PR is labeled `objstore real cloud` (evidence to be posted).

## Notes

- **Least privilege**: an Azure account key is account-wide, so a **dedicated test storage account** is used (the product's azure backend authenticates with account + key — `ObjectStoreCredentials::Azure`).
- **Azure `cloud_identity` is intentionally not covered** — the product rejects keyless Azure (`build_object_store_ambient` returns a permanent error for `azure_blob`; managed identity needs a non-secret account name the keyless config doesn't carry). So Azure has only the credential_ref path.

Refs: #569 (S3+GCS real-cloud), #552, AISIX-Cloud#724.

## Summary by CodeRabbit

* **Tests**
  * Expanded cloud storage smoke tests to include Azure Blob Storage support.
  * Added capability to test with both real Azure accounts and Azurite emulator.

* **Chores**
  * Updated CI/CD workflows to validate Azure Blob Storage configuration alongside existing S3 and GCS providers.

### DIFF
--- a/.github/workflows/objstore-real-cloud.yml
+++ b/.github/workflows/objstore-real-cloud.yml
@@ -1,7 +1,8 @@
 # Real-cloud object_store smoke e2e — credential_ref (static-key) write path
-# against REAL Amazon S3 and Google Cloud Storage.
+# against REAL Amazon S3, Google Cloud Storage, and Azure Blob.
 #
-# Runs the `objstore_smoke_s3` / `objstore_smoke_gcs` round-trips (write a batch
+# Runs the `objstore_smoke_s3` / `objstore_smoke_gcs` / `objstore_smoke_azure`
+# round-trips (write a batch
 # → assert exactly one object under the prefix → re-deliver the same batch →
 # assert no duplicate key → clean up) against the actual cloud APIs. These
 # complement the emulator smoke in ci.yml's unit job, which cannot exercise real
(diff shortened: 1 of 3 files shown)


moonming:
## Real-cloud verification — S3 + GCS + Azure all green ✅

The gated `objstore-real-cloud` workflow ran with all three providers configured:
```
S3: configured / GCS: configured / Azure: configured
objstore_smoke_azure_roundtrip ... ok    # real Azure Blob
objstore_smoke_s3_roundtrip ... ok       # real AWS S3
objstore_smoke_gcs_roundtrip ... ok      # real GCS
→ job: success
```
Plus a local run against a real Azure storage account: `objstore_smoke_azure ... ok (3.66s)`. Each is the full observable contract — real shared-key auth + PUT/LIST/DELETE (write → exactly one object → re-deliver → still one → clean up).

## Independent audit (repo §8) — CLEAR

A fresh agent verified at head `0640738` against `object_store` 0.13.2 source: **no HIGH/MEDIUM/LOW.**
- **Endpoint-optional Azure builder**: `None` → `object_store` derives `https://<account>.blob.core.windows.net`; `Some` → `with_endpoint` (+ `allow_http` for Azurite's `http://`). Correct for both real-account and emulator.
- **No emulator regression**: the Azurite path (`object-store-emulators.env`, endpoint set) is preserved; the well-known Azurite key appears only in the pre-existing emulators env, never in `real-cloud.env.example`.
- **Workflow**: no `push` trigger (no real-cloud calls on push); same-repo + label gate intact; Azure added to the preflight + the "deliberate run with nothing configured → fail" guard + the cargo filter.
- **Security**: example is placeholders only; secrets referenced by name, never echoed; the least-privilege note is honest (an account key is account-wide → a dedicated test storage account is used).
- **cloud_identity exclusion correct**: `build_object_store_ambient` returns a permanent error for `azure_blob` (keyless azure isn't a product feature), so Azure's credential_ref-only coverage is complete, not an omission.

**Merge gate:** no HIGH/MEDIUM. (Non-blocking note, consistent with the S3/GCS smokes: the black-box round-trip doesn't assert which host it targeted — a wrong-endpoint regression would surface as an auth/connection failure.)

PR #578, opened 2026-06-09
fix(bedrock): wire OpenAI tools/tool_choice through Converse (#560)

### LINKED ISSUE #560: BYO Bedrock Converse tool-calling returns no tool_calls — DP drops tools/tool_choice (confirmed: Nova + Mistral)
## Summary

A `/v1/chat/completions` request through the AISIX DP to a **real** AWS Bedrock model (`us.amazon.nova-lite-v1:0`, Converse path) with an OpenAI `tools` definition returns **no `tool_calls`** — both with `tool_choice:'auto'` and with a **forced** `tool_choice:{type:'function',function:{name:'get_weather'}}`. The response is HTTP 200 but `choices[0].message.tool_calls` is empty (`length === 0`) in both cases.

## Repro

- BYO `bedrock` ProviderKey (real AWS creds), model `us.amazon.nova-lite-v1:0`, region `us-west-2`.
- Request:
  ```jsonc
  {
    "model": "<alias>",
    "tools": [{ "type": "function", "function": {
      "name": "get_weather", "description": "Get the current weather for a city.",
      "parameters": { "type": "object",
        "properties": { "location": { "type": "string" } }, "required": ["location"] } }}],
    "tool_choice": { "type": "function", "function": { "name": "get_weather" } },
    "messages": [{ "role": "user", "content": "What is the current weather in Paris? Use the get_weather tool." }]
  }
  ```
- Result: HTTP 200, `choices[0].message.tool_calls` empty.
- Real-chain CI runs (AISIX-Cloud): `27121925072` (auto) and `27122291025` (forced) — both report `Received: 0` for `tool_calls.length`.

## Expected

With a tool offered — and especially with a **forced** `tool_choice` (Bedrock Converse `toolChoice:{tool:{name}}` makes the model emit a `toolUse` block server-side) — the DP should (1) forward OpenAI `tools`/`tool_choice` → Converse `toolConfig`/`toolChoice`, and (2) translate Bedrock's `toolUse` content block back → OpenAI `tool_calls`. A non-empty `tool_calls` is expected.

## Scope / unknowns (source-blind discovery)

Could be (a) the DP not forwarding `tools`/`tool_choice` to Converse, (b) the DP not translating `toolUse` → `tool_calls`, or (c) a Nova-specific limitation. There is **no existing coverage (mock or real) for Bedrock tool-calling** — `adapter-bedrock-live.spec.ts` (mock) covers Converse SigV4 + streaming + cross-region + Anthropic `/invoke`, but not tools. Worth checking against a second Bedrock publisher (e.g. Claude-on-Bedrock) to isolate DP-vs-model.

**→ Isolation done 2026-06-09 (see "Isolation results" below): rules out (c) Nova-specific; evidence points to (a) — the DP isn't forwarding `tools`/`tool_choice` into the Converse `toolConfig`.**

## Why it matters

Function/tool calling is a core gateway capability; if BYO-Bedrock drops it, those customers can't use tools through the gateway.

## Status

The Bedrock boundary spec (AISIX-Cloud #734) **omits** the tool-calling leg (held back, not shipped failing); the other 4 boundary legs pass. Add the tool-calling leg once this is resolved.

Refs: AISIX-Cloud #694, #731, #734, #735.

## Isolation results — 2026-06-09 (rules out Nova-specific; points to the DP)

Ran the identical **forced** `tool_choice` request through the DP to three Bedrock models (one-off diagnostic, AISIX-Cloud run `27184550090`):

| model | DP dispatch | result |
|---|---|---|
| `us.amazon.nova-lite-v1:0` | Converse | HTTP 200, **tool_calls=0**, `finish_reason=stop`, content = prose ("To provide you with the current weather …") |
| `us.anthropic.claude-3-5-sonnet-20241022-v2:0` | /invoke | ERROR 404 "upstream model not found" — not enabled on this account; no /invoke signal |
| `mistral.mistral-large-2407-v1:0` | Converse | HTTP 200, **tool_calls=0**, `finish_reason=stop`, content = a text code block (` ```python {"name":"get_weather", …} `) |

**Conclusion: DP-side bug on the Bedrock Converse path — not a Nova limitation.**

- Two **different publishers** (Amazon Nova *and* Mistral) on the **same Converse path** fail **identically**: under a *forced* `tool_choice` they return `finish_reason=stop` with the tool call leaking out as **prose/code text**, never a structured `toolUse` block → `tool_calls=0`. If this were a Nova model limitation, Mistral would have worked — it didn't.
- A *forced* Converse `toolChoice:{tool:{name}}` is a **server-side guarantee**: Bedrock would return `stopReason=tool_use` + a `toolUse` block. Getting `finish_reason=stop` + improvised text is exactly what happens when the model **never receives a `toolConfig`** — i.e. root cause **(a): the DP isn't forwarding OpenAI `tools`/`tool_choice` → Converse `toolConfig`/`toolChoice`**. Mistral's ` ```python ` code-block output is the classic signature of a model asked in *prose* to use a tool without the native tool interface.
- The DP doesn't log its outbound Converse body, so the definitive confirmation for the fix-owner is to **inspect the Bedrock Converse request builder: is `toolConfig`/`toolChoice` populated from OpenAI `tools`/`tool_choice`?** The behavioral evidence says no.

**Severity:** confirmed functional bug — **function/tool calling is broken for all BYO-Bedrock _Converse_ models** (silently degrades to a text answer), not just Nova. Suggest re-labeling from `[investigate]` to a confirmed bug. The `/invoke` (anthropic.*) path is still untested here — Claude 404'd on this account; track separately if needed.

Diagnostic spec + branch were throwaway (deleted post-run); the AISIX-Cloud #734 tool-calling leg stays held back until this is fixed.

moonming:
## Summary

Fixes #560. On the Bedrock **Converse** path the dispatch built the SDK call with only `system` / `messages` / `inferenceConfig` and **never set `toolConfig`**, so OpenAI `tools`/`tool_choice` (carried in `ChatFormat.extra`) were dropped. Every non-Anthropic publisher silently degraded tool calling to prose (`finish_reason=stop`, `tool_calls=[]`) — the exact Nova + Mistral behavior in the issue — and streaming dropped tool calls for **all** publishers including Anthropic. The `/invoke` (anthropic.\*) non-stream path was unaffected (it forwards tools via the Anthropic request crate).

## Changes (`crates/aisix-provider-bedrock`)

- **Request** — `build_tool_config` translates OpenAI `tools` → `toolConfig.tools[].toolSpec` (JSON Schema verbatim into `inputSchema.json`) and maps `tool_choice`:
  - `none` → no `toolConfig` (Converse has no `none`; model answers in prose)
  - `auto` / omitted → tools sent, **no explicit** `toolChoice` (Converse defaults to `auto`; avoids the field on publishers that only accept the default)
  - `required` → `toolChoice.any`
  - `{type:function, function:{name}}` → `toolChoice.tool{name}`
- **Response (non-stream)** — `toolUse` content blocks → OpenAI `tool_calls`, with `arguments` as a **JSON-encoded string** (per the OpenAI object spec) and `content:null` when there is no prose. `finish_reason` `ToolUse→tool_calls` was already mapped.
- **Stream** — emits `tool_calls` deltas from `ContentBlockStart`/`ContentBlockDelta(toolUse)`, correlating fragments to a **dense 0-based** `tool_calls[].index` (a leading text block no longer shifts the first call).
- **Multi-turn history** — assistant `tool_calls` → `toolUse` blocks; `role:"tool"` results → `toolResult` blocks **coalesced into one user message** (Converse requires a turn's results in a single user message and rejects consecutive user messages).
- **New `convert.rs`** — `serde_json::Value ↔ aws_smithy_types::Document` (the SDK ships no serde bridge; preserves int-vs-float).

## Approach vs. the ecosystem

Mainstream OpenAI-compatible gateways converge on the same Converse mapping: OpenAI `tools` → `toolConfig.toolSpec` with the JSON Schema forwarded verbatim, `tool_choice` → `auto|any|tool`, and `toolUse` → `tool_calls` with `arguments` rendered as a **stringified** JSON blob (not a parsed object) so OpenAI-SDK callers' `JSON.parse(arguments)` works. This change follows that consensus. Forcing (`any`/`tool`) is honored only by models that support it; on others Bedrock returns a validation error, which is the correct surface for an explicit unsupported force (we don't silently swallow it).

Primary sources:
- OpenAI `tools` / `tool_choice` / response `tool_calls`: <https://platform.openai.com/docs/api-reference/chat/create>
- Bedrock Converse `ToolConfiguration` / `ToolChoice` / `ToolUseBlock`: <https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html>, <https://docs.aws.amazon.com/bedrock/latest/userguide/tool-use.html>
- SDK contract (`toolChoice.tool` is Anthropic Claude 3 + Amazon Nova only; `auto` is the default): `aws-sdk-bedrockruntime` 1.130.0 `types::ToolChoice`.

## Test plan

- `cargo test -p aisix-provider-bedrock` — 68 pass (10 new): forced/required/auto/none request shaping (outbound body capture), `toolUse`→`tool_calls` response (arguments-is-a-string), streaming tool-call deltas, multi-turn history coalescing, and `Document` round-trip.
- `cargo clippy -p aisix-provider-bedrock --all-targets -- -D warnings` clean; `cargo fmt`; `cargo check --workspace` clean.
- **Authoritative end-to-end:** the AISIX-Cloud #734 Bedrock boundary **tool-calling leg** (held back pending this fix) — unholstered separately; runs label-gated against live Bedrock (Nova + Mistral).

Closes #560.

## Summary by CodeRabbit

* **New Features**
  * End-to-end Bedrock Converse tool-calling support, including tool choice options and coalescing of tool results for multi-turn conversations
  * Streaming now preserves correct tool-call indexing and emits incremental tool argument deltas
  * JSON ↔ Bedrock document conversion to reliably translate tool inputs/outputs and numeric types

* **Tests**
  * Added comprehensive tool-calling and conversion test coverage, including streaming, index correlation, and malformed-parameter handling

### DIFF
--- a/crates/aisix-provider-bedrock/src/bridge.rs
+++ b/crates/aisix-provider-bedrock/src/bridge.rs
@@ -1573,12 +1573,15 @@ fn build_tool_config(req: &ChatFormat) -> Option<ToolConfiguration> {
         if let Some(description) = function.get("description").and_then(|v| v.as_str()) {
             spec = spec.description(description);
         }
-        // OpenAI `parameters` is a JSON Schema object; default to an empty
-        // object schema for a no-argument tool.
-        let parameters = function
-            .get("parameters")
-            .cloned()
-            .unwrap_or_else(|| serde_json::json!({"type": "object", "properties": {}}));
+        // OpenAI `parameters` is a JSON Schema object. Default to an empty
+        // object schema when omitted; skip the tool if `parameters` is
+        // present but not an object — a non-object schema is malformed and
+        // Bedrock would reject the whole Converse request with a 400.
+        let parameters = match function.get("parameters") {
+            Some(p) if p.is_object() => p.clone(),
+            Some(_) => continue,
+            None => serde_json::json!({"type": "object", "properties": {}}),
+        };
         spec = spec.input_schema(ToolInputSchema::Json(json_to_document(&parameters)));
         if let Ok(spec) = spec.build() {
             tools.push(Tool::ToolSpec(spec));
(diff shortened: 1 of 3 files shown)


moonming:
## Independent pre-merge audit (per CLAUDE.md §8)

A fresh, cold-context `general-purpose` agent audited this PR against the claimed contract. **Verdict: APPROVE-WITH-NITS — no CRITICAL or HIGH findings.** It verified (against the real `aws-sdk-bedrockruntime` 1.130.0 source + the gateway's `render_response`/`render_chunk` path): request `toolConfig` shape, the `tool_choice` mapping, `tool_calls` with `arguments` as a JSON-encoded string, `content:null` vs `""`, the streaming dense-index logic, multi-turn coalescing, panic-freedom on malformed input, the closed leak surface (canned error message + fixed code taxonomy), and that the no-tool path is byte-identical to base (non-breaking).

Findings addressed in `57a3bf8`:

|---|-----|---------|------------|
| M1 | MED | Streaming index test only covered a single tool block at index 0 — the "leading text doesn't shift the first call" + parallel-call claims were unverified | Added `emit_converse_chunk_tool_call_index_is_dense_despite_leading_text`: text at block 0, two tool blocks at 1/2 → asserts dense OpenAI indices 0,1 |
| M2 | MED | No test for prose **+** tool call (the `content` non-null branch of `build_assistant_message`) | Added `chat_converse_preserves_prose_alongside_tool_calls`: response with a text block + toolUse → `content:"On it."` preserved alongside `tool_calls` |
| M3 | MED | `parameters` forwarded verbatim — a non-object (caller-controlled) would produce an invalid `inputSchema.json` and a Bedrock 400 | Guard added: skip a tool whose `parameters` is present-but-not-an-object; valid sibling tools still go through. Pinned by `chat_converse_skips_tool_with_non_object_parameters` |
| L1 | LOW | `role:"tool"` without `tool_call_id` dropped | Already documented inline as intentional (can't correlate to a `toolUse`; the alternative is an upstream 400 on an ambiguous `toolResult`) — behavior kept, justified |
| L2 | LOW | `inputSchema` optional in the SDK builder but required by the service | No change — code always sets `input_schema`; noted as a future-refactor caution only |

Crate tests now 71 pass / 0 fail; `clippy -D warnings` + `fmt` clean.

The authoritative end-to-end (real Bedrock) coverage is the AISIX-Cloud #734 tool-calling leg, wired in at api7/AISIX-Cloud#758.

PR #598, opened 2026-06-12
refactor(obs): unify Prometheus scrape on the dedicated metrics listener

jarvis9443:
## Problem

The Prometheus scrape surface depended on deployment mode: standalone DPs served `/metrics` on the admin listener, while managed DPs (which never bind the admin listener) needed a dedicated listener opted in via `observability.metrics.prometheus.addr`. Two topologies for the same endpoint made the mechanism unnecessarily confusing to configure and document.

## Change

The dedicated metrics listener is now the **only** scrape surface, identical in standalone and managed mode:

- `observability.metrics.prometheus.addr` becomes a required field with default `0.0.0.0:9090`; the listener is bound whenever `prometheus.enabled` is true (still failing boot loudly if the port is taken).
- The admin router no longer mounts `/metrics` — the admin-side handler, the `AdminState` metrics/prometheus fields, and the OpenAPI `/metrics` entry are removed. `metrics_router` keeps serving the configured path on the dedicated listener.
- `config.example.yaml` and `config.managed.yaml` both pin `addr: "0.0.0.0:9090"` explicitly; the Dockerfile now `EXPOSE`s 9090.
- Docs (metrics-and-logs, admin-api, admin-api-reference, bootstrap-config, network-and-security) updated to describe the single unified surface.

## Behavior change / compatibility

- Scraping `http://<dp>:3001/metrics` (admin listener) no longer works for standalone deployments — point Prometheus at `<dp>:9090/metrics` instead. The project is in early development with no backward-compatibility constraints.
- Managed DPs are unaffected (they already scraped `0.0.0.0:9090` via the baked managed config).
- `prometheus.enabled: false` now means no metrics listener is bound at all.

## Tests

- Unit: config default/parse/validation tests updated; admin tests now pin that the admin router returns 404 for `/metrics`; `metrics_router` path/normalization tests retained.
- E2E (`tests/e2e`): the harness always provisions the metrics listener and exposes `metricsUrl`; all scrape-based tests moved to it. New/updated cases pin that the metrics listener is the only scrape surface (admin 404s `/metrics`, metrics port carries no admin routes) and that `enabled: false` leaves the port unbound. The TLS test now uses an explicit free metrics port so concurrent test files cannot collide on the 9090 default.

LiteLLM baseline note: LiteLLM mounts `/metrics` on its main proxy app and has no admin/managed listener split, so there is no equivalent topology to compare against; the unified dedicated-listener design was decided explicitly for this project.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -86,8 +86,8 @@ COPY config.managed.yaml /etc/aisix/config.managed.yaml
 COPY docker/entrypoint.sh /usr/local/bin/aisix-entrypoint
 RUN chmod 0755 /usr/local/bin/aisix-entrypoint
 
-# Proxy + admin listeners from config.example.yaml.
-EXPOSE 3000 3001
+# Proxy + admin + metrics listeners from config.example.yaml.
+EXPOSE 3000 3001 9090
 
 USER aisix
 
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -54,13 +54,11 @@ observability:
     prometheus:
       enabled: true
       path: "/metrics"
-      # Optional: serve `/metrics` on a dedicated listener instead of
-      # relying on the admin listener. Set this when the admin surface is
-      # private but you still want Prometheus to scrape, or to keep the
-      # scrape port separate from the data path. Required for managed-mode
-      # DPs (the admin listener is not bound there). Unset = `/metrics`
-      # stays on the admin listener.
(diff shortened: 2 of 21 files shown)


jarvis9443:
The CodeRabbit notice above is an automated rate-limit/billing message, not code feedback — nothing to address. All CI checks are green and the change was verified locally (full cargo test + 89-file e2e suite), so proceeding with merge.

PR #604, opened 2026-06-15
feat: ensemble model — panel + judge fan-out dispatch (config + non-streaming)

### LINKED ISSUE #601: [DP] ensemble model — parallel panel fan-out + judge synthesis
## Scope

This is the **data-plane (ai-gateway) implementation** of the `ensemble` virtual-model feature. The umbrella issue — full rationale, the CP / dashboard / schema work, and the resolved v1 decisions — lives in the control-plane planning repo: **api7/AISIX-Cloud#804**.

An `ensemble` model fans one `/chat/completions` request out to a **panel** of N upstream models in parallel, then calls a **judge** model to synthesize a single final answer. It extends today's virtual-model machinery: `routing` models pick **one** target; `ensemble` calls **all** panel targets concurrently and synthesizes. All fan-out + synthesis happen in the DP.

## Wire contract the DP must accept

The CP projects this KV into kine; the DP deserializes it on `Model`. Adding the `ensemble` block requires extending the DP `Model` struct (`#[serde(deny_unknown_fields)]`) **and** the generated `schemas/resources/model.schema.json` (`additionalProperties:false`) together — otherwise the row is silently dropped and the model 404s.

```json
{
  "display_name": "<ensemble model alias>",
  "ensemble": {
    "panel": [ { "model": "<direct model display_name>", "temperature": 0.7, "seed": 1, "weight": 1 } ],
    "judge": { "model": "<direct model display_name>", "synthesis_prompt": "<optional override>" },
    "min_responses": 2,
    "timeout_ms": 60000
  }
}
```

- Panel / judge reference **direct** models only (no nesting), same environment.
- `panel[].temperature` (+ optional `seed`) overrides the request temperature **for panel members** — the diversity source for self-ensemble.
- `panel[].weight` is reserved for the Phase-2 voting strategy; ignored in v1.
- The judge runs at a fixed low temperature (~0.2).

## DP work breakdown

- [x] **Config types**: `ensemble: Option<EnsembleConfig>` on `Model` (`crates/aisix-core/src/models/model.rs`); `EnsembleConfig { panel: Vec<PanelMember>, judge: Judge, min_responses, timeout_ms }`; deserialize from the kine KV (respect `deny_unknown_fields`). Mirror the `Routing` types' shape and validation.
- [x] **Dispatch**: detect the ensemble model in `chat_completions` (`crates/aisix-proxy/src/chat.rs`) → ensemble executor instead of the single-bridge path. Reject on non-chat endpoints (`crates/aisix-proxy/src/lib.rs`) and on requests carrying a non-empty `tools` array (400).
- [x] **Fan-out executor**: resolve each panel member → Bridge; concurrent `Bridge::chat()` (`FuturesUnordered`) with a per-target timeout; apply the per-member `temperature`/`seed` override; enforce `min_responses`.
- [x] **Synthesis**: judge `Bridge::chat()` at the fixed low temp using the default template (umbrella Appendix) or the per-model `judge.synthesis_prompt`; judge message shape = neutral `Answer 1..N` (never model names) + the original request; per-answer truncation / total-budget cap to avoid judge context overflow.
- [x] **Streaming**: support `stream=true` (do not 400 — many clients default to it). Buffer the panel with SSE keep-alive heartbeats (reuse the `KeepAlive` from the routing stream path), then stream the judge via `chat_stream()`. Document the high TTFT.
- [x] **Usage / cost**: client-facing `usage` = aggregate (sum panel + judge); telemetry = one `UsageEvent` per sub-call sharing the parent `request_id`, with a new `attempt_kind` = `panel` / `judge` (`crates/aisix-obs/src/usage.rs`, per-attempt model from #655).
- [ ] **Rate limiting (net-new)**: reserve each panel member's own `rate_limit` before dispatch (today only the entry model is debited — `quota.rs`, `ModelRateLimit::from_model`); define rollback on a `min_responses` abort.
- [x] **Guardrails**: input guardrail once pre-fan-out; output guardrail on the synthesized answer. Verify ordering.
- [ ] **Caching**: the cache key must include the ensemble config (panel + judge + prompt), not just the input.
- [x] **Error semantics**: partial panel OK if successes ≥ `min_responses`; judge failure → 5xx after one in-process retry; no silent fallback to a raw panel answer.
- [x] **schema.json**: regenerate `schemas/resources/model.schema.json` to admit `ensemble`; add a round-trip contract test (the CP keeps the mirror).
- [x] **Tests**: concurrency, k-of-N partial failure, judge fail + retry, per-target timeout, per-member sampling, usage-per-sub-call, non-chat + tools rejection.

## Suggested PR sequence

1. **Config layer** — `EnsembleConfig` types + deserialization + validation + `schema.json` + contract test. Self-contained; unblocks the rest.
2. **Fan-out executor (non-streaming)** — dispatch detection + concurrent panel + `min_responses` + judge synthesis with the default prompt + usage-per-sub-call. End-to-end non-streaming.
3. **Streaming** — panel keep-alive + judge token stream.
4. **Rate limiting** — per-target reservation + rollback.
5. **Hardening** — caching key, guardrail ordering, truncation / budget cap, error / retry edges.

The resolved v1 decisions binding this work are listed in AISIX-Cloud#804 under "Resolved decisions (2026-06-15)".

moonming:
## Summary

First slice of the **ensemble** virtual-model feature (DP side, tracked in #601; CP umbrella api7/AISIX-Cloud#804). An `ensemble` model fans one `/v1/chat/completions` request out to a **panel** of N upstream models in parallel, then a **judge** model synthesizes a single answer. It extends the existing virtual-model machinery: `routing` picks **one** target; `ensemble` calls **all** and synthesizes.

This PR lands the **config layer** and the **non-streaming dispatch**. Streaming, per-target rate-limiting, and hardening are separate follow-ups (see Deferred) — so this does **not** close #601.

## What's in this PR

**`1a89a6e` — config layer (new `ensemble` model kind)**
- `EnsembleConfig { panel: [{model, temperature?, seed?, weight?}], judge: {model, synthesis_prompt?}, min_responses?, timeout_ms? }` in `aisix-core`.
- `Model.ensemble` + `Model::is_ensemble()` (DP convention: virtual-model kind = presence of the block, same as `routing`).
- Both DP schemas updated in lockstep so an ensemble row is never silently dropped on the kine watch path: the schemars-generated `schemas/resources/model.schema.json` (+ a standalone `ensemble.schema.json`) and the hand-written runtime validator `models/schema.rs::model_schema()`, whose `oneOf` is now a 3-way XOR (`direct | routing | ensemble`).

**`8375c85` — non-streaming fan-out dispatch**
- `ensemble.rs`: pure executor `run_ensemble` (parallel `join_all` fan-out, `min_responses` gate, per-member `temperature`/`seed` override for self-ensemble diversity, judge synthesis with neutral `Answer 1..N` labels + per-candidate truncation, judge retry-once-on-transient), plus the production `ProxyModelCaller` that resolves each member → ProviderKey → Bridge via the existing routing helpers.
- `chat.rs::dispatch_ensemble`: branches at the post-rate-limit-reservation seam; rejects `tools`/`tool_choice` and `stream:true` with 400; runs the output guardrail on the synthesized answer; commits the aggregate panel+judge tokens once against the single entry-level reservation; emits **one usage event per sub-call** (`attempt_kind` = `panel`/`judge`, sharing `request_id`) **before** the guardrail check so a blocked response still bills the full fan-out; suppresses the entry-level usage event (no double-emit).
- The client sees the ensemble's own model name; panel/judge model names never reach the client (synthesized answer, `response.model`, or error messages).

## Behavior / contract

- Config-driven: the operator curates the panel + judge; clients just call the model name. No per-request panel override.
- Self-ensemble works with a single provider key (panel = the same model with per-member temperature).
- v1 scope: chat-only; no server-side web retrieval; `tools`/streaming rejected (see Deferred).

## Deferred (separate PRs, tracked in #601)

- **Streaming** — `stream:true` is 400-rejected here; there is no single-`ChatResponse`→SSE path today, so it's net-new.
- **Per-target rate-limiting** — today only the entry model is debited; per-panel-member reservation is net-new (avoids N× provider-quota amplification).
- **Hardening** — the response cache key must include the ensemble config; misc edges.
- **[CP] api7/AISIX-Cloud#804** — schema/projection/dashboard so operators can create ensemble models; the dashboard must also learn the new `attempt_kind` values `panel`/`judge` (free-string on the wire — won't break ingestion).

## Independent audit (merge-gate)

An independent cold-review agent audited this change. It confirmed: no double-emit / double-commit; authorization consistent with routing (panel/judge are internal targets — no escalation beyond what routing already allows); sound snapshot lifetimes across the fan-out; model names not leaked to the client response body. It caught and we fixed:

- **HIGH** — on an output-guardrail *block*, the per-sub-call usage events were skipped while the panel tokens were already committed → cp-api under-reported a blocked-but-billed request. Fixed (emit before the guardrail check on both paths; `charge: None` on block) and locked with `ensemble_output_block_still_emits_panel_and_judge_usage`.
- **LOW** — a misconfigured judge's model name leaked into the client-visible `error.message` → redacted (detail kept in server logs).
- **LOW** — `timeout_ms` doc/behavior mismatch → behavior made uniform (now also applies to the judge call) + doc corrected.

## Test plan

- `cargo test -p aisix-core` — **259 pass** (config types, runtime-validator 3-way XOR, schema round-trip; e.g. `model_ensemble_with_direct_fields_fails`, `model_ensemble_with_routing_fails`).
- `cargo test -p aisix-proxy` — **433 pass**: 8 executor unit tests + 5 handler e2e through a real `OpenAiBridge` against wiremock:
  - `ensemble_fans_out_to_panel_and_returns_judge_synthesis` (judge answer returned, `model == "council"`, no upstream-id leak)
  - `ensemble_rejects_tool_requests_with_400`, `ensemble_rejects_streaming_with_400`
  - `ensemble_insufficient_panel_returns_502`
  - `ensemble_output_block_still_emits_panel_and_judge_usage` (HIGH regression lock — asserts 2 panel + 1 judge usage events still fire on a block)
- `cargo clippy --all-targets -- -D warnings` — clean; full workspace builds.
- [ ] CP-side e2e (create an ensemble model via the dashboard → call it end-to-end) lands with api7/AISIX-Cloud#804.

## New Features
* Added ensemble model support for chat completions, running multiple panel models in parallel and using a designated judge model to synthesize the final response.
* Supports per-panel sampling overrides plus configurable minimum successful responses and optional per-member timeouts.
* Enforces ensemble-only behavior in chat completions, with validation for streaming/tool fields and improved error handling.

## Bug Fixes
* Ensured schema generation and validation include the new ensemble configuration, with strict JSON field rejection and correct canonical schema outputs.

### DIFF
--- a/crates/aisix-core/src/bin/dump-schema.rs
+++ b/crates/aisix-core/src/bin/dump-schema.rs
@@ -31,8 +31,8 @@ use std::path::{Path, PathBuf};
 use schemars::JsonSchema;
 
 use aisix_core::models::{
-    ApiKey, CachePolicy, Guardrail, Model, ObservabilityExporter, ProviderKey, RateLimit,
-    RateLimitPolicy, Routing,
+    ApiKey, CachePolicy, EnsembleConfig, Guardrail, Model, ObservabilityExporter, ProviderKey,
+    RateLimit, RateLimitPolicy, Routing,
 };
 
 fn main() {
@@ -41,6 +41,7 @@ fn main() {
 
     dump::<ApiKey>(&out_dir, "api_key");
     dump::<CachePolicy>(&out_dir, "cache_policy");
+    dump::<EnsembleConfig>(&out_dir, "ensemble");
     dump::<Guardrail>(&out_dir, "guardrail");
     dump::<Model>(&out_dir, "model");
     dump::<ObservabilityExporter>(&out_dir, "observability_exporter");
--- a/crates/aisix-core/src/models/ensemble.rs
+++ b/crates/aisix-core/src/models/ensemble.rs
@@ -112,10 +112,6 @@ impl EnsembleConfig {
             .filter(|&ms| ms > 0)
             .map(std::time::Duration::from_millis)
     }
-
-    pub fn is_empty(&self) -> bool {
-        self.panel.is_empty()
-    }
 }
 
 #[cfg(test)]
@@ -157,7 +153,7 @@ mod tests {
         assert!(e.panel[0].temperature.is_none());
         assert!(e.judge.synthesis_prompt.is_none());
         assert_eq!(e.timeout(), None);
-        assert!(!e.is_empty());
+        assert!(!e.panel.is_empty());
     }
 
     #[test]
--- a/crates/aisix-core/src/models/ensemble.rs
+++ b/crates/aisix-core/src/models/ensemble.rs
@@ -1,0 +1,215 @@
+//! Ensemble config attached to a [`Model`](super::Model).
+//!
+//! When a Model carries an `ensemble` block, the proxy fans the request
+//! out to every panel member concurrently, then calls the judge model to
+//! synthesize a single answer from the panel responses. Unlike a
+//! [`Routing`](super::Routing) model — which picks ONE target per request
+//! — an ensemble model calls ALL panel members and combines their output.
+//!
+//! Panel members and the judge reference other (direct) Models by
+//! `display_name`, the same way routing targets do. Mutual exclusivity
+//! with the direct-upstream fields and `routing` is enforced by the
+//! runtime schema (`super::schema`), not by this type.
+
+use serde::{Deserialize, Serialize};
+
+/// One member of an ensemble panel. `model` references another
+/// `Model.display_name` in the snapshot (must be a direct model).
+#[derive(Debug, Clone, Serialize, Deserialize, schemars::JsonSchema, PartialEq)]
+#[serde(deny_unknown_fields)]
+pub struct PanelMember {
+    pub model: String,
+    /// Sampling temperature for THIS member's call. Overrides the
+    /// request's `temperature`, so a panel of the same model repeated N
+    /// times still produces diverse answers (self-ensemble). Absent =
+    /// leave the request's temperature unchanged.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub temperature: Option<f32>,
+    /// Optional sampling seed for THIS member's call — pairs with
+    /// `temperature` for reproducible diversity. Absent = no seed override.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub seed: Option<u64>,
+    /// Reserved for the future voting/quorum strategy; ignored by the v1
+    /// synthesis path. Carried on the wire now so enabling voting later
+    /// does not require a wire-format change.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub weight: Option<u32>,
+}
+
+impl PanelMember {
+    pub fn new(model: impl Into<String>) -> Self {
+        Self {
+            model: model.into(),
+            temperature: None,
+            seed: None,
+            weight: None,
+        }
+    }
+}
+
+/// The judge model that synthesizes the panel responses into one answer.
+/// `model` references another `Model.display_name` (must be a direct model).
+#[derive(Debug, Clone, Serialize, Deserialize, schemars::JsonSchema, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
+pub struct Judge {
+    pub model: String,
+    /// Optional override for the built-in synthesis prompt template.
+    /// Absent = use the default template.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub synthesis_prompt: Option<String>,
+}
+
+impl Judge {
+    pub fn new(model: impl Into<String>) -> Self {
+        Self {
+            model: model.into(),
+            synthesis_prompt: None,
+        }
+    }
+}
+
+/// Default minimum number of successful panel responses required before
+/// the judge synthesizes, when the operator does not set `min_responses`.
+/// Always clamped to the panel size by [`EnsembleConfig::min_responses_or_default`],
+/// so a single-member (self-ensemble) panel still only needs one response.
+const DEFAULT_MIN_RESPONSES: usize = 2;
+
+#[derive(Debug, Clone, Serialize, Deserialize, schemars::JsonSchema, PartialEq)]
+#[serde(deny_unknown_fields)]
+pub struct EnsembleConfig {
+    pub panel: Vec<PanelMember>,
+    pub judge: Judge,
+    /// Minimum successful panel responses required to proceed to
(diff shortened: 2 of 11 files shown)


moonming:
## Post-PR independent audit — all findings resolved (`917d692`)

Per our merge-gate, an independent cold-review agent audited this PR. Findings and resolutions:

**HIGH**
- **Client-facing leak** — a misconfigured judge's `display_name` + `provider_key_id` reached the client `error.message` (the earlier redaction only covered the `get_by_name` path; `resolve_provider_key` still forwarded it). **Fixed** — redacted; detail kept in server logs. Test: `ensemble_misconfigured_judge_does_not_leak_internal_config`.
- **Schema-drift CI would be red** — the `timeout_ms` doc edit wasn't regenerated into the schema JSON. **Fixed** — regenerated; `dump-schema` + `git diff --exit-code schemas/` clean.

**MEDIUM**
- **Billing under-report on error exits** — panel members that already round-tripped upstream weren't billed on the insufficient-panel (502) path. **Fixed** — emit usage + commit tokens for the survivors on the 502 path, matching the output-guardrail-block path. Also extended to the **judge-failure** path (same class, surfaced during the fix). Tests: `ensemble_insufficient_panel_returns_502`, `ensemble_judge_failure_still_bills_panel`.
- **`tools: []` wrongly 400'd** — empty arrays (which many SDKs always send) were rejected. **Fixed** — only a non-empty `tools` or a forcing `tool_choice` rejects. Tests: `ensemble_allows_empty_tools_array`, `ensemble_allows_tool_choice_none`.
- **Non-chat rejection incidental + misleading** — non-chat endpoints rejected ensemble models with a "routing" message and no test. **Fixed** — explicit, accurate 400. Test: `ensemble_model_on_embeddings_returns_400_with_explicit_message`.

**Verified clean by the audit**: the 3-way schema XOR (`direct | routing | ensemble`), single token-commit, double-emit suppression, response-body redaction, and the cross-repo wire contract vs api7/AISIX-Cloud#804.

**Tracked follow-ups (not blocking this PR)**:
- Total judge-prompt budget cap (per-candidate truncation exists; a panel-wide cap folds into the hardening PR under #601).
- [CP] the dashboard should recognize the new `attempt_kind` values `panel`/`judge` (free-string on the wire — won't break ingestion) — tracked under api7/AISIX-Cloud#804.

259 core + 439 proxy tests green; `clippy --all-targets -- -D warnings` clean; schema-drift gate clean.

PR #606, opened 2026-06-15
feat: stream ensemble judge synthesis

moonming:
## Summary

Adds streaming support for `ensemble` models (DP; follow-up to #604, tracked in #601). `stream:true` to an ensemble model now streams the judge's synthesized answer token-by-token, instead of the previous 400.

## Design: Option A — judge-leg streaming, shared builder reused unmodified

The panel must be buffered to synthesize, so phases 1-2 run non-streaming; only the **judge's** tokens stream. This reuses the existing `build_sse_stream` **unmodified** — all change is isolated to the ensemble dispatch branch; the shared SSE builder and the single-upstream streaming path are byte-for-byte untouched. (Safety property: a bug in the shared builder would hit *every* streaming request, so this PR deliberately doesn't touch it.)

**Deferred — panel-phase keep-alive.** The panel fan-out runs before the `Sse` is constructed, so the 15s SSE keep-alive doesn't cover it. This is **no worse than the already-merged non-streaming ensemble path**, which likewise holds the socket bytes-free across panel + judge; TTFB actually improves (judge-first-token vs full answer). True panel keep-alive needs an early SSE handshake before the panel completes — a self-contained follow-up kept out of this PR to avoid refactoring the shared builder. Marked with a `TODO`.

## What's in it

- **`ensemble.rs`**: split `run_ensemble` into `run_ensemble_panel` (phases 1-2 + `judge_request`) + the judge call, so the judge leg can be streamed. No behavior change to the non-streaming path (all existing executor + handler tests pass unchanged).
- **`chat.rs::dispatch_ensemble`** streaming arm: run the panel → resolve the judge bridge → `judge.chat_stream()` → feed it to `build_sse_stream` with an `on_complete` that emits one usage event per sub-call (panel + judge, `attempt_kind` `panel`/`judge`, sharing `request_id`) and commits the panel+judge tokens via `add_tokens_post_stream` (the streaming token-commit API). `telemetry_handled_by_stream: true` suppresses the entry-level emit — no double-emit.
- The **"successful panel members are always billed"** invariant now also covers the judge-connect-failure and judge-resolution-failure exits (the panel already round-tripped upstream).
- Client sees the ensemble model name on every chunk (re-stamped); judge model names never leak; judge-resolution errors are redacted to generic messages (detail to server logs).

## Test plan

- `cargo test -p aisix-core` — **259 pass**.
- `cargo test -p aisix-proxy` — **440 pass**, incl. 2 new e2e through real wiremock bridges:
  - `ensemble_streams_judge_synthesis` (replaces the now-obsolete `ensemble_rejects_streaming_with_400`): `stream:true` → SSE body contains the judge content + `[DONE]`, every chunk `model == "council"` (not the upstream id), and the usage sink shows 2 panel + 1 judge events sharing one `request_id`.
  - `ensemble_streaming_judge_connect_failure_still_bills_panel`: judge 500 on the streaming connect → 502, panel usage events still fire.
- `cargo clippy --all-targets -- -D warnings` clean; schema-drift gate clean.

An independent audit will follow (merge-gate); I'll post findings + resolutions before merge.

## Summary by CodeRabbit

* **New Features**
  * Ensemble models now support streaming API responses, enabling real-time output delivery.
  * Proper usage tracking and billing for streamed ensemble responses.

### DIFF
--- a/crates/aisix-proxy/src/chat.rs
+++ b/crates/aisix-proxy/src/chat.rs
@@ -1899,15 +1899,6 @@ async fn dispatch_ensemble<'a>(
             "ensemble models do not support tools".into(),
         )));
     }
-    // TODO(PR3): ensemble streaming. There is no single-ChatResponse→SSE
-    // helper today, and streaming the judge's synthesized answer is a
-    // separate piece of work; reject streaming requests for now.
-    if req.is_streaming() {
-        return Err(with_model(ProxyError::InvalidRequest(
-            "streaming is not supported for ensemble models".into(),
-        )));
-    }
-
     // `is_ensemble()` is the branch guard, so `ensemble` is always Some
     // here; surface a 400 rather than panic if a future refactor breaks
     // that invariant.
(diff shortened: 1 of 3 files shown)


moonming:
## Independent audit — safe to merge

A cold-review agent audited all 8 risk points; **verdict: safe to merge** (zero HIGH, zero MEDIUM).

Verified:
- `build_sse_stream` + the single-upstream streaming path are **byte-for-byte unchanged** — the diff only touches `dispatch_ensemble` (+ a `#[cfg(test)]` block in lib.rs).
- **No double-emit / double-commit**: the borrow checker forces the reservation consumed once (`commit_tokens` on the error exits XOR `into_stream_hold` on success); `telemetry_handled_by_stream:true` gates the entry-level emit.
- `on_complete` fires **exactly once** via `CompleteOnDrop` on both normal completion and client disconnect; disconnect-before-first-chunk bills panel-full + judge-prompt-only (matches the #419 cost-leak gate).
- The "successful panel members are always billed" invariant holds on **all five** error exits (insufficient panel, judge connect fail, unknown judge model, unresolved PK, no bridge).
- No judge identity leaks: chunk `model` re-stamped to the ensemble name, judge upstream 5xx message suppressed, resolution errors are static literals.
- Both new e2e drive the **real** handler + `build_sse_stream` + `OpenAiBridge` + wiremock (no mock caller).

**LOW follow-ups (non-blocking; tracked for the hardening pass under #601):**
- **L1** — add a streaming output-guardrail-block e2e (the non-streaming path has one; the streaming `comp.guardrail_blocked`→panel-event wiring is audit-verified-correct but not test-locked).
- **L2** — client-disconnect-mid-stream billing (correct by inspection; hard to make deterministic).
- **L3** — `stream_options.include_usage:true` e2e (shared code path with single-upstream).
- **L4** — panel-phase socket keep-alive (already documented `TODO`).

PR #636, opened 2026-06-22
docs: use release image in Docker examples

kayx23:
## Summary

- Update the self-hosted quickstart Compose example from `ghcr.io/api7/aisix:dev` to `ghcr.io/api7/aisix:0.1.0`.
- Update entrypoint Docker examples to use the same release image tag.

## Context

AISIX now has versioned release artifacts. User-facing examples should use released image tags by default, while `dev` remains available for validating unreleased changes.

## Verification

- Confirmed `v0.1.0` points to current `main` at `e80dac38e1617ee4b0a0da5e9fff00aa1ed89075`.
- Previously verified `ghcr.io/api7/aisix:0.1.0` pulls successfully.
- Ran `git diff --check`.

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -5,14 +5,14 @@
 # /etc/aisix/config.yaml). Two intended modes:
 #
 #   Standalone — operator mounts their own config:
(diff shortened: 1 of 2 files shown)


kayx23:
Changes not needed. Docs will be replaced later in batch and example script change is out of scope.

PR #660, opened 2026-06-29
feat(mcp): McpBridge — upstream MCP client over Streamable HTTP (DP-2)

moonming:
## What

First implementation step of the **MCP gateway** ([AISIX-Cloud#894](https://github.com/api7/AISIX-Cloud/issues/894), `Revision 2026-06-29` → DP-2). Adds a new `aisix-mcp` crate that opens a governed session to **one** upstream MCP server over **Streamable HTTP** and exposes it behind an `McpBridge` trait:

- `list_tools()` → normalised `Vec<McpTool>`
- `call_tool(name, args)` → normalised `McpToolResult`
- upstream auth (`None` / `Bearer`) is **held gateway-side**, never sourced from the downstream caller.

This is the highest-risk slice of the feature (does the official Rust SDK actually work for our client needs over Streamable HTTP, on our toolchain?), so it ships first as a foundation. Aggregating N upstreams behind the downstream-facing `/mcp` endpoint, tool namespacing, per-tool ACL, and wiring MCP traffic into the shared guardrail/quota pipeline are the next steps (DP-3/DP-4) and are **not** in this PR.

## How

- Built on the official MCP Rust SDK **`rmcp`, pinned `=1.8.0`** (targets MCP spec `2025-11-25`). rmcp does the JSON-RPC + Streamable HTTP + session plumbing; we own the aggregation/governance layer (later).
- **API-drift containment:** all rmcp types are converted to this crate's own DTOs (`McpTool`, `McpToolResult`, `McpUpstream`, `McpAuth`) at the boundary. The rest of the data plane depends only on the `McpBridge` trait, not on rmcp. rmcp is <16 months old with ~monthly breaking changes, hence the exact pin + the trait seam.
- **Minimal dependency surface:** production links only rmcp's **client**; the **server** side is a `dev-dependency`, used solely to stand up a real upstream in the test.
- `rmcp` relationship is a **pure in-process Rust library** (compiled into the binary) — no sidecar, no extra network hop.

## Reference implementations studied (per repo CLAUDE.md §7)

- **rmcp** (official Rust SDK) — chosen build-on; its gap (no high-level proxy/aggregation) is exactly the layer we must own.
- **LiteLLM / Kong** MCP gateways — confirmed the intersection feature set; upstream-auth-held-server-side + no-token-passthrough is both their behaviour and a spec `MUST` (MCP authorization 2025-11-25).
- **MCP spec 2025-11-25** transports + authorization (primary source). Note: the `2026-07-28` RC removes the session header → the `/mcp` *server* endpoint (DP-3) will be built stateless; this client step is unaffected.

## Test plan

Integration test (`tests/upstream_roundtrip.rs`) runs against a **real** rmcp Streamable HTTP server on an ephemeral port (no mock transport):

- [x] `initialize` → `tools/list` → `tools/call` round-trips; echo tool returns its argument; schema surfaced.
- [x] unknown tool surfaces as an **error**, not a silent empty result.
- [x] gateway-held **Bearer is forwarded** to the upstream on every request; **absence is rejected** (401 at `initialize`).
- [x] `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test -p aisix-mcp` all green locally.

Refs AISIX-Cloud#894

## Summary by CodeRabbit

* **New Features**
  * Added a new MCP client crate to connect to an upstream server over Streamable HTTP.
  * Exposed tool discovery and tool invocation through a simple bridge interface.
  * Added support for optional bearer-token authentication when connecting upstream.
  * Introduced a structured error type for connection and request failures.

* **Tests**
  * Added end-to-end coverage for listing tools, calling tools, and verifying bearer-token forwarding.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -119,7 +119,7 @@ dependencies = [
  "dashmap",
  "futures",
  "http 1.4.0",
- "reqwest",
+ "reqwest 0.12.28",
  "serde",
  "serde_json",
  "thiserror 1.0.69",
@@ -142,7 +142,7 @@ dependencies = [
  "chrono",
  "hmac 0.12.1",
  "regex",
- "reqwest",
+ "reqwest 0.12.28",
  "serde",
  "serde_json",
  "sha1",
@@ -151,6 +151,19 @@ dependencies = [
  "tracing",
  "uuid",
  "wiremock",
+]
+
+[[package]]
+name = "aisix-mcp"
+version = "0.1.0"
+dependencies = [
+ "async-trait",
+ "axum",
+ "rmcp",
+ "serde",
+ "serde_json",
+ "thiserror 1.0.69",
+ "tokio",
 ]
 
 [[package]]
@@ -171,7 +184,7 @@ dependencies = [
  "metrics-exporter-prometheus",
  "object_store",
  "parking_lot",
- "reqwest",
+ "reqwest 0.12.28",
  "serde",
  "serde_json",
  "sha2 0.10.9",
@@ -193,7 +206,7 @@ dependencies = [
  "async-trait",
  "bytes",
  "futures",
- "reqwest",
+ "reqwest 0.12.28",
  "serde",
  "serde_json",
  "thiserror 1.0.69",
@@ -214,7 +227,7 @@ dependencies = [
  "bytes",
  "futures",
  "http 1.4.0",
- "reqwest",
+ "reqwest 0.12.28",
  "serde",
  "serde_json",
  "tokio",
@@ -256,7 +269,7 @@ dependencies = [
  "bytes",
  "futures",
  "http 1.4.0",
- "reqwest",
+ "reqwest 0.12.28",
  "serde",
  "serde_json",
  "tokio",
@@ -278,7 +291,7 @@ dependencies = [
  "futures",
  "http 1.4.0",
  "jsonwebtoken",
- "reqwest",
+ "reqwest 0.12.28",
  "serde",
  "serde_json",
  "tokio",
@@ -307,7 +320,7 @@ dependencies = [
  "http-body-util",
  "ipnet",
  "rand 0.8.5",
- "reqwest",
+ "reqwest 0.12.28",
  "serde",
  "serde_json",
  "thiserror 1.0.69",
@@ -371,7 +384,7 @@ dependencies = [
  "etcd-client",
  "hostname",
  "rcgen",
- "reqwest",
+ "reqwest 0.12.28",
  "rustls 0.23.38",
  "serde",
  "serde_json",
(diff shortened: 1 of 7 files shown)


moonming:
## Independent audit resolved (CLAUDE.md §8 merge gate)

A fresh audit agent reviewed this PR cold and returned **BLOCK** with 2 HIGH + 2 MEDIUM + LOWs. All addressed in `6dfa19e`:

|---|-----|---------|-----------|
| 1 | HIGH | **Client was HTTPS-incapable** — `transport-streamable-http-client-reqwest` wires reqwest with no TLS backend; rustls is gated behind rmcp's separate `reqwest` feature. Every `https://` upstream would fail; the `http://127.0.0.1` test masked it. | Enabled the `reqwest` feature → `reqwest?/rustls`. Verified `rustls`/`hyper-rustls` now resolve in the crate's dep tree. |
| 2 | HIGH | **No timeouts anywhere** — rmcp's high-level client + reqwest both default to no request timeout; a hung upstream pins the gateway task. | Added a per-operation deadline (`DEFAULT_UPSTREAM_TIMEOUT` = 30s, `McpUpstream::with_timeout`) around connect / list / call. New test `upstream_call_times_out_instead_of_hanging` proves the deadline fires before a 2s server sleep. |
| 3 | MED | **`Debug` leaked the Bearer** — derived `Debug` on `McpAuth`/`McpUpstream` prints the token in plaintext. | Hand-written redacting `Debug` (`Bearer(***redacted***)`). |
| 4 | MED | **`structured_content` dropped** — a tool may return only structured content with empty `content`. | Added `structured_content: Option<Value>` to `McpToolResult`. |
| 5 | LOW | Loose `content.to_string().contains(...)` assertion. | Tightened to assert the decoded text block equals the input. |
| 6 | LOW | Dropped tool `annotations`/`output_schema`. | Noted in the DTO doc as a **DP-4** follow-up (the ACL/guardrail layer needs them). |
| — | LOW | Upstream-URL bleed into error strings once surfaced to the agent. | Flagged for **DP-3**, when these errors first reach the `/mcp` endpoint and the client-visible envelope is built. |

**Justified gap (not blocking):** a positive end-to-end *HTTPS* round-trip test isn't feasible through DP-2's public surface — `RmcpBridge::connect` uses rmcp's default reqwest client, which validates certs, so a self-signed test server can't be trusted without exposing a custom-rooted/danger client (out of scope). TLS is now compiled in (verified in the lock); a real-upstream HTTPS leg lands with **DP-3**'s e2e.

Audit-verified non-issues: no-token-passthrough holds by construction (no downstream surface in this crate); `aisix-mcp` is not yet a dependency of any binary, so the gateway doesn't link rmcp/reqwest-0.13 today; client/server feature split is correct (server is dev-only).

fmt + `clippy --workspace --all-targets -D warnings` + `cargo test -p aisix-mcp` (3/3) green.

PR #661, opened 2026-06-29
feat(mcp): downstream /mcp aggregating endpoint (DP-3)

moonming:
## What

Second step of the **MCP gateway** ([AISIX-Cloud#894](https://github.com/api7/AISIX-Cloud/issues/894) → DP-3). Adds `McpGateway` — the **downstream-facing** half of the dual role. With DP-2 (the upstream `McpBridge`), the runtime is now complete end-to-end: AISIX is an MCP **client** to each upstream **and** an MCP **server** to the agent.

```
downstream agent ──► McpGateway (/mcp) ──► upstream "alpha"
                                      └──► upstream "beta"
```

- **`tools/list`** — fans out concurrently across upstreams, returns one aggregated list, each tool namespaced `server__tool`. A failing upstream is skipped (logged), so one bad upstream doesn't blind the agent to the rest.
- **`tools/call`** — strips the prefix, routes to the owning upstream. Unknown server / prefixless names are rejected; an upstream **tool-level error** is propagated faithfully as `Ok(error_result)`, not collapsed into a protocol error.
- **Stateless transport** — served via `streamable_http_service` with `stateful_mode=false` + `json_response=true`. The aggregator holds no per-session state, so governance never depends on a session — matching the MCP `2026-07-28` direction (session header removed). Verified a real downstream rmcp client works against it.
- **No error bleed** — upstream failure detail (which may carry the upstream URL, per the DP-2 audit LOW) is logged server-side, never in the client-visible error.

`rmcp`'s server half moves from dev-dependency → production dependency (the gateway is now genuinely dual-role).

## Stacking

Based on **`feat/mcp-bridge-rmcp` (DP-2, [#660](https://github.com/api7/aisix/pull/660))**, not `main` — so this diff shows only the DP-3 changes. Merge #660 first, then this rebases onto `main`.

## Scope / not in this PR

`McpGateway` is **not mounted on any production listener**. Wiring it behind the gateway auth (AISIX key) / per-tool ACL / quota / observability pipeline, and sourcing upstreams from the resource snapshot, is **DP-4**. So there is no unauthenticated production endpoint introduced here.

## Test plan

Integration test (`tests/gateway_aggregation.rs`), all real Streamable HTTP, no mock transport: two real labeled upstream echo servers behind the gateway, driven by a **real downstream rmcp client**.

- [x] aggregated, namespaced `tools/list` (`alpha__echo`, `beta__echo`).
- [x] `tools/call` routes to the correct upstream (proven by each upstream's label prefix in the reply).
- [x] unknown server (`ghost__echo`) and prefixless name (`echo`) both error.
- [x] stateless gateway config works with a real downstream client.
- [x] `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test -p aisix-mcp` (4/4) green.

Refs AISIX-Cloud#894

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -159,11 +159,13 @@ version = "0.1.0"
 dependencies = [
  "async-trait",
  "axum",
+ "futures",
  "rmcp",
  "serde",
  "serde_json",
  "thiserror 1.0.69",
  "tokio",
+ "tracing",
 ]
 
 [[package]]
--- a/crates/aisix-mcp/Cargo.toml
+++ b/crates/aisix-mcp/Cargo.toml
@@ -46,3 +46,4 @@ features = [
 [dev-dependencies]
 tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
 axum.workspace = true
+async-trait.workspace = true
--- a/crates/aisix-mcp/Cargo.toml
+++ b/crates/aisix-mcp/Cargo.toml
@@ -6,27 +6,31 @@ rust-version.workspace = true
 license.workspace = true
 repository.workspace = true
 authors.workspace = true
-description = "aisix: MCP gateway — upstream MCP server client (Streamable HTTP) behind the McpBridge trait"
+description = "aisix: MCP gateway — upstream client + downstream-facing aggregating endpoint (Streamable HTTP)"
 
 [dependencies]
 tokio.workspace = true
 async-trait.workspace = true
+futures.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 thiserror.workspace = true
+tracing.workspace = true
 
 # Official MCP Rust SDK. Pinned exactly: rmcp is <16 months old and still
 # ships breaking changes on a roughly-monthly cadence, so we hold a fixed
-# version and keep all rmcp types behind this crate's `McpBridge` trait to
-# absorb API drift. Production only needs the upstream-facing client; the
-# server side (used to stand up a real MCP server in tests, and later for the
-# DP `/mcp` endpoint) is a dev-dependency so the gateway binary does not link
-# it yet.
+# version and keep all rmcp types behind this crate's own surface (the
+# `McpBridge` trait and the gateway DTOs) to absorb API drift.
+#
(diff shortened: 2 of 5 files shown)


moonming:
## Independent audit resolved (CLAUDE.md §8 merge gate)

Audit returned **BLOCK** with 2 MEDIUM. Both addressed in `701a84c`:

|---|-----|---------|-----------|
| 1 | MED | **Untested contract clauses** — graceful-skip of a failing upstream in `tools/list`, and tool-level-error propagation as `Ok(error_result)` (not a protocol `Err`) — the two most regression-prone paths, with zero coverage. | Added 3 tests: `skips_failing_upstream_keeping_the_rest` (a `FailingBridge` stub registered alongside a healthy upstream → list still returns only the healthy tool), `propagates_tool_level_error_as_ok` (real upstream returns a tool-level error → downstream sees `Ok` with `is_error=Some(true)`), and `duplicate_upstream_name_keeps_first`. |
| 2 | MED | **Duplicate upstream names silently shadowed** — `find()` returned the first; the second was dead weight but still listed (duplicate tool names on the wire). DP-4's snapshot-sourced names could trigger it. | `McpGateway::new` now dedups by name: first registration wins, later ones dropped with a `tracing::warn!`; added a `debug_assert!` that names don't contain the separator. New test pins first-wins routing. |

**Audit-verified non-issues (no change):**
- LOW — `tracing::warn!(error = %error)` logs the rmcp error Display; verified rmcp `ServiceError` variants never echo the `Authorization` header, and DP-2 redacts the bearer in `Debug`. Only the upstream URL reaches the **server** log (operator-side); the client-visible `ErrorData` is the generic message. No credential leak.
- LOW — non-object `input_schema` coerced to `{}` is deliberate (advertise rather than drop); acknowledged in-code.
- Namespace `split_once("__")` on first occurrence is correct given the documented "server names must not contain `__`, tool names may" invariant (now also `debug_assert`-guarded).
- `aisix-mcp` is still linked by no binary, so moving rmcp's server features to prod deps bloats nothing shipped.

7/7 tests green; `fmt` + `clippy --workspace --all-targets -D warnings` clean.

PR #664, opened 2026-06-29
feat(mcp): McpServer admin resource (registration + CRUD)

moonming:
## What

Adds **`McpServer`** as a first-class Admin resource (parallel to `ProviderKey`): an upstream MCP server registration that the MCP gateway endpoint will source its upstreams from, instead of the explicit list it takes today. This is the control-plane prerequisite for wiring the gateway behind the governance pipeline.

Fields (customer-facing, rendered into the Admin API reference): `display_name` (unique; namespace prefix for the server's tools, so `__` is rejected), `url` (Streamable HTTP endpoint), `transport` (enum, `streamable_http`), `auth_type` (`none`/`bearer`) + `secret` (gateway-held), `timeout_ms`, `enabled`.

## How (mirrors the existing resource scaffolding)

- **aisix-core** — `McpServer` struct + `McpTransport`/`McpAuthType` enums + `Resource` impl (`kind = "mcp_servers"`); schema validator built through the same `struct_root_schema` producer the other resources use, so the **published schema == the enforced schema**; registered in `dump-schema`; new `mcp_servers` snapshot table.
- **aisix-etcd** — loader decode arm + supervisor merge / present-check / delete / clone dispatch for the new kind.
- **aisix-admin** — `ConfigStore` get/put/list/delete on both `InMemoryStore` and `EtcdConfigStore` (etcd subkey `mcp_servers`); `/admin/v1/mcp_servers[/:id]` handlers (schema-validate, dup-name 409, reject reserved `__`, uuid on POST, revision bump on PUT) + routes.

## Test plan

- [x] Resource unit tests (defaults, bearer auth, unknown-field/enum rejection, Resource trait, round-trip).
- [x] `mcp_servers_round_trip_through_real_etcd` — full Admin CRUD over a real etcd.
- [x] `loader_picks_up_every_admin_write` extended to seed an `mcp_servers` row and assert the loader accepts it — this catches **subkey-constant drift** between `EtcdConfigStore` and `aisix_etcd::loader` (the most likely wiring bug).
- [x] `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test -p aisix-core -p aisix-etcd -p aisix-admin` all green; `schemas/resources/mcp_server.schema.json` regenerated via `dump-schema`.

## Scope / explicitly deferred

[...]

PR #665, opened 2026-06-29
feat(mcp): source gateway upstreams from the snapshot

moonming:
## What

Connects the `McpServer` resource (merged in #664) to the MCP gateway endpoint (#662): `McpGateway::from_snapshot(&AisixSnapshot)` builds the aggregator from the **enabled** `mcp_servers` in the snapshot, instead of the explicit `(name, bridge)` list it took before. This is the first slice of wiring the gateway into the data plane.

## How

- **`EphemeralBridge`** — an `McpBridge` that opens a fresh upstream session per operation and drops it. The `/mcp` endpoint is already configured stateless, so the gateway holds **no long-lived upstream connections**: every `tools/list` / `tools/call` connects → runs → disconnects. A config change is therefore picked up on the next request with nothing to reconcile. (Connection pooling is a deliberate later optimization, not needed for correctness.)
- **`upstream_from_mcp_server`** — maps a registered server's `url` / `auth_type` / `secret` / `timeout_ms` onto the bridge connection params (`McpAuthType::Bearer` + `secret` → `McpAuth::Bearer`; `timeout_ms` → per-op deadline; default otherwise).
- `aisix-mcp` now depends on `aisix-core` (one direction; no cycle).

## Test plan

- [x] `upstream_from_mcp_server_maps_auth_and_timeout` — bearer/secret/timeout mapping + `none`/absent defaults.
- [x] `from_snapshot_sources_only_enabled_upstreams` — end-to-end against **two real** upstreams: the **disabled** one is skipped from `tools/list`, the enabled one lists + routes correctly (ephemeral connect per call), driven by a real downstream rmcp client.
- [x] `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test -p aisix-mcp` (6 + 3) green.

## Scope / next slices

Not in this PR (each is a separate audited slice):
- **Mount** `/mcp` in the proxy/server router behind **AISIX-key authN** (reuse `AuthenticatedKey`).
- **Per-tool ACL** (key/team `allowed_tools` filtering — needs the key identity plumbed into the handler).
- **Guardrail / rate-limit / budget reuse** over MCP traffic.
- **UsageEvent** emission for MCP calls.

Refs AISIX-Cloud#894

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -157,6 +157,7 @@ dependencies = [
 name = "aisix-mcp"
 version = "0.1.0"
 dependencies = [
+ "aisix-core",
  "async-trait",
  "axum",
  "futures",
--- a/crates/aisix-mcp/Cargo.toml
+++ b/crates/aisix-mcp/Cargo.toml
@@ -9,6 +9,7 @@ authors.workspace = true
 description = "aisix: MCP gateway — upstream client + downstream-facing aggregating endpoint (Streamable HTTP)"
 
 [dependencies]
+aisix-core = { path = "../aisix-core" }
 tokio.workspace = true
 async-trait.workspace = true
 futures.workspace = true
@@ -44,6 +45,7 @@ features = [
 ]
 
 [dev-dependencies]
+aisix-core = { path = "../aisix-core" }
 tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
 axum.workspace = true
 async-trait.workspace = true
--- a/crates/aisix-mcp/src/bridge.rs
+++ b/crates/aisix-mcp/src/bridge.rs
@@ -13,6 +13,7 @@
 
 use std::time::Duration;
 
+use aisix_core::{McpAuthType, McpServer};
 use async_trait::async_trait;
 use rmcp::model::CallToolRequestParams;
 use rmcp::service::{RoleClient, RunningService};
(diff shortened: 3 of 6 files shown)


moonming:

Auditor built, clippy'd (`-D warnings`), and ran the full `aisix-mcp` suite — all green. All three contract claims verified: `from_snapshot` filters on `enabled` and preserves `new`'s dedup/`__`-guard; `EphemeralBridge` is genuinely stateless (session drops per call); the auth/timeout mapping is correct and pinned by tests.

Three LOW findings, none blocking:
| Sev | Finding | Disposition |
|-----|---------|-------------|
| LOW | `timeout_ms: 0` → instant timeout (upstream silently dropped) | **Pre-existing** Admin-validation gap, not introduced here; the mapping faithfully reflects stored config. Filed as **#666** (schema `range(min=1)` fix). |
| LOW | Tool-list ordering nondeterministic (DashMap iteration) | Cosmetic; doc comment is honest about it; correctness unaffected (Admin enforces name uniqueness). No fix. |
| LOW | Bearer through `from_snapshot` not e2e'd | Both halves covered separately (mapping unit test + `forwards_gateway_held_bearer_to_upstream`). Acceptable for this slice. |

Non-issues confirmed: no dep cycle (`aisix-core` has no `aisix-mcp` edge), no binary bloat (aisix-mcp linked by nobody yet), no secret leak (redacted `Debug` untouched, `EphemeralBridge` doesn't derive Debug).

Merging — all required CI green.

PR #667, opened 2026-06-29
feat(mcp): mount the /mcp gateway endpoint behind AISIX-key auth

moonming:
## What

Mounts the downstream-facing MCP gateway in the proxy router at **`/mcp`** (DP-4 slice 2). This is the first time the `/mcp` endpoint goes live in the gateway binary — authenticated and sourced from the live `mcp_servers` config.

## How

- Handler `mcp::mcp_endpoint` takes the existing **`AuthenticatedKey`** extractor — a missing/invalid AISIX API key is rejected with **401** (the proxy's standard `invalid_api_key` envelope) before the request reaches the gateway. AuthN only this slice; per-tool ACL is the next one.
- Builds an `McpGateway::from_snapshot(&snapshot)` **per request**, so the endpoint always reflects the live `mcp_servers` set with nothing to reconcile (the gateway holds no connections — `EphemeralBridge` connects lazily).
- `StreamableHttpService` dispatches on **method, not path**, and is `Error = Infallible`, so the handler calls it directly (`oneshot`) and maps its boxed body back to axum's — no `nest_service` prefix-stripping needed.
- `/mcp` added to the metric endpoint-label allowlist (`inbound_protocol = "mcp"`).
- `aisix-proxy` now depends on `aisix-mcp` (one direction, no cycle); the gateway binary links the MCP stack (rmcp + its reqwest) for the first time.

## Test plan

Router-level tests (`oneshot` against `build_router`):
- [x] no API key → **401**.
- [x] wrong API key → **401**.
- [x] valid key + MCP `initialize` → **200**, body carries `serverInfo` (auth passed and the gateway served the handshake).
- [x] `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test -p aisix-proxy` (504 pass, no regression), `cargo build -p aisix-server --bin aisix` all green.

## Scope / next slices

Not in this PR: per-tool **ACL** (filter tools by the key's allowed set — needs the resolved key plumbed into the gateway handler), **guardrail / rate-limit / budget** reuse over MCP traffic, and **UsageEvent** emission for MCP calls.

Refs AISIX-Cloud#894

## Summary by CodeRabbit

* **New Features**
  * Added a new `/mcp` endpoint in the proxy for MCP gateway traffic.
  * Requests now route consistently for both `/mcp` and `/mcp/`, with matching request labeling and protocol handling.

* **Bug Fixes**
  * Improved request handling for the gateway so invalid or missing API keys are rejected.
  * Large unauthenticated requests are blocked early by the request-size limit.
  * The gateway now accepts any `Host` header for this endpoint, while access controls remain enforced elsewhere.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -309,6 +309,7 @@ dependencies = [
  "aisix-core",
  "aisix-gateway",
  "aisix-guardrails",
+ "aisix-mcp",
  "aisix-obs",
  "aisix-provider-anthropic",
  "aisix-provider-openai",
--- a/crates/aisix-mcp/src/gateway.rs
+++ b/crates/aisix-mcp/src/gateway.rs
@@ -221,6 +221,15 @@ pub fn streamable_http_service(
     let mut config = StreamableHttpServerConfig::default();
     config.stateful_mode = false;
     config.json_response = true;
+    // Disable rmcp's `Host`-header allowlist. Its default
+    // (`localhost`/`127.0.0.1`/`::1`) is a DNS-rebinding guard for
+    // browser-driven local servers — it 403s every request whose `Host` is
+    // the deployment's real DNS name. This endpoint is not browser-driven: it
+    // is reached server-to-server by agents and is gated by the AISIX API key,
+    // which is the real access control. An empty allowlist accepts any `Host`;
+    // the request is still authenticated upstream of this service. (Operators
+    // who want Host pinning can layer it at their ingress.)
+    config.allowed_hosts = Vec::new();
     StreamableHttpService::new(
         move || Ok(gateway.clone()),
         Arc::new(LocalSessionManager::default()),
--- a/crates/aisix-proxy/Cargo.toml
+++ b/crates/aisix-proxy/Cargo.toml
@@ -11,6 +11,9 @@ description = "aisix: /v1/* proxy router (OpenAI-compatible) + middleware + hook
 [dependencies]
 aisix-core = { path = "../aisix-core" }
 aisix-gateway = { path = "../aisix-gateway" }
+# Mounts the downstream-facing `/mcp` MCP gateway endpoint (mcp.rs): the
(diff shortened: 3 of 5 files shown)


moonming:
## Independent audit (CLAUDE.md §8): BLOCK → resolved

Auditor checked out the branch, ran build/test/clippy, and wrote adversarial probes. **Auth enforcement verified solid**: no-key GET/DELETE/OPTIONS/`POST /mcp/` all 401; the request only reaches rmcp after auth passes; oversized unauthenticated body → 413 before auth (no resource pin); no dep cycle; no info leak on 401.

One **MEDIUM** (deployment-blocking) — fixed in `b78b084`:

| Sev | Finding | Resolution |
|-----|---------|-----------|
| MED | rmcp's `StreamableHttpServerConfig::default()` sets `allowed_hosts = [localhost/127.0.0.1/::1]` (DNS-rebinding guard) → the `/mcp` endpoint **403'd every non-loopback `Host`**, i.e. any real client connecting via the deployment's DNS name, even after valid auth. Tests passed only because they sent `Host: localhost`. | `streamable_http_service` now clears the Host allowlist. DNS-rebinding host validation is a **browser** threat model; this endpoint is server-to-server and gated by the AISIX API key (the real access control). The authed test now uses a **non-loopback** Host (`mcp.aisix.example.com`) to prove it. |

LOW (coverage) — folded in: added tests that non-POST methods are auth-gated, `/mcp/` is auth-gated, and the oversized-body 413 path fires before auth.

LOW (informational): the binary now links rmcp + a second reqwest (0.13 alongside 0.12) / two TLS stacks — noted, acceptable, pre-existing in aisix-mcp.

507 aisix-proxy tests + aisix-mcp suite green; `clippy --workspace --all-targets -D warnings` clean; `cargo build -p aisix-server` builds.

Issue #663, opened 2026-06-29
Document /admin/v1/mcp_servers in the Admin API OpenAPI


moonming:
The `mcp_servers` Admin resource (CRUD at `/admin/v1/mcp_servers` and `/admin/v1/mcp_servers/{id}`) ships its handlers, store, schema validation, and etcd round-trip, but its routes are not yet added to the hand-written OpenAPI in `crates/aisix-admin/src/openapi.rs`.

Follow-up: add the two path blocks (mirroring the `provider_keys` block), register the `McpServer` + `McpServerEntry` component schemas, and add the paths/schemas to the two coverage tests (`openapi_json_is_well_formed_and_documents_admin_paths`, `openapi_documents_exact_admin_path_set`).

Deferred from the resource PR to keep that change focused; no functionality depends on it (the routes work; only the generated OpenAPI reference omits them). Refs AISIX-Cloud#894.

moonming:
Done in #675, which added the `/admin/v1/mcp_servers` + `/admin/v1/mcp_servers/{id}` path blocks, the `McpServer`/`McpServerEntry` component schemas, the `McpAuthType`/`McpTransport` ReDoc variant titles, and extended the path-coverage tests. Verified on current main: the merged OpenAPI documents both paths and all four schemas, and the 19 openapi tests pass. (#675 didn't carry a close keyword, so closing here.)

PR #668, opened 2026-06-29
feat(mcp): per-tool ACL — govern MCP tools by the caller's API key

moonming:
## What

DP-4 slice 3. Scopes the MCP gateway to the tools the caller's AISIX API key permits — so MCP tool access is governed by the **same key object** as LLM access (the "single-pipeline same-governance" differentiation). Builds on the mounted, authenticated `/mcp` endpoint (#667) and the snapshot-sourced gateway (#665).

## How

- **aisix-core** — `ApiKey.allowed_tools: Option<Vec<String>>` (namespaced `<server>__<tool>` names; `"*"` = all) + `can_access_tool`, mirroring `allowed_models` / `can_access`. **Deny-by-default**: a key with no `allowed_tools` may call no MCP tools — access is granted explicitly (consistent with `allowed_models`).
- **aisix-mcp** — `ToolAcl` (`AllowAll` / `Allow(set)`, built from a key's `allowed_tools`) + `McpGateway::with_tool_acl`. `tools/list` exposes only permitted tools; `tools/call` rejects the rest with a **neutral** message (defense-in-depth — doesn't reveal whether the tool exists upstream).
- **aisix-proxy** — the `/mcp` handler builds the gateway scoped to `auth.key().allowed_tools` (per request; the handler already holds the resolved key, so no rmcp request-context plumbing is needed).
- **aisix-admin** — the self-hosted apikeys API (`StandaloneApiKeyBody` + `PublicApiKey`) round-trips `allowed_tools`; `api_key.schema.json` regenerated (so the loader + admin validation accept it).

## Test plan

- [x] `can_access_tool` semantics (none/empty → deny, `["*"]` → all, exact names).
- [x] `ToolAcl::from_allowed` mapping.
- [x] **end-to-end** (real upstreams): a key permitting only `alpha__echo` → `beta__echo` is hidden from `tools/list` and its call is rejected; `alpha__echo` lists and calls. (`tool_acl_filters_list_and_rejects_calls`)
- [x] apikeys etcd round-trip carries `allowed_tools`.
- [x] `fmt`, `clippy --workspace --all-targets -D warnings`, `cargo test` (core/mcp/proxy/admin) green.

## Scope notes

- ACL behavior lives in `aisix-mcp` and is e2e-tested there (the real-upstream harness); the proxy is one-line wiring (`ToolAcl::from_allowed(key.allowed_tools)`), covered by compile + the existing `/mcp` auth tests.
- OpenAPI doc for the new `allowed_tools` request/response field is not added (consistent with how the resource OpenAPI was handled in #663/#664; the field IS accepted by the admin API via the regenerated schema). Can fold into the #663 OpenAPI follow-up.
- Remaining DP-4 slices: ④ guardrail / rate-limit / budget reuse over MCP; ⑤ UsageEvent for MCP calls.

Refs AISIX-Cloud#894

### DIFF
--- a/crates/aisix-admin/src/apikeys_handlers.rs
+++ b/crates/aisix-admin/src/apikeys_handlers.rs
@@ -32,6 +32,10 @@ struct StandaloneApiKeyBody {
     #[serde(default)]
     #[serde(skip_serializing_if = "Option::is_none")]
     rate_limit: Option<aisix_core::models::RateLimit>,
+    /// MCP tools this key may call (namespaced `<server>__<tool>`, or `"*"`).
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    allowed_tools: Option<Vec<String>>,
 }
 
 #[derive(Debug, Clone, Serialize)]
@@ -40,6 +44,8 @@ pub struct PublicApiKey {
     pub allowed_models: Vec<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub rate_limit: Option<aisix_core::models::RateLimit>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub allowed_tools: Option<Vec<String>>,
 }
 
 impl From<ApiKey> for PublicApiKey {
@@ -48,6 +54,7 @@ impl From<ApiKey> for PublicApiKey {
             key_hash: value.key_hash,
             allowed_models: value.allowed_models,
             rate_limit: value.rate_limit,
+            allowed_tools: value.allowed_tools,
         }
     }
 }
--- a/crates/aisix-admin/tests/etcd_integration.rs
+++ b/crates/aisix-admin/tests/etcd_integration.rs
@@ -175,7 +175,11 @@ async fn apikeys_round_trip_through_real_etcd() {
     admin_crud_round_trip(
         state,
         "/admin/v1/apikeys",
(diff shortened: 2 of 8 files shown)


moonming:

Auditor ran clippy/test/build/dump-schema in an isolated worktree — all green, **zero schema drift**. Verified: deny-by-default holds; ACL filters `tools/list` on namespaced names and gates `tools/call` **before** routing to any upstream (no upstream contact for a disallowed call); neutral rejection (no existence leak); no bypass (`__`-in-name, `*`-literal vs AllowAll, case/trailing — all deny, never looser); no regression to LLM auth/routing; admin round-trips `allowed_tools`.

Five LOW findings, none blocking. Two folded in (`commit pushed`):
- **Neutral-message contract was untested** → e2e now asserts the rejection says "not available" and does not contain "permitted"/"forbidden"/"exists".
- **`McpGateway::new` AllowAll default could fail-open** for a future caller → documented that `new`/`from_snapshot` are unrestricted until `with_tool_acl`, and that the proxy mount is the single enforcement point (always scopes).

Acknowledged deferrals (LOW): OpenAPI doc for `allowed_tools` → folds into #663; a proxy-level ACL e2e (logic is e2e-tested at the aisix-mcp layer; proxy is one-line wiring); asserting the stored `allowed_tools` value (rides the same serde path as `allowed_models`, acceptance is tested).

Merging on CI green.

PR #669, opened 2026-06-30
feat(mcp): apply rate-limit + budget governance to MCP tool calls

moonming:
## What

DP-4 slice 4 (governance reuse). An MCP `tools/call` now passes through the **same rate-limit and budget gate** as an LLM request — `quota::enforce(&state, &auth, None)`, keyed on the caller's API key. This is the core of the "single-pipeline same-governance" story: the same key, the same limiter, the same budget govern LLM *and* MCP work.

## How

- **Gate placement** — in the `/mcp` handler, after auth, before the gateway. `quota::enforce` applies the key's inline `rate_limit` (RPM/TPM/concurrency) **and** the budget pre-check (`budgets.check`). On 429 / budget-exceeded it returns before any upstream is contacted.
- **Scoped to the unit of work** — the handler peeks at the JSON-RPC `method` and only meters `tools/call`. `initialize` / `tools/list` (handshake + discovery) pass through ungated, so a client can always connect and enumerate even when the key is at its tool-call limit. (Reading the method means buffering the body — already size-capped by the global body-limit layer — and rebuilding the request for the gateway.)
- **No token cost** — a tool call carries no token figure (per-tool cost is a separate future concern), so the reservation is dropped after the call without committing tokens, which releases the concurrency slot. Budget is a gating check here, not a charge.

## Test plan

- [x] With an `rpm=1` key: first `tools/call` passes the gate (non-429), second `tools/call` within the minute → **429**, and `initialize` is **never** rate-limited (proves handshake/discovery bypass).
- [x] `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test -p aisix-proxy` (508 pass), green.

## Scope / remaining

- This delivers **rate-limit + budget** reuse over MCP. The last slice — **guardrails over MCP** — is separate and meatier (guardrails are built around chat text, so MCP tool-call *arguments* and *results* need a JSON↔text adapter to flow through `guardrail_index.resolve` + `check_input`/`check_output`). And **UsageEvent** emission for MCP calls.

Refs AISIX-Cloud#894

## Summary by CodeRabbit

* **Bug Fixes**
  * Added rate-limit enforcement to MCP tool calls, helping prevent excessive usage before requests reach the backend.
  * MCP connection setup now respects existing request limits while leaving initial handshake and discovery steps unaffected.
  * Improved handling of incoming MCP requests so they continue to work reliably after being inspected for routing and limits.

### DIFF
--- a/crates/aisix-proxy/src/mcp.rs
+++ b/crates/aisix-proxy/src/mcp.rs
@@ -170,13 +170,13 @@ mod tests {
         snapshot
     }
 
-    /// A `tools/call` request authenticated with `TOKEN`.
-    fn tools_call_request() -> HttpRequest<Body> {
+    /// A JSON-RPC request to `/mcp` for `method`, authenticated with `TOKEN`.
+    fn mcp_request(method: &str, params: serde_json::Value) -> HttpRequest<Body> {
         let body = serde_json::json!({
             "jsonrpc": "2.0",
             "id": 1,
-            "method": "tools/call",
-            "params": { "name": "ghost__tool", "arguments": {} }
+            "method": method,
+            "params": params
         });
         HttpRequest::post("/mcp")
             .header("host", "mcp.aisix.example.com")
@@ -185,6 +185,13 @@ mod tests {
             .header("authorization", format!("Bearer {TOKEN}"))
             .body(Body::from(body.to_string()))
             .unwrap()
+    }
+
+    fn tools_call_request() -> HttpRequest<Body> {
+        mcp_request(
+            "tools/call",
+            serde_json::json!({ "name": "ghost__tool", "arguments": {} }),
+        )
     }
 
     #[tokio::test]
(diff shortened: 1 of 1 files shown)


moonming:

Auditor verified cold against source (incl. rmcp 1.8.0) + ran build/test/clippy in an isolated worktree — all green. Key confirmations:
- **No rate-limit bypass.** rmcp dispatches to `call_tool` *only* on an exact-byte `method == "tools/call"` (untagged enum), and `JsonRpcMessage` has **no batch/array variant** — so a malformed/non-JSON/batch/`tools/list`/`initialize` body is `is_tool_call=false` *and* can never execute a tool. "Ungate everything but tools/call" is safe by construction; the handler's peek uses the same discriminator as rmcp's dispatch.
- **Reservation lifetime correct.** Drop releases only the concurrency slot (RPM increment persists → 2nd call 429), charges zero tokens, runs even on unwind. No leak.
- **`enforce(None)` enforces the key's own rate_limit + budget** (not a no-op); test budget is `disabled → allow_all` so no false-positive 429; 429 envelope leaks no key/internal.
- **DoS-safe**: body already capped by the global limit layer before `to_bytes`.

Two LOW findings; one folded in (`commit pushed`):
- Coverage: `tools/list`-at-the-limit not tested → **added** (now both `initialize` and `tools/list` are pinned as ungated).
- Latency note (every `/mcp` request now buffered): negligible for small JSON-RPC, LLM endpoints buffer identically — no change.

Merging on CI green.

PR #670, opened 2026-06-30
feat(mcp): emit usage events for MCP tool calls

moonming:
## What

DP-4 slice ⑤b (observability). Each MCP `tools/call` now emits a **UsageEvent into the same sink/pipeline as LLM usage**, so MCP activity lands in the unified usage record — matching how the leading gateways log MCP calls (Kong: MCP audit log + metrics; Portkey: unified MCP+LLM traces), framed AISIX's way: *the same usage pipeline*, not a separate MCP surface.

## How

- The `/mcp` handler peeks the JSON-RPC `params.name` (it already buffers the body for rate-limit), splits the namespaced `<server>__<tool>`, and after the gateway call emits a `UsageEvent` via `state.usage_sink.try_emit("mcp", ..)` (non-blocking, fire-and-forget — same as LLM handlers).
- Event fields: `api_key_id`, `status_code`, `latency_ms`, `inbound_protocol="mcp"`, and two new `mcp_server_name` / `mcp_tool_name`. **Rate-limited calls are recorded too** (the 429 path emits before returning). `initialize` / `tools/list` are **not** metered.
- **No token cost**: MCP tool calls carry no token figure yet (per-tool cost is deferred, #894 Phase 3), so token/cost fields stay zero — consistent with Kong/Portkey, who also log MCP calls without a per-call cost.
- `UsageEvent` gains two optional, skip-when-empty fields. Non-MCP events leave them empty (the chat path now uses `..Default::default()`). **DP-first rollout** like `user_name`: older cp-api ignores the unknown fields until the CP migration lands — filed as **[AISIX-Cloud#908](https://github.com/api7/AISIX-Cloud/issues/908)**.

## Reference implementations (CLAUDE.md §7)

Re-verified against primary sources (2026-06-30): **all three** mainstream gateways log MCP calls — LiteLLM (same spend-log table, `mcp_namespaced_tool_name`, `call_type=call_mcp_tool`), Kong (MCP audit log + Prometheus metrics, 3.12+), Portkey (unified MCP+LLM logs/traces). So MCP usage logging is table-stakes; AISIX's framing is unifying it into the *same* `UsageEvent` as LLM. **Per-tool cost**: only LiteLLM has it — we defer, like Kong/Portkey.

## Test plan

- [x] `emits_usage_event_for_tool_call_only`: a `tools/call` emits exactly one event with `inbound_protocol="mcp"`, `mcp_server_name`/`mcp_tool_name` parsed from the namespaced name, the right `api_key_id`, and zero tokens; `initialize` emits **none**. (Injects a real `UsageSink` with a receiver and asserts on the channel.)
- [x] `fmt`, `clippy --workspace --all-targets -D warnings`, `cargo test -p aisix-obs -p aisix-proxy` green.

## Remaining

Slice ⑤a — **guardrails over MCP** (the differentiator: Kong won't, Portkey "coming soon", LiteLLM experimental+buggy) — is the next and final DP-4 PR.

Refs AISIX-Cloud#894

## Summary by CodeRabbit

* **New Features**
  * MCP tool calls are now tracked as usage events, including server/tool attribution and request latency.
  * MCP-related usage is labeled distinctly, improving reporting and visibility.

* **Bug Fixes**
  * MCP handshake/discovery traffic no longer appears as billable tool usage.
  * Usage events now include default values consistently, preventing missing data in non-MCP chat flows.

### DIFF
--- a/crates/aisix-obs/src/usage.rs
+++ b/crates/aisix-obs/src/usage.rs
@@ -313,6 +313,19 @@ pub struct UsageEvent {
     /// Empty when the client sent none. cp-api stores empty as NULL.
     #[serde(default, skip_serializing_if = "String::is_empty")]
     pub client_user_agent: String,
+
+    // ─── MCP gateway attribution ───
+    /// Registered name of the upstream MCP server a `tools/call` was routed
+    /// to (the namespace prefix of the requested tool). Empty for non-MCP
+    /// events; cp-api stores empty as NULL. Older cp-api images that predate
+    /// this field ignore it (DP-first rollout).
+    #[serde(default, skip_serializing_if = "String::is_empty")]
+    pub mcp_server_name: String,
+
+    /// Name of the MCP tool invoked, without the server prefix. Empty for
+    /// non-MCP events; cp-api stores empty as NULL.
+    #[serde(default, skip_serializing_if = "String::is_empty")]
+    pub mcp_tool_name: String,
 }
 
 #[inline]
@@ -405,6 +418,7 @@ impl UsageSink {
         let bounded_protocol: &'static str = match event.inbound_protocol.as_str() {
             "openai" => "openai",
             "anthropic" => "anthropic",
+            "mcp" => "mcp",
             _ => "other",
         };
 
--- a/crates/aisix-proxy/src/chat.rs
+++ b/crates/aisix-proxy/src/chat.rs
@@ -2875,6 +2875,8 @@ fn emit_usage_event(
         // extractor (control-char strip + cap); IP is a formatted addr.
         client_source_ip: client.source_ip.clone(),
         client_user_agent: client.user_agent.clone(),
+        // MCP attribution does not apply to the chat path.
+        ..Default::default()
     };
     // Handler label "chat" matches the documented enumeration for
     // `aisix_usage_events_emitted_total` (#408). Keep `&'static str`
--- a/crates/aisix-proxy/src/mcp.rs
+++ b/crates/aisix-proxy/src/mcp.rs
@@ -12,6 +12,9 @@
 //! traffic are layered on in subsequent steps; this step establishes the
 //! authenticated, snapshot-sourced endpoint.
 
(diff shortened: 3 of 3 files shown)


moonming:
## Independent audit (CLAUDE.md §8): BLOCK → resolved

Auditor built/tested/clippy'd in an isolated worktree — all green (509 tests). Verified: **exactly one** emit per tool call (429 path emits then returns → post-gateway emit unreachable; success path emits once; initialize/tools-list emit zero); `latency_ms` spans only the gateway call; attribution split uses the **same** `split_once("__")` as the gateway's routing (so it matches exactly); **no tool arguments/results/secrets** reach the event (only the tool *name*); new fields are **append-only** with the same lenient `skip_serializing_if` pattern as the other optional `UsageEvent` fields → CP-safe (CP migration tracked AISIX-Cloud#908).

One **MEDIUM** — fixed in the pushed commit:
| Sev | Finding | Resolution |
|-----|---------|-----------|
| MED | "rate-limited calls are recorded" had emit code but **no test** (existing 429 test used a disabled sink) | Added `rate_limited_tool_call_still_emits_usage_event` (rpm=1 → 2nd call 429 → asserts a UsageEvent with `status_code=429` + mcp attribution); plus an "exactly one event" assert on the happy path. |

Non-blocking:
- **LOW** — a JSON-RPC tool error logs `status_code=200` (rmcp returns HTTP 200 for tool-level errors). Acceptable for v1; refining to surface tool-error as a distinct status/`error_class` is a follow-up for the cost/observability pass.
- **Cosmetic** — the PR prose cited `user_name` as the DP-first precedent; `UsageEvent` has no such field. The real precedents are the existing `skip_serializing_if` string fields (e.g. `client_user_agent`), whose comments state cp-api binds `/dp/telemetry` JSON leniently. The pattern claim holds; only the cited name was off.

`clippy --workspace --all-targets -D warnings` + `cargo test -p aisix-obs -p aisix-proxy` green. Merging on CI green.

Issue #671, opened 2026-06-30
Output guardrails over MCP: scan tool-call results


moonming:
DP-4 slice ⑤a (api7/aisix#671) added **input** guardrails over MCP — tool *arguments* run through the same guardrail chain as LLM input (`check_input`), blocking with a JSON-RPC error before the upstream is reached.

The **output** side (scanning the tool *result* with `check_output` — catching secrets/PII a tool returns) is not yet done. It needs the `/mcp` handler to buffer the gateway *response* body, extract the JSON-RPC `result`, synthesize a `ChatResponse` (`ChatMessage::assistant(result_text)`), run `Guardrail::check_output(&chain, &resp)`, and on Block replace the result with a JSON-RPC error — mirroring the input path. Reuse the same env/key/team-scoped chain already resolved in the handler.

This completes guardrails-over-MCP (input + output), matching what LiteLLM does (input `pre_mcp_call` + output `async_post_mcp_tool_call_hook`) — but on the single DP pipeline. Refs AISIX-Cloud#894.

PR #672, opened 2026-06-30
feat(mcp): input guardrails over MCP tool calls

moonming:
## What

DP-4 slice ⑤a — the **differentiating** slice. An MCP `tools/call` now runs its **arguments** through the **same guardrail chain** as LLM input, blocking a policy-violating call before it reaches the upstream. This completes the "single-pipeline same-governance" story: one key, one ACL, one rate-limiter/budget, one guardrail chain govern LLM **and** MCP.

## How

- After auth + rate-limit, the `/mcp` handler resolves guardrails for the call: `RequestContext { model_id: "", api_key_id, team_id }`. **MCP has no model**, so an empty `model_id` matches env / api-key / team-scoped guardrails and never a Model-scoped one (verified in `applies_to`). An **empty chain short-circuits**, so the no-guardrail path stays cheap.
- The tool arguments are wrapped in a `ChatFormat` (`ChatMessage::user(args_json)`) and run through `Guardrail::check_input` — the exact call the LLM path uses.
- On `Block`, the tool **never reaches the gateway/upstream**; the caller gets an **MCP-native JSON-RPC error** (`{"error":{...}}`, HTTP 200, echoing the request id). This diverges from the LLM path's HTTP 422 on purpose — an MCP client expects a JSON-RPC envelope, not an HTTP error, so the block surfaces as a tool-call error it can handle. The block is recorded in the usage event (`guardrail_blocked=true`).

## Reference implementations (CLAUDE.md §7)

Re-verified against primary sources (2026-06-30): **guardrails over MCP is where AISIX leads.** Kong's `ai-mcp-proxy` lists "Applying guardrails to MCP" under **Not Supported** (and forbids co-locating AI plugins on an MCP route); another mainstream gateway's MCP-guardrails page says **"coming soon"**; the one that ships it keeps it under `_experimental/` with open bugs (hooks bypassed for some tools, result-redaction incomplete, a crash). AISIX runs MCP tool calls through the **production** guardrail path on the single DP pipeline. Divergence: we return a JSON-RPC error (MCP-native) rather than the LLM 422.

## Scope

**Input-side** (tool arguments) this PR. **Output-side** (scanning the tool *result* with `check_output`) needs response-body buffering and is filed as a tracked follow-up: **#671**.

## Test plan

- [x] `input_guardrail_blocks_tool_call_with_forbidden_args`: a `tools/call` whose arguments contain a keyword-guardrail-blocked token → JSON-RPC content-policy error (HTTP 200), before the gateway; clean arguments are not blocked. (Seeds an env-scoped keyword guardrail into the live snapshot.)
- [x] `fmt`, `clippy --workspace --all-targets -D warnings`, `cargo test -p aisix-proxy` (511 pass), green.

Refs AISIX-Cloud#894

## Summary by CodeRabbit

* **New Features**
  * `/mcp` tool calls now follow the same safety checks and usage tracking as other AI requests.
  * Tool arguments are scanned for policy violations, and blocked calls return a JSON-RPC error response.

* **Bug Fixes**
  * Improved handling of tool-call request details so the request ID, method, tool name, and arguments are consistently captured.
  * Usage records now clearly indicate when a tool call was blocked by a guardrail.

### DIFF
--- a/crates/aisix-proxy/src/mcp.rs
+++ b/crates/aisix-proxy/src/mcp.rs
@@ -8,9 +8,11 @@
 //! current configuration snapshot on each request, so it always reflects the
 //! live `mcp_servers` set.
 //!
-//! Per-tool access control, guardrail / quota reuse, and usage logging over MCP
-//! traffic are layered on in subsequent steps; this step establishes the
-//! authenticated, snapshot-sourced endpoint.
+//! A `tools/call` is governed by the SAME pipeline as an LLM request, keyed on
+//! the caller's API key: per-tool access control (the key's `allowed_tools`),
+//! rate-limit + budget (`quota::enforce`), input guardrails on the tool
+//! arguments, and a usage event into the shared sink. Output-side guardrails
+//! (scanning the tool result) are a follow-up.
 
 use std::time::{Duration, Instant};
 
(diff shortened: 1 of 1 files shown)


moonming:
## Independent audit (CLAUDE.md §8): APPROVE — no findings

Auditor reviewed cold + ran build/test/clippy in an isolated worktree (10 mcp tests pass, clippy clean, server builds). All three contract claims verified against source. Key confirmations:

- **No guardrail bypass** (the angle pushed hardest): the handler scans `params.arguments` from the buffered body and rebuilds the request with the **same** bytes; the gateway's `call_tool` forwards **only** `arguments` to the tool (not `_meta`/`task`). So *scanned == executed*. Malformed/absent args → the guardrail over-scans relative to what the tool receives (fail-safe direction).
- **Block precedes the upstream** (early `return` before `service.oneshot`) → no upstream contact, single usage emit.
- `model_id:""` can never match a Model-scoped guardrail (UUID ids); env/key/team resolve identically to the LLM path.
- **No leakage**: `args_text` is never logged or returned; the client message surfaces only the operator-assigned guardrail *name* — exact parity with the LLM 422 path.
- **200 + JSON-RPC error** divergence from the LLM 422 is justified (rmcp client expects a JSON-RPC envelope).

Non-blocking note (pre-existing, not this PR): a malformed/un-prefixed tool name emits usage with empty `mcp_tool` — from the ACL slice, call rejected downstream anyway.

Output-side guardrails honestly scoped as the tracked follow-up #671. Merging on CI green.

PR #750, opened 2026-07-10
test(e2e): add SeedClient — seed resources by writing canonical documents to etcd

moonming:
## What

- New harness client `SeedClient`: seeds `provider_keys` / `models` / `api_keys` / `observability_exporters` by writing the canonical resource document (the `schemas/resources/` shapes) straight to etcd under `<prefix>/<kind>/<id>`, instead of POSTing to the Admin API. The interface mirrors `AdminClient`'s create methods — `{id, value}` return, generated id, the same `provider`/`adapter` defaulting — so call sites can migrate mechanically (`admin.createModel({...})` → `seed.createModel({...})`).
- New characterization case `seed-vs-admin-characterization-e2e.test.ts`, pinning the equivalence the migration relies on through two lenses:
  - **Behavior** — chat completes through fully seed-created resources, and `allowed_models` authz rejects a seeded caller on a non-allowed model with `403`, exactly like an admin-created key. The positive probes double as propagation gates for both front doors.
  - **Shape** — read back through the store's serde round-trip (admin GET), a seeded sparse document is field-identical to the admin-created one, modulo identity fields and cross-references; identity fields themselves are pinned byte-exact. `api_keys` are additionally compared as **raw stored bytes** (via a new `EtcdClient.get`), because their GET view is a public projection that omits attribution fields — without the raw lens, handler-side enrichment of the stored document could hide behind the projection.

## Why

Writing documents directly is the same front door the control plane uses in managed mode, so cases seeded this way exercise the production write path rather than the Admin API, which only standalone deployments use for writes. The pattern already existed in the harness for resources the Admin API doesn't expose (`rate_limit_policies` in `team-member-ratelimit-e2e`); this generalizes it so any case can seed without the Admin API in the write path.

Follow-ups land separately: the mechanical sweep of existing cases' seeding, then a wait-condition audit.

## Verification

- New case: 2/2 green, including the raw-bytes lens with optional fields (`rate_limit`, `expires_at`) carried on both sides.
- Full local suite against etcd v3.5.15 + redis:7-alpine and a freshly built debug binary: **133 files / 269 tests passed** (~220s), re-run after review fixes.
- `tsc --noEmit`: zero errors in the added files (the 5 pre-existing `TS18048` warnings in untouched cases remain untouched).

## Review notes

An independent review of the first push was applied before re-push: the `api_keys` shape assertion was made non-vacuous (raw-bytes comparison + optional-field fixtures — its GET view is a projection), the round-trip comment now states precisely which lens covers what (store serde round-trip vs the loader's additional JSON-Schema validation, covered by the behavioral probes), and identity fields are pinned byte-exact. One known limitation kept as-is by design: the characterization covers the four `AdminClient` kinds; other kinds get their own coverage when their seeding migrates.

### DIFF
--- a/tests/e2e/src/cases/seed-vs-admin-characterization-e2e.test.ts
+++ b/tests/e2e/src/cases/seed-vs-admin-characterization-e2e.test.ts
@@ -1,0 +1,268 @@
+import { createHash } from "node:crypto";
+import OpenAI, { APIError } from "openai";
+import { afterAll, beforeAll, describe, expect, test } from "vitest";
+import {
+  AdminClient,
+  EtcdClient,
+  SeedClient,
+  spawnApp,
+  startOpenAiUpstream,
+  waitConfigPropagation,
+  type OpenAiUpstream,
+  type SpawnedApp,
+} from "../harness/index.js";
+
+// Characterization: a resource document seeded straight to etcd
+// (`SeedClient`) must be indistinguishable from the same resource
+// created through the Admin API — both after the store's serde
+// round-trip (read back via admin GET) and behaviorally on the proxy
+// (chat succeeds through seeded resources; allowed_models authz reads
+// a seeded caller key exactly like an admin-created one).
+//
+// This pins the contract the e2e seed migration relies on: the Admin
+// API's create handlers add nothing to the stored document beyond the
+// id in the key, so tests may write canonical documents directly —
+// the same front door the control plane uses in managed mode.
+// Transitional: this pin de-risks migrating case seeding to
+// direct document writes; once that migration completes it can be
+// folded into the general suite or retired.
+
+const ADMIN_CALLER = "sk-char-admin-caller";
+const SEED_CALLER = "sk-char-seed-caller";
+const sha256 = (s: string) => createHash("sha256").update(s).digest("hex");
+
+/** Strip the fields a pair intentionally differs by (identity and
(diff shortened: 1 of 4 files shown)


moonming:
Applied the independent review of the first push (force-pushed `38b2676`):

- **api_keys shape assertion was vacuous** — its GET view is a public projection that omits attribution fields, and with the varied fields stripped the comparison reduced to `{} == {}`. Fixed: the pair now carries optional fields (`rate_limit`, `expires_at`) so the projection lens has residue, and the raw stored bytes are compared via a new `EtcdClient.get` — handler-side enrichment of the stored document can no longer hide behind the projection.
- **Round-trip comment overstated its lens** — admin GET is the store's serde round-trip only; the proxy loader additionally applies JSON-Schema validation (covered by the behavioral 200-probes for traffic-bearing kinds). Comment and the `find()` error hint now say exactly that.
- **Identity fields pinned byte-exact** (`display_name`), so a hypothetical canonicalization can't hide behind `normalize()`.
- Wording in comments/PR body tightened to plain architecture terms.

Kept as-is by design: the propagation probe's opaque-timeout behavior follows the existing harness convention (`waitConfigPropagation`), and the characterization intentionally covers the four `AdminClient` kinds — other kinds get their own coverage when their seeding migrates.

Re-verified after the fixes: new case 2/2, full local suite 133 files / 269 tests green.

PR #756, opened 2026-07-12
test(e2e): seed cases via canonical documents instead of the Admin API

moonming:
## What

Migrates the e2e suite's **seeding** from the Admin API (`AdminClient`) to `SeedClient` (#750): canonical resource documents written straight to etcd at `<prefix>/<kind>/<id>` — the same direct-document front door the control plane uses in managed mode. The Admin API remains in the suite exactly where it is the *subject* or the *read lens*.

- **132 case files** now seed via `seed.createX(...)` / `seed.update(...)`.
- **SeedClient** grows `createGuardrail`, `createCachePolicy`, `createRateLimitPolicy` (kind segments pinned from `crates/aisix-admin/src/etcd_store.rs` `*_SUBKEY` constants and `crates/aisix-core/src/models/rate_limit_policy.rs`), plus generic `update(kind, id, value)` and `delete(kind, id)`. **EtcdClient** grows a single-key `delete` (etcd v3 `deleterange` with `range_end` omitted applies to only `key`).
- The **seed-vs-admin characterization** gains a third leg: a `seed.update`'d model demonstrably forwards the *new* `model_name` to the upstream, and a `seed.delete`'d api key stops authenticating (fail-closed, polled on the negative condition) while an admin-created control caller keeps serving.

## Why

Seeding through the same write path production uses in managed mode makes the suite exercise the loader/watcher contract directly, and drops a synchronous HTTP hop per seeded resource. PR #750 pinned seed≡admin equivalence (behavioral + serde round-trip + raw bytes); this PR banks on that pin.

## Propagation races the sweep exposed (fixed, not papered over)

Direct writes land faster than admin round-trips, which exposed **latent readiness races** in ~15 cases that admin-write latency had been masking — e.g. gates that polled `admin.listModels()` (a *store*-level read that never proved DP-snapshot propagation), chat probes without try/catch that aborted on the transitional 401, and a `status >= 400` gate that couldn't distinguish "guardrail live (422)" from "caller key not yet propagated (401)". Repairs strengthen the gates; **no assertion was weakened**:

- **Proxy-visible gates**: poll `/v1/models` with the caller bearer — it authenticates only once the key is in the DP snapshot and lists a model only once the snapshot has it, touching no upstream.
- **Revision-order canary**: virtual/router models don't appear in `/v1/models`, so routing cases either write the router *before* its targets or seed a throwaway canary api key *after* the resources under test — once the canary authenticates, everything written before it is in the snapshot (watch events apply in revision order). This never dispatches through the router, so cooldown warm-up and per-target counts stay clean.
- **Env-scoped roots**: cases that override the DP's etcd config (`cache-env-namespace`, `ratelimit-cluster`) now seed the exact root the DP reads (`<prefix>/<env_id>` / the shared replica prefix) instead of the harness default.
- **401-tolerant probes**: propagation conditions treat the transitional unknown-key 401 as "keep polling", never as success.

## Deliberately NOT migrated

| Case | Reason |
|---|---|
| `seed-vs-admin-characterization-e2e` | Admin is half the comparison by definition |
| `smoke.test.ts` | Its subject **is** the admin-write → proxy-read path |
| `apikey-budget-e2e` | Subject is admin-side validation (unknown-field 400) — unobservable via direct writes |
| `openai-sdk-compat.test.ts`, `datadog-exporter-e2e` | Deprecation-window coverage pair, marked `// Deliberately seeds via the Admin API: deprecation-window coverage.` |
| Read paths everywhere | `listModels` / `listModelStatuses`, revision GETs, the guardrail store-existence gate, admin-auth probes |
| `apikeys/:id/rotate` | An RPC with a server-generated secret, not a document write |
| `team-member-ratelimit` raw `etcd.put` | Case controls a fixed policy id; SeedClient generates ids |

## Verification

- Full suite green **twice**: `138 files / 288 tests` per run (~232s each), against a binary built from current `main`.
- `tsc --noEmit`: unchanged — the 5 pre-existing TS18048 warnings, zero new diagnostics.
- No `.only` / `.skip` introduced; assertions untouched.

## Summary by CodeRabbit

* **Tests**
  * Expanded end-to-end coverage for routing, caching, guardrails, rate limits, streaming, telemetry, passthrough endpoints, and provider integrations.
  * Improved configuration-propagation checks to reduce flaky test results and better reflect runtime availability.
  * Added coverage for updating and deleting seeded configuration, including lifecycle and credential-rotation scenarios.

* **Chores**
  * Improved test environment setup for consistent provisioning of models, credentials, policies, and observability settings.

### DIFF
--- a/tests/e2e/src/cases/aliyun-sls-content-capture-e2e.test.ts
+++ b/tests/e2e/src/cases/aliyun-sls-content-capture-e2e.test.ts
@@ -1,9 +1,9 @@
 import { createHash } from "node:crypto";
 import { afterAll, beforeAll, describe, expect, test } from "vitest";
 import {
-  AdminClient,
   decodedTextFor,
   EtcdClient,
+  SeedClient,
   spawnApp,
   startMockSls,
   startOpenAiUpstream,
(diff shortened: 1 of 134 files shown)


moonming:
Independent review: no HIGH/MEDIUM findings. Confirmed globally that zero assertion lines changed across the 10.8k-line diff (only poll gates were strengthened), the readiness idioms are snapshot-sound (canary writes ride etcd revision order; /v1/models probes read the proxy snapshot), the etcd-root fixes match `effective_prefix()`, and the held-back set has no stragglers. Non-blocking notes for a future touch-up: the new delete-gate can narrow `401 || 403` to `401`; two files gained a stray double blank line.

PR #763, opened 2026-07-14
feat(etcd): periodic auth token refresh to prevent expiry

okaybase:
## Summary

The etcd auth token acquired at connect time was never refreshed. Once it expired (default 300s `--auth-token-ttl`), every subsequent Admin API write and etcd watch operation failed with `etcdserver: invalid auth token`.

## Changes

- **`aisix-etcd`**: add `start_token_refresh_task` — a background Tokio task that calls `Client::refresh_token` on a configurable interval (default 240s, leaving a 60s margin below etcd's default 300s TTL). No-ops when no credentials are configured, or when called outside a Tokio runtime.
- **`aisix-core`**: add `EtcdConfig.auth_token_refresh_secs` field (default 240). Documented in `config.example.yaml` and `config.managed.yaml`.
- **`aisix-server`**: thread `auth_token_refresh_secs` from config through `EtcdConfigProvider::connect` and `EtcdConfigStore::new`.
- **`aisix-admin`**: `EtcdConfigStore::new` now spawns the refresh task for the admin write-path client, preventing Admin API writes from failing after token expiry.
- **Dependency**: bump `etcd-client` from 0.14 to 0.19 to pick up `Client::refresh_token` and the merged `WatchStream` (no longer requires a separate `Watcher` handle).
- **Tests**: add `admin_write_survives_token_expiry` and `token_refresh_prevents_auth_expiry` regression tests against an auth-enabled etcd with a 5s TTL, exercising the refresh loop.
- **CI**: add `etcd-auth` service container (short TTL) and "Enable auth on etcd-auth" step to run the token-refresh tests.

## Verification

- [x] `cargo test --workspace` passes
- [x] `cargo clippy --workspace -- -D warnings` clean
- [x] CI `rust-unit` job runs the new token-refresh tests against the auth-enabled etcd container
- [x] Manual: start gateway with `etcd.user` + `password_env`, verify Admin API writes still succeed after 5+ minutes

## Summary by CodeRabbit

- **New Features**
  - Added configurable etcd authentication-token refresh intervals.
  - Tokens refresh automatically to keep authenticated connections and watch operations available beyond token expiry.
  - The default refresh interval is 240 seconds and can be overridden in configuration.
  - Added retry handling for temporary token-refresh failures.

- **Bug Fixes**
  - Improved reliability for authenticated admin requests and watch operations.

- **Tests**
  - Added integration coverage for token refresh, authentication expiry, and authenticated watch events.


moonming:
Thanks for working on this — the underlying issue is real, and periodically refreshing credential-based etcd auth is valuable. The overall direction looks reasonable.

Before this is ready to merge, please address the following blocking items:

1. **Sync with the latest `main` and restore workspace compilation.**
   The current PR changes `EtcdConfigProvider::connect_with_policy` to take a fifth `refresh_secs` argument. After combining the PR with the latest `main`, `crates/aisix-server/src/export/mod.rs` still calls it with four arguments, and `cargo check --workspace` fails with `E0061`. Please update that call (the export path should likely pass `None`) or preserve the existing API and add a separate refresh-aware method.

2. **Retry promptly after a token-refresh failure.**
   The current loop logs a failed refresh and then waits until the next full interval. With the default 240s refresh interval and a 300s token TTL, a transient failure at the first refresh can let the current token expire and leave the process unauthenticated until the next tick (up to roughly 180s after expiry). Please add a short bounded retry/backoff after failures, then return to the normal interval after a successful refresh.

3. **Exercise the authenticated watch path in the regression test.**
   `token_refresh_prevents_auth_expiry` currently waits past the TTL and calls `load_all()`, so it only verifies a range/get request. Please create/recreate `provider.watch(...)` after the TTL window, write an event, and assert that the stream receives it. This is especially important because the PR also changes the `WatchStream` ownership/API while claiming to fix watch operations.

After updating the branch, please ensure the new fork workflow run is approved and actually executes all jobs. The current workflow runs are `action_required` with no jobs, so CodeRabbit and CLA are currently the only reported checks. At minimum, the updated branch should pass:

```bash
cargo fmt --all -- --check
cargo clippy --workspace --all-targets -- -D warnings
cargo test --workspace
cargo check --workspace
```

Once these items are addressed, this PR should be worth proceeding with.

okaybase:
@moonming Thanks for the detailed review! All blocking items have been addressed — compilation fix, retry logic, watch path coverage, and sync with main. Please take another look.

jarvis9443:
Thanks for this, and sorry it sat — the diagnosis here is right and the reproduction with a short `--auth-token-ttl` is what made the problem concrete.

We're going to fix it a different way though, and I'd rather explain than quietly supersede you.

The case that pushed us elsewhere is an etcd **restart** rather than a TTL expiry. A restart invalidates the token immediately, so a scheduled refresh only recovers at its next tick — up to 240s of failing calls with the default here. Reacting to the `Unauthenticated` error instead recovers on the very next call, and it covers plain TTL expiry too, since it keys off what actually happened rather than a predicted schedule.

Two smaller things pointed the same way. The 240s default is really a client-side copy of etcd's own `--auth-token-ttl`, which the gateway can't observe — set a shorter TTL server-side and the default silently stops being enough. And #1135 (merged today) added error classification that separates "server never answered" from "server refused these credentials", which is exactly what a reactive path needs to avoid turning a wrong password into an infinite re-auth loop. That classification didn't exist when you opened this.

The 0.14 → 0.19 bump is also more awkward now than it was in July: #1134 and #1135 both changed the watch path, and the merged `WatchStream` in 0.19 lands right on top of that.

What we are keeping is your test setup — the auth-enabled etcd service container with the short TTL, and the integration tests built on it. That infrastructure didn't exist in this repo before your PR and it's what any fix needs to prove itself against, so it's going in with attribution to this PR.

I'll link the replacement here once it's up so you can see whether it actually handles your case; if it misses something you hit in production, please say so on it.

nic-6443:
Opened #1136 with the reactive version of this — on `Unauthenticated` the client is discarded and re-dialled, which re-runs `Authenticate`, and the call is retried once. It carries over your CI infrastructure (the authenticated etcd container with the short `--auth-token-ttl`) and the idea of asserting recovery against a real expiry rather than a simulated one, with the tests rewritten to assert recovery instead of refresh. Credited in the PR description.

Not carried over: `start_token_refresh_task`, the `auth_token_refresh_secs` config field, and the dependency bump.

One thing that came out of building it, in case it's useful: restarting etcd 3.5 does *not* by itself invalidate a simple token — `Authenticate` is a raft entry, so replaying the WAL on startup re-registers the tokens it minted, and a client connected before the restart keeps working. Verified against 3.5.18. What does invalidate is the token store actually being cleared.

jarvis9443:
Thanks for this, @okaybase, and sorry it sat for so long. The underlying problem is fixed on main by #1136, which takes a different route: instead of refreshing the token on a timer, the etcd client treats an expired or staled token (gRPC Unauthenticated, plus etcd's `revision of auth store is old` / `user name is empty` errors) as a signal to re-dial and re-authenticate, then retries the call once. That also covers the case a timer can't — an auth-store change such as `etcdctl user add` invalidating every outstanding token at once — and it needs no TTL setting to keep in step with etcd's `--auth-token-ttl`.

#1136 kept your approach to CI (etcd with auth enabled) and your recovery-after-a-real-expiry test, and credits you in its commit message. Closing this one as superseded.

PR #777, opened 2026-07-16
feat(heartbeat): report applied config_hash for cross-plane config verification

moonming:
## What

Adds an optional `config_hash` field to the data plane's heartbeat body: the SHA-256 of the config snapshot the gateway currently has applied. The control plane can now compare each node's reported hash against what it expects, turning "is this node actually running the config I pushed?" into a direct per-node check.

## Why

The hash is already computed and exposed locally by `Supervisor::config_status()` (the same value served on `/status/config`). Reporting it on the heartbeat lets the control plane do expected-vs-reported verification per data plane node. The control-plane ingestion side already landed and tolerates the field's absence, so this is purely additive on the wire.

## Contract

- Serialized as `config_hash` with `skip_serializing_if = "Option::is_none"` — heartbeats from configs without the fetcher wired emit the historical body shape unchanged.
- Written on telemetry-bearing beats; clamped to 128 chars (defensive — the hash is 64 hex chars).
- Hash definition matches the `/status/config` surface: etcd mode = sha256 over `key '\0' canonical_json '\n'` in ascending key order; file mode = sha256 of the raw file bytes.

## Verification

- Unit: body serializes `config_hash` when the fetcher yields a value, omits it (legacy shape) when None, clamp behavior.
- fmt/clippy -D clean; cargo test --workspace green; full e2e suite unaffected (no proxy/loader behavior change).

## Summary by CodeRabbit

* **New Features**
  * Heartbeats now report the hash of the currently applied configuration.
  * Configuration hashes are omitted when unavailable and safely limited to 128 characters.
  * Added visibility into whether the latest configuration has been applied through status reporting.

### DIFF
--- a/crates/aisix-core/src/config_status.rs
+++ b/crates/aisix-core/src/config_status.rs
@@ -387,6 +387,15 @@ impl ConfigStatus {
         self.inner.lock().unwrap().ever_applied
     }
 
+    /// The hash of the last applied (served) config snapshot, or `None`
+    /// when no snapshot has been applied yet this boot. A cheap targeted
+    /// read for callers that need only the hash — the heartbeat's
+    /// per-node config-verification field — without building the full
+    /// [`Self::view`] / [`Self::metrics`] snapshot.
+    pub fn applied_config_hash(&self) -> Option<String> {
+        self.inner.lock().unwrap().config_hash.clone()
+    }
+
     /// Point-in-time JSON view for `GET /status/config`.
     pub fn view(&self) -> ConfigStatusView {
         self.inner.lock().unwrap().view()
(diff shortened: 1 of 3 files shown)


moonming:
Independent cold audit: **MERGE**, no HIGH/MEDIUM correctness issues. Verified directly in code that `applied_config_hash()` returns the served (applied) hash, never the observed source_hash, and that a wholly-rejected reload retains the last-good served hash; `skip_serializing_if` preserves the legacy body shape; char-boundary-safe 128 clamp; managed/etcd-mode scoping correct.

Fixed the audit's MEDIUM-1 + LOW-1 (`fdb` follow-up commit): the accessor test used identical source/applied values so it couldn't catch reading the wrong field — now uses distinct `src*`/`applied*` values and adds a wholly-rejected-reload assertion pinning last-good retention.

LOW-2 (cross-plane completeness) is satisfied: the paired CP ingestion already landed (AISIX-Cloud#1047 — nullable `config_hash` column + heartbeat reader), so this is not a report-into-the-void half-feature.

PR #778, opened 2026-07-16
feat(observability): serve per-model runtime health on the status listener

moonming:
## What

`GET /status/models` on the dedicated metrics/status listener: the per-model runtime health view (cooldown / background-check state) as an operational read, next to `/status/config` and `/status/ready`. The response is exactly what `GET /admin/v1/models/status` serves; the admin endpoint is unchanged.

## Why

Operational reads now have a home on the operational listener. The runtime health view is proxy-runtime state — which targets are cooling down and why — the kind of state probes, dashboards, and scrape-side tooling poll on day 2. Serving it only on the admin listener couples those reads to the management surface and its admin-key auth (and managed mode does not bind an admin listener at all). The metrics/status listener is already the unauthenticated, network-restricted operational surface (#774), identical in every mode, which mirrors the ecosystem-wide split between a management API and a dedicated status/health port.

Sensitivity class matches the listener: the view exposes the model catalog (ids and display names) and health states — the same class as `/status/config`'s `resource_counts` and `rejected[]` resource ids already served there. No auth, like the scrape — restrict at the network layer.

## Design: one render, one source

- `render_models_status(...)` is extracted in `models_status_handler.rs` and is the single render behind **both** endpoints, so the two responses cannot drift while both exist.
- The status listener reads through the **same** `Arc<dyn ConfigStore>` handle the admin surface uses (etcd store in etcd mode, snapshot read-view in file mode) plus the same shared `ModelRuntimeStatusTracker` — same data, same ordering, same serialization. A unit test pins the two bodies **byte-for-byte**.
- Managed mode (no admin store) serves the applied snapshot through the same read-only store view file mode uses, keeping the metrics/status listener surface identical across modes.
- Store failures answer a fixed `{"error_msg": "failed to list models"}` 500 on this unauthenticated surface — backend detail (etcd endpoints, connection errors) stays in the server log, while the admin-key-gated endpoint keeps its detailed envelope. Success bodies stay byte-identical.
- The store read is bounded by a 5-second timeout that degrades into the same fixed 500: a hung store (blackholed etcd) cannot park anonymous pollers or hold the shared store client while they pile up.
- No OpenAPI change: the generated Admin API spec documents the admin listener only; status-listener endpoints follow the `/status/config` precedent. Docs-site follow-up will list `/status/models` alongside the other status endpoints.

## e2e harness repoint

`AdminClient.listModelStatuses()` now reads `GET /status/models` on the metrics/status listener (new optional `metricsBaseUrl` constructor argument; method signature and return shape unchanged). The existing runtime-status consumers — background-health, cooldown-contract, retry-on-429-vs-background-ignore, runtime-mixed-filtering, runtime-status — exercise the new endpoint end-to-end with their assertions untouched.

## Tests

- Unit: `status_models_serves_the_admin_view_byte_for_byte` (same store + tracker handles → byte-identical bodies, covering a cooldown row and a virtual `not_applicable` row); `status_models_store_failure_never_leaks_backend_detail` (failing store → fixed 500 body, no backend detail); `status_models_answers_the_fixed_500_when_the_store_hangs` (paused-clock test with a never-resolving store → the timeout arm answers the same fixed body); `admin_router_does_not_serve_status_models`; the existing metrics-listener tests updated for the new router argument.
- e2e (`status-models-e2e.test.ts`, new):
  - etcd mode: one model tripped into cooldown by a real 401 upstream, then `GET /status/models` vs `GET /admin/v1/models/status` asserted deep-equal **and** raw-text-equal (same order, same bytes), plus the expected cooldown contract (`status=cooldown`, `status_reason=upstream_auth_failure`, `cooldown_until` set).
  - Auth split: the status listener serves without a key while the admin route still requires one.
  - File mode: the view serves on the status listener and matches the admin endpoint.
- Managed mode is wired but not e2e-covered (the e2e harness has no control plane); its serving path — the read-only snapshot store view feeding the shared render on the status listener — is the same one the file-mode e2e cases exercise, just with a different snapshot feeder.
- `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` green; full e2e suite green twice locally.

## Summary by CodeRabbit

* **New Features**
  * Added an unauthenticated `GET /status/models` endpoint on the metrics/status listener.
  * Exposes per-model runtime health and status information, matching the admin model-status view.
  * Supports model status reporting in both managed and standalone deployments.

* **Bug Fixes**
  * Model listing failures and timeouts now return a consistent, non-sensitive error response.
  * Improved status-client handling for metrics endpoint requests.

### DIFF
--- a/crates/aisix-admin/Cargo.toml
+++ b/crates/aisix-admin/Cargo.toml
@@ -28,7 +28,7 @@ etcd-client.workspace = true
 aisix-obs = { path = "../aisix-obs" }
 
 [dev-dependencies]
-tokio = { workspace = true, features = ["macros", "rt"] }
+tokio = { workspace = true, features = ["macros", "rt", "test-util"] }
 testcontainers.workspace = true
 aisix-gateway = { path = "../aisix-gateway" }
 aisix-provider-openai = { path = "../aisix-provider-openai" }
--- a/crates/aisix-admin/src/lib.rs
+++ b/crates/aisix-admin/src/lib.rs
@@ -440,14 +440,35 @@ async fn status_ready_handler(
 /// while both exist. Unauthenticated like `/status/config` — it exposes
 /// model names and health states, the same sensitivity class; restrict
 /// access at the network layer.
+///
+/// Store failures answer a fixed `{"error_msg": "failed to list models"}`
+/// 500: this listener is unauthenticated, so backend detail (etcd
+/// endpoints, connection errors) stays in the server log instead of the
+/// response body. The admin-key-gated endpoint keeps its detailed
+/// envelope.
 async fn status_models_handler(
     axum::extract::State(state): axum::extract::State<MetricsState>,
-) -> Result<axum::Json<Vec<models_status_handler::ModelStatusView>>, AdminError> {
-    let all_models = state.models_status.store.list_models().await?;
-    Ok(axum::Json(models_status_handler::render_models_status(
+) -> Response {
+    use axum::response::IntoResponse;
+
+    let all_models = match state.models_status.store.list_models().await {
+        Ok(models) => models,
+        Err(e) => {
+            tracing::error!(error = %e, "GET /status/models: listing models failed");
+            return (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                axum::Json(ErrorBody {
+                    error_msg: "failed to list models".into(),
+                }),
+            )
+                .into_response();
+        }
+    };
+    axum::Json(models_status_handler::render_models_status(
         all_models,
         state.models_status.runtime_status_tracker.as_deref(),
-    )))
+    ))
+    .into_response()
 }
 
 fn normalized_prometheus_path(path: &str) -> String {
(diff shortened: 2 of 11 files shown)


moonming:
**Review triage**

- **Store-error detail on the unauthenticated listener** (CodeRabbit security check; also raised by internal review) — addressed in the follow-up commit: `GET /status/models` now answers a fixed `{"error_msg": "failed to list models"}` 500 on store failure and keeps the backend detail (etcd endpoints, connection errors) in the server log; the admin-key-gated endpoint keeps its detailed envelope, and success bodies remain byte-identical. Pinned by the new `status_models_store_failure_never_leaks_backend_detail` unit test (failing store → fixed body, no backend detail in the response).
- **Transient non-200 handling inside the new e2e polls** — kept as-is deliberately: it matches the suite's existing convention (the previous admin-listener helper threw on non-200 inside polls too), and a non-200 here only occurs during a store outage, which the test should surface loudly rather than absorb.
- **`prometheus.path` colliding with a `/status/*` route** — pre-existing class (`/status/config` and `/status/ready` have the same property): the router panics loudly at boot on an overlapping route, which is the preferable failure mode for an operator misconfiguration; silently rewriting the operator's configured scrape path would hide the mistake. Can revisit separately as a config-load validation if wanted.
- **Managed-mode e2e coverage** — now stated in the PR body: the e2e harness has no control plane, and managed mode's serving path (read-only snapshot view feeding the shared render) is the same one the file-mode e2e cases exercise.

moonming:
@coderabbitai review

moonming:
@coderabbitai full review

moonming:
@coderabbitai summary

moonming:
**Independent cold audit** (second pass, at `4ac09c0`)

- **MEDIUM — unbounded store read on the unauthenticated endpoint**: a hung configuration store (blackholed etcd, as opposed to a fast-failing one) would park anonymous `/status/models` pollers indefinitely while holding the shared store client, contradicting the "store failures answer a fixed 500" contract. **Fixed in `44fff83`**: the store read is bounded by a 5s timeout that degrades into the same fixed 500; pinned by a paused-clock unit test with a never-resolving store (`status_models_answers_the_fixed_500_when_the_store_hangs`).
- **LOW — exposure wording**: "model names and health states" understated the surface slightly; doc comment and PR body now say "the model catalog (ids and display names) and health states". No new fields exposed — the audit verified no provider name, `api_base`, or upstream model name reaches the body, and every `status_reason` is a fixed static string.
- **LOW — `prometheus.path: /status/models` collision** and **LOW — e2e polls abort on transient non-200**: both verified as the pre-existing classes already dispositioned in the triage above; non-blocking.

All four earlier triage dispositions were independently re-verified as sound (redaction covers the only fallible arm; route collisions panic loudly at boot rather than shadowing; managed mode shares the file-mode serving path; byte-for-byte pinned across empty/tracker-unknown/virtual rows).

PR #791, opened 2026-07-21
feat(admin): add admin.enabled switch to run without the admin listener

moonming:
## What

Add an `admin.enabled` startup-config switch (default `true`) that lets a standalone gateway run **without binding the admin listener**, and wire the e2e harness to spawn in that mode. This is the foundation for retiring the Admin API: it makes the admin-listener-off configuration a runnable, tested reality during the coexistence window, before any admin code is deleted.

## Why

The gateway already runs admin-less in managed mode (config comes from the control plane over etcd). But in standalone etcd/file mode the admin listener was always bound — there was no way to preview the post-removal world, and the e2e harness had no way to prove the request path works without it. `admin.enabled = false` closes that gap:

- Operators can run the admin-free configuration on the coexistence release, de-risking the eventual code removal.
- The e2e suite gains a spawn mode that exercises the declarative-only path (resources seeded straight to etcd via `SeedClient`, never the Admin API).

`admin.enabled` is a process-level startup-config toggle (like `proxy.addr`), not a per-resource setting, so it carries no control-plane/schema work.

## Changes

- **`AdminConfig.enabled: bool`** (`crates/aisix-core/src/config.rs`), defaults to `true`. When `false`, `Config::validate` no longer requires `admin_keys`/`admin.addr` (there is no surface to authenticate), mirroring the existing managed-mode relaxation.
- **Bind gating** (`crates/aisix-server/src/main.rs`): the admin listener is spawned only when its store is present **and** `admin.enabled`. `/status/config`, `/status/ready`, `/status/models`, `/metrics` (metrics listener) and the proxy `/livez` are unaffected — the metrics/status listener is bound independently and still reads models through the same store handle, so `/status/models` keeps working with the admin listener off.
- **Harness** (`tests/e2e/src/harness/app.ts`): `AppOverrides.admin?: boolean` (default `true`). With `admin: false`, the generated config carries `admin.enabled = false` and readiness gates on the proxy `/livez` + the metrics listener instead of `/admin/v1/health`.
- **New e2e** (`admin-disabled-e2e.test.ts`): spawns `admin: false`, seeds a provider key + model + caller key **only through etcd**, and asserts (1) a proxy chat request succeeds, (2) `/status/config` reports the applied config on the metrics listener, (3) the admin port is not bound (connection refused).
- Config unit tests: `enabled` defaults to `true`; `enabled: false` relaxes the `admin_keys` requirement.

## Verification

- `cargo fmt`, `cargo clippy -p aisix-core -p aisix-server -p aisix-admin --all-targets -D warnings`, `cargo test -p aisix-core --lib` (config, +2 new) and `-p aisix-admin --lib` (114) green.
- New `admin-disabled-e2e` green (3/3). Regression: `allowed-models` and `seed-vs-admin-characterization` (exercises both the admin and etcd-seed paths) green — with `admin` defaulting to `true`, the admin-on config + readiness are byte-identical to before, so existing tests are unaffected. Full e2e suite green locally.

## Scope / follow-ups (this is P0-1 step 1 of the Admin API removal — AISIX-Cloud#1007 Phase 4)

This PR delivers the **switch**. It does not yet flip the suite: several tests still seed through `AdminClient`, and a few deliberately exercise the Admin API (characterization, file-mode 409 rejection, key rotation, auth baseline). Follow-ups:

1. Migrate the remaining seed-only holdouts (`AdminClient` → `SeedClient`) and relocate the harness's default readiness probe off `/admin/v1/health`.
2. Segregate the Admin-API-testing cases as the held-back set that runs admin-on until removal, and add a suite lane that runs the rest admin-off.
3. Document `admin.enabled` in the configuration-files reference (api7/docs).

## Summary by CodeRabbit

* **New Features**
  * Added an option to disable the admin listener while keeping the gateway operational.
  * Gateway readiness no longer depends on the admin health endpoint when the listener is disabled.
  * Configuration seeded through etcd remains available for requests and status reporting.

* **Bug Fixes**
  * Improved configuration validation so admin settings are only required when the admin listener is enabled.

* **Tests**
  * Added end-to-end coverage for gateway operation without the admin listener.

### DIFF
--- a/crates/aisix-admin/src/lib.rs
+++ b/crates/aisix-admin/src/lib.rs
@@ -545,6 +545,7 @@ mod tests {
 
     fn cfg() -> AdminConfig {
         AdminConfig {
+            enabled: true,
             addr: "127.0.0.1:0".into(),
             admin_keys: vec!["admin-secret".into()],
             tls: None,
--- a/crates/aisix-admin/src/playground_handler.rs
+++ b/crates/aisix-admin/src/playground_handler.rs
@@ -68,6 +68,7 @@ mod tests {
 
     fn admin_cfg() -> AdminConfig {
         AdminConfig {
+            enabled: true,
             addr: "127.0.0.1:0".into(),
             admin_keys: vec!["admin-key".into()],
             tls: None,
--- a/crates/aisix-admin/tests/etcd_integration.rs
+++ b/crates/aisix-admin/tests/etcd_integration.rs
@@ -57,6 +57,7 @@ async fn build_state_with_real_etcd(url: &str, prefix: &str) -> AdminState {
     let store: Arc<dyn ConfigStore> = Arc::new(EtcdConfigStore::new(client, prefix));
     let handle = SnapshotHandle::new(AisixSnapshot::new());
     let cfg = AdminConfig {
+        enabled: true,
         addr: "127.0.0.1:0".into(),
         admin_keys: vec![ADMIN_KEY.into()],
         tls: None,
--- a/crates/aisix-core/src/config.rs
+++ b/crates/aisix-core/src/config.rs
@@ -434,6 +434,15 @@ impl RealIpConfig {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub struct AdminConfig {
+    /// When `false`, the admin listener is not bound even in standalone
+    /// (etcd or file) mode. The proxy and the metrics/status listener are
+    /// unaffected, so resources are managed declaratively — a resources
+    /// file, or direct writes to the configuration store — with
+    /// `GET /status/config` and the proxy `GET /livez` as the operational
+    /// feedback. Managed mode never binds the admin listener regardless.
+    /// Defaults to `true`.
+    #[serde(default = "AdminConfig::default_enabled")]
+    pub enabled: bool,
     #[serde(default = "AdminConfig::default_addr")]
     pub addr: String,
     /// Statically-provisioned admin keys. A request is authorised if it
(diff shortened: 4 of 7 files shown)


moonming:
@coderabbitai full review

moonming:
**Independent cold audit** (diverse-lens: correctness / security+breaking / e2e-coverage, 3 agents → synthesis). All findings addressed in `f086784`.

**MEDIUM — file-mode admin-off had zero coverage.** File mode binds the admin surface through a distinct store branch (`FileManagedStore` vs `EtcdConfigStore`), and every new test was etcd-only; the doc advertises admin-off "even in standalone (etcd or file) mode". Behavior was correct today (all three lenses confirmed the runtime is right and default-preserving), but the file branch had no driver → blocks under the e2e-contract merge gate. **Fixed**: added `admin_disabled_relaxes_admin_key_requirement_in_file_mode` config unit test + a file-source admin-off e2e leg (declarative `resources.yaml`, one proxy request, `/status/config` reports the file-loaded model, admin port refused).

**LOW (batched into the same commit):**
- Redundant admin etcd client connected at boot when admin-off (etcd mode) → guard extended to `is_managed() || !admin.enabled`; `/status/models` then reads the snapshot (as in managed mode), and boot no longer risks failing on a connection it immediately drops.
- Admin-unbound assertion accepted any throw → now asserts the specific `ECONNREFUSED` cause, so a stray listener or a DNS/abort error can't satisfy it.
- `/status/config` count asserted `>= 1` → now exact (`models/provider_keys/api_keys == 1`) under the unique etcd prefix.
- admin-off + prometheus-off reduces readiness to proxy `/livez` → documented at the readiness site (latent; the default keeps prometheus on).

The correctness and security lenses independently confirmed `enabled = true` reproduces prior behavior bit-for-bit (non-breaking) and admin-off drops no security control other than the admin surface itself. Verification: `cargo fmt`/`clippy`/`aisix-core` config (35, +new file-mode) green; `admin-disabled-e2e` 6/6 (etcd + file legs) green; full suite green earlier (3 transient etcd/ECONNRESET flakes, pass on isolated re-run).

moonming:
@coderabbitai review

PR #792, opened 2026-07-21
test(e2e): migrate the seed-tail tests off the Admin API onto etcd/status

moonming:
## What

Migrate the e2e tests that still depended on the **Admin API for setup or readiness** onto the direct-etcd `SeedClient` and the metrics/status listener, and run them with `admin: false`. This is step 2 of retiring the Admin API (after #791 landed the `admin.enabled` switch): it moves the tests that *don't test the Admin API itself* out of the admin-on world, shrinking the Admin API's remaining e2e footprint to only the tests whose subject is the admin surface.

9 files migrated (all now spawn `admin: false`):

**Seed-only (4)** — the Admin API was used purely to set up resources or as a removable oracle:
- `cache-ttl-eviction` — already seeded via `SeedClient`; only needed `admin: false`.
- `datadog-exporter` — `AdminClient` → `SeedClient` for the exporter + routing seed (the exporter delivery is the subject, not how it was seeded; the admin-write-path coverage lives in the held-back characterization test).
- `guardrail-disabled-bypass` — the Gate-A readiness probe (`GET /admin/v1/guardrails`, proving the rule *loaded*) becomes a `GET /status/config` check on `applied.resource_counts.guardrails`, which reflects the applied snapshot on the metrics listener — **stronger** than reading etcd back (which would only prove the write landed), preserving the anti-vacuous intent.
- `provider-key-rotation` — the revision oracle (`GET /admin/v1/provider_keys/:id` → `.revision` bumped) becomes an etcd read-back asserting the key now holds the **rotated secret** — a more direct no-op guard than a revision bump, using the existing `EtcdClient.get` (no shared-harness change).

**Mixed (5)** — seeding was already on `SeedClient`; the only remaining admin-shaped call is `listModelStatuses`, which reads `GET /status/models` on the **metrics listener** (not the admin API) and survives `admin: false`: `background-health`, `runtime-mixed-filtering`, `cooldown-contract` (7 spawn sites), `retry-on-429-vs-background-ignore`, `runtime-status`.

## Out of scope (deliberately unchanged)

Tests whose **subject is the Admin API surface** stay admin-on until the Admin API is removed, then get deleted or split — tracked on the removal checklist (AISIX-Cloud#1007 Phase 4): `seed-vs-admin-characterization` (seed≡admin equivalence), `file-resource-source` (file-mode write-409), `status-models` (`/status/models` vs `/admin/v1/models/status` equivalence), `health-minimal` (`/admin/v1/health`), `apikey-lifecycle`/`apikey-budget` (rotate/budget admin ops), `auth-baseline`, `openai-sdk-compat` (RFC 9745 Deprecation-header contract), `prometheus-metrics`. `status-config` only uses `/admin/v1/health` as a probe (it doesn't seed via admin) — left for the readiness-default step.

## Method

Classification (seed-only / mixed / held-back / probe-only) was done by fanning out one reader per candidate file, then synthesized into this migration plan — so each file's admin usage was read and classified individually rather than pattern-matched.

## Verification

`npx tsc --noEmit` adds no new type errors in the changed files. Each migrated file passes in isolation; full e2e suite green locally. No harness or Rust changes — the direct-etcd seed infra (`SeedClient`/`EtcdClient`) already existed, and `/status/models` / `/status/config` on the metrics listener already serve with the admin listener off (#791). The migrated tests keep their assertions; only the seed/readiness transport moved off the Admin API (two readiness/no-op guards were strengthened, noted above).

## Summary by CodeRabbit

* **Tests**
  * Updated end-to-end test setups to run without the admin listener where it is not required.
  * Improved readiness and configuration checks using status endpoints and direct resource seeding.
  * Updated exporter, guardrail, and provider-key rotation scenarios to validate behavior through supported runtime and storage checks.
  * Preserved coverage for health, caching, routing, retries, filtering, and runtime status behavior.

### DIFF
--- a/tests/e2e/src/cases/background-health-e2e.test.ts
+++ b/tests/e2e/src/cases/background-health-e2e.test.ts
@@ -59,7 +59,10 @@ describe("background health e2e", () => {
       },
     });
 
-    app = await spawnApp();
+    // The admin listener is off; `admin` here is used only for
+    // listModelStatuses, which reads GET /status/models on the metrics
+    // listener. Resources are seeded straight to etcd via `seed`.
+    app = await spawnApp({ admin: false });
     admin = new AdminClient(app.adminUrl, app.adminKey, app.metricsUrl);
     seed = new SeedClient(etcd, app.etcdPrefix);
 
--- a/tests/e2e/src/cases/cache-ttl-eviction-e2e.test.ts
+++ b/tests/e2e/src/cases/cache-ttl-eviction-e2e.test.ts
@@ -57,7 +57,8 @@ describe("cache TTL eviction e2e: entry expires after ttl_seconds", () => {
     if (!etcdReachable) return;
 
     upstream = await startOpenAiUpstream();
-    app = await spawnApp();
+    // No admin listener: every resource here is seeded straight to etcd.
+    app = await spawnApp({ admin: false });
     seed = new SeedClient(etcd, app.etcdPrefix);
 
     const pk = await seed.createProviderKey({
--- a/tests/e2e/src/cases/cooldown-contract-e2e.test.ts
+++ b/tests/e2e/src/cases/cooldown-contract-e2e.test.ts
@@ -74,7 +74,9 @@ describe("cooldown contract (H1) — 401 cools down despite being non-retryable"
       },
     });
 
-    app = await spawnApp();
+    // Admin listener off; `admin` reads only /status/models on the metrics
+    // listener, and every resource is seeded to etcd via `seed`.
+    app = await spawnApp({ admin: false });
     admin = new AdminClient(app.adminUrl, app.adminKey, app.metricsUrl);
     seed = new SeedClient(etcd, app.etcdPrefix);
 
@@ -225,7 +227,9 @@ describe("cooldown contract (M1) — 429 cools down even when retry_on_429=false
       },
     });
 
-    app = await spawnApp();
+    // Admin listener off; `admin` reads only /status/models on the metrics
+    // listener, and every resource is seeded to etcd via `seed`.
+    app = await spawnApp({ admin: false });
     admin = new AdminClient(app.adminUrl, app.adminKey, app.metricsUrl);
     seed = new SeedClient(etcd, app.etcdPrefix);
 
@@ -378,7 +382,9 @@ describe("cooldown contract (H2) — Retry-After header from upstream drives TTL
       },
     });
 
-    app = await spawnApp();
+    // Admin listener off; `admin` reads only /status/models on the metrics
+    // listener, and every resource is seeded to etcd via `seed`.
+    app = await spawnApp({ admin: false });
     admin = new AdminClient(app.adminUrl, app.adminKey, app.metricsUrl);
     seed = new SeedClient(etcd, app.etcdPrefix);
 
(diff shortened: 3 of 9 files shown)


moonming:
@coderabbitai full review

moonming:
**Independent cold audit** (3 diverse lenses — coverage-preservation / admin-off-safety / classification-boundary). **All three PASS, zero HIGH/MEDIUM.**

- **Coverage-preservation** (traced to DP source): both "strengthened" guards are genuinely stronger, not weaker. guardrail Gate-A now reads `/status/config` `applied.resource_counts.guardrails`, which is the **same published snapshot the proxy consults** at dispatch (`supervisor.rs` resource_counts over `AisixSnapshot`) — strictly stronger than the old admin etcd-store read, which ran one layer *ahead* of the applied snapshot (could clear before the rule loaded). `enabled:false` guardrails still load (filter is at dispatch, not load), so the gate is satisfiable. provider-key-rotation's secret read-back catches "no write" AND "wrong content" (vs revision-bump = "some write"); the rotation itself was already a declarative `seed.update`, and the real DP-liveness signal (`success === TOTAL_REQUESTS`) is unchanged. MIXED `/status/models`: model list comes from the applied snapshot under admin-off, but the runtime health overlay uses the **same shared tracker** in both modes — assertions are `.find(row)!` preceded by proxy traffic guaranteeing snapshot presence, so no false-PASS race.
- **Admin-off-safety**: no residual admin-API call in the 9 (only `listModelStatuses` → metrics listener survives); `AdminClient` construction safe under admin:false; all 7 cooldown spawn sites hit; imports/vars clean; `etcd!` assertions proven safe.
- **Classification-boundary**: the 9 migrated files assert nothing about the admin surface; every held-back file genuinely tests it (401s, deprecation headers, rotate endpoint, 409 path-leak, seed≡admin equivalence); `status-config` seeds via etcd and only probes `/admin/v1/health` (correctly excluded); **no admin-seeding file was missed**. datadog's dropped admin `observability_exporter` write is still covered by the held-back characterization test — no net coverage loss.

LOWs: provider-key-rotation's stale `PUT /admin/v1/...` comments refreshed in `d102574`; guardrail Gate-A `>= 1` is immaterial today (exactly one guardrail seeded) and stronger on the applied-vs-written axis (kept). Verification: `npx tsc` clean on the changed files; each migrated file green in isolation; full suite green (the one cooldown ECONNRESET was a transient under 148-file load — passes 2× on isolated re-run).

moonming:
**CI + CodeRabbit triage**

**CI e2e failure = pre-existing flake, not this PR.** The failing test was `status-config-e2e` ("clean config reports synced with revisions and resource counts") — a file this PR does not touch, and this PR makes no change to `/status/config` behavior (test-only, no Rust/harness change). It waits for `state === "synced"` then asserts `resource_counts.models === 1`; the same response returned synced with `resource_counts` not yet populated. Passes 3/3 on isolated local re-run. Filed as #793. Re-triggered CI on `eafbbbb`.

**CodeRabbit comments:**
1. *guardrail-disabled-bypass — check `res.ok` before `.json()`*: **fixed in `eafbbbb`** (the readiness poll now returns false on a non-200 instead of parsing an error body).
2. *provider-key-rotation — skip only on `!etcdReachable`, fail on other missing prerequisites*: valid convention, but this is a **pre-existing suite-wide guard pattern** (`if (!etcdReachable || !app || …) ctx.skip()`) used across the e2e cases; this PR only swapped `!admin` → `!etcd` inside it and did not change the skip semantics. Aligning it belongs in a separate suite-wide consistency pass rather than piecemeal in the migrated files (which would leave the suite inconsistent) — deferring, not silently ignoring.

PR #795, opened 2026-07-21
test(e2e): fix propagation race in background-health short-stale-window case

jarvis9443:
The short-stale-window case reads `/admin/v1/models/status` immediately after seeding the model and asserts the row exists. Config propagation is asynchronous, so on a loaded CI runner the immediate read races the snapshot refresh and fails with `expected undefined to be truthy` — hit on an unrelated PR's e2e run (#794's `e2e (vitest) + coverage` job, where the only delta was a `cargo fmt` commit).

Fix: poll for the row via `waitConfigPropagation`, exactly like the unhealthy-transition check a few lines below already does. The scenario is unchanged — the row must appear and then go unhealthy; only the eager, un-waited existence assert is gone.

Verified locally: the file's 2 tests pass; the fix only touches the racy setup assert.

### DIFF
--- a/tests/e2e/src/cases/background-health-e2e.test.ts
+++ b/tests/e2e/src/cases/background-health-e2e.test.ts
@@ -249,9 +249,14 @@ describe("background health e2e", () => {
       },
     });
 
-    const statuses = await admin.listModelStatuses();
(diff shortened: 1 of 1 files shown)


jarvis9443:
@coderabbitai review

PR #800, opened 2026-07-22
test(e2e): default the harness to admin-off; held-back tests opt in

moonming:
## What

Flip the e2e harness default to **`admin: false`**, so the whole suite runs with **no admin listener bound** by default. The only tests that keep the admin listener are the ones whose subject **is** the Admin API surface — they now opt in explicitly with `admin: true`.

This is the last step of the e2e "de-admin-API" work (after #791 added the `admin.enabled` switch and #792 migrated the seed-tail tests). It proves the acceptance for retiring the Admin API: **the suite is green with the admin listener off by default**, so when the Admin API is removed, only the explicitly-`admin: true` files need deletion — everything else already runs admin-off.

## Changes

- **`tests/e2e/src/harness/app.ts`**: `AppOverrides.admin` now defaults to `false` (`overrides.admin ?? false`). Doc updated: admin-off is the norm; only held-back admin-surface tests opt in with `admin: true`.
- **9 held-back files** get an explicit `admin: true` on every `spawnApp` site — a no-op behaviorally (this was the previous default), just making the dependency explicit now that the default flipped: `apikey-budget`, `apikey-lifecycle` (admin key rotate / removed-field validation), `auth-baseline` (admin bearer 401), `file-resource-source` (file-mode admin write 409 + path-leak), `health-minimal` / `prometheus-metrics` (admin `/admin/v1/health`), `openai-sdk-compat` (RFC 9745 deprecation-header contract), `seed-vs-admin-characterization` (seed≡admin equivalence), `status-models` (`/status/models` vs `/admin/v1/models/status` equivalence).

Everything else — the large majority of the suite, which already seeds via `SeedClient`/etcd and never touches the admin listener — now runs admin-off with **no per-file change** (the default carries it).

## Scoping method

Swept every case file for real admin-listener usage (`app.adminUrl` in a request / `new AdminClient` calling anything other than `listModelStatuses`, which reads the metrics listener). The result was exactly the 9 held-back files above. `admin-disabled-e2e` and the 9 files migrated in #792 already pass `admin: false` explicitly; `status-config` seeds via etcd and only references `/admin/v1/health` in a comment, so it runs admin-off unchanged.

## Verification

`npx tsc --noEmit` clean on the changed files. The 9 held-back files pass unchanged (admin:true = their prior behavior); `allowed-models` and `status-config` confirm previously-default-admin-on files now run admin-off. Full e2e suite green with the admin-off default — the definitive check that no admin-dependent file was missed (a miss would fail now that the listener is off by default).

## Follow-up

With this merged, the remaining Admin-API-removal work (AISIX-Cloud#1007 Phase 4) is: `/readyz` relocation + read-surface decision, a v0.4.0 coexistence release, then the removal PR itself (which deletes the admin listener + the 9 `admin: true` held-back tests).

## Summary by CodeRabbit

* **Tests**
  * Updated end-to-end test setup to explicitly enable the Admin API where required.
  * Continued validating API key budgets, authentication failures, health checks, metrics, model status, and resource behavior.
* **Refactor**
  * Test applications now start with the Admin API disabled by default, matching the updated configuration.
  * Readiness checks no longer require the Admin API unless it is explicitly enabled.

### DIFF
--- a/tests/e2e/src/cases/apikey-budget-e2e.test.ts
+++ b/tests/e2e/src/cases/apikey-budget-e2e.test.ts
@@ -22,7 +22,9 @@ describe("apikey max_budget_usd e2e: standalone admin rejects removed field", ()
     etcdReachable = await new EtcdClient().ping();
     if (!etcdReachable) return;
 
-    app = await spawnApp();
+    // Held-back: this test drives the Admin API surface itself, so it
+    // keeps the admin listener bound (the suite default is now admin-off).
+    app = await spawnApp({ admin: true });
     admin = new AdminClient(app.adminUrl, app.adminKey);
   });
 
--- a/tests/e2e/src/cases/apikey-lifecycle-e2e.test.ts
+++ b/tests/e2e/src/cases/apikey-lifecycle-e2e.test.ts
@@ -54,7 +54,9 @@ describe("api key lifecycle e2e: expired/disabled keys fail closed, rotate swaps
     if (!etcdReachable) return;
 
     upstream = await startOpenAiUpstream();
-    app = await spawnApp();
+    // Held-back: this test drives the Admin API surface itself, so it
+    // keeps the admin listener bound (the suite default is now admin-off).
+    app = await spawnApp({ admin: true });
     admin = new AdminClient(app.adminUrl, app.adminKey);
     seed = new SeedClient(etcd, app.etcdPrefix);
 
--- a/tests/e2e/src/cases/auth-baseline-e2e.test.ts
+++ b/tests/e2e/src/cases/auth-baseline-e2e.test.ts
@@ -50,7 +50,9 @@ describe("auth baseline e2e: missing/malformed/unknown bearer all fail closed", 
     if (!etcdReachable) return;
 
     upstream = await startOpenAiUpstream();
-    app = await spawnApp();
+    // Held-back: this test drives the Admin API surface itself, so it
+    // keeps the admin listener bound (the suite default is now admin-off).
+    app = await spawnApp({ admin: true });
     seed = new SeedClient(etcd, app.etcdPrefix);
 
     // A single Model + ProviderKey + ApiKey configured. The valid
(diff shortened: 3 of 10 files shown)


moonming:
@coderabbitai full review

moonming:
**Independent cold audit** (2 lenses — completeness / over-marking + default correctness). **PASS, zero blocking.** Both lenses converged: the admin-off default flip is correct and complete — exactly the 9 case files that make a real admin-listener request are marked `admin: true` (every `spawnApp` site within them, including the ctx.skip-gated status-models file-mode block); the 5 unmarked `AdminClient` files only call `listModelStatuses` (metrics listener); #792-migrated + admin-disabled correctly stay admin-off. No admin-dependent file or site was left on the new admin-off default — the 150/150 green full suite confirms it.

Two LOW nits, both the **inverse of a break** (admin over-applied in the held-back `prometheus-metrics`, nothing runs admin-off-where-needed), both **fixed in `639a30a`**:
1. The `beforeAll` comment misstated why admin is bound. The real coupling is the `fetch(`${app.adminUrl}/metrics`)` → 404 assertion (proving the admin listener is bound but does not serve `/metrics`); an unbound listener would connection-refuse, not 404. Comment corrected.
2. The custom-path and `prometheus: false` auxiliary spawns carried redundant `admin: true` (they only probe the metrics port). Dropped — they now run admin-off, keeping `admin: true` only on the `beforeAll` app the 404 assertion needs.

Verification: prometheus-metrics 5/5 green after the change (the 2 auxiliary spawns pass admin-off); tsc clean; full suite was 150/150 green on the prior head.

moonming:
CodeRabbit (readiness vs `extra.admin`): **fixed in the latest commit.** `spawnApp` now throws if `overrides.extra` contains an `admin` key, so the `admin` boolean stays the single source of truth for the admin listener and readiness can't diverge from the effective config. Verified no existing test sets admin via `extra` (grep clean), and an `extra`-using test still passes.

PR #801, opened 2026-07-22
ci(release): lead draft release notes with a Get started + Download header

moonming:
## What & why

Follow-up to the v0.4.0 release: its notes were a bare auto-generated `What's Changed` list, so I hand-added a **Get started + Download** header (docs, self-hosted quickstart, `docker pull`) — the same Kong-style entry points our published releases show. This PR makes that header **automatic** for every future draft instead of a manual step.

## Changes

- **`.github/release-notes-header.md`** (new) — the header template: Documentation + self-hosted quickstart links, and a version-stamped `docker pull ghcr.io/api7/aisix:__VERSION__` block. `__VERSION__`/`__MINOR__` are substituted from the tag.
- **`.github/workflows/release-draft.yml`** — now checks out the repo, stamps the tag version into the template, appends GitHub's generated notes (via the `releases/generate-notes` API), and creates the draft with `--notes-file`. Behavior otherwise unchanged: still draft-only, still no-op if a release already exists, still flags `--prerelease` for `-rc` tags.
- **`RELEASING.md`** — step 2 updated: the header + full-changelog link are pre-filled, so polishing = inserting highlights in the middle; change the header via the template, not per release.

## Notes / accuracy

- Image tag carries **no** leading `v` (`0.4.0`), matching `docker-image.yml`'s output; the git tag stays `v0.4.0`.
- `__MINOR__` is `${VERSION%.*}` — exact for stable (`0.4.0`→`0.4`); for `-rc` tags it's approximate, which is why the header phrases rolling tags as "listed on the package page" rather than hard-claiming them.
- Verified locally: YAML lints clean, and the assembled body for `v0.4.0` renders correctly (header → `---` → What's Changed → Full Changelog).

## Test plan

- [ ] Next `vX.Y.Z` tag push produces a draft whose body leads with the Get started + Download header above What's Changed.
- [ ] `docker pull` line in the draft resolves (image tag published by `docker-image.yml`).

### DIFF
--- a/.github/release-notes-header.md
+++ b/.github/release-notes-header.md
@@ -16,3 +16,22 @@ docker pull ghcr.io/api7/aisix:__VERSION__
 Rolling tags (`__MINOR__`, `latest`) and the `docker.io/api7/aisix` mirror for private/offline deployments are listed on the [package page](https://github.com/orgs/api7/packages/container/package/aisix).
 
 ---
+
+<!-- ─────────────────────────────────────────────────────────────────────────
+     RELEASE NOTES — polish this section before publishing, then delete this
+     comment. Full house style + examples: RELEASING.md §2.
+
+     1. Headline (optional, one line) — the story of the release, if it has one.
+        e.g. "AISIX becomes a gateway for AI agents, not just LLM calls."
+     2. Highlights — 3–6 plain-language bullets, most important first.
+     3. Themed sections — keep only the ones that apply, e.g.:
+          ## 🧭 Routing              ## 🛡️ Guardrails       ## 🔭 Observability
+          ## 🔐 Security & isolation ## 🔌 API surface       ## 🧩 MCP & A2A
+        Cite PRs as (#123); describe each feature by its own function — no
+        comparisons to other products.
+     4. ## ⚠️ Breaking changes — its own section: old → new spelling + what to
+        update.
+
+     The auto-generated "What's Changed" list below is the raw commit index:
+     fold it into the sections above, then trim it or keep it as a full log.
+     ───────────────────────────────────────────────────────────────────────── -->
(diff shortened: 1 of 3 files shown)


moonming:
## Independent audit — resolved (no HIGH)

Cold audit (verified against `main` + PR head) returned **no HIGH**, two MEDIUM, two LOW. All addressed in `cfc60d5`:

- **MEDIUM — prerelease advertised a nonexistent `X.Y` tag.** `${VERSION%.*}` on `v0.5.0-rc.1` yielded `0.5.0-rc`, but `docker/metadata-action` skips `major.minor` for pre-release semver, so that tag is never pushed. Fixed: the rolling-tags sentence is now a `__ROLLING_TAGS__` template slot, branched in the workflow — stable keeps `Rolling tags (X.Y, latest)…`; prereleases say `Other image tags…` with no copy-pasteable tag that would 404. (The actionable `docker pull …:__VERSION__` line was always valid.)
- **MEDIUM — `actions/checkout@v4` off-convention.** Every other workflow pins `@v6`; aligned.
- **LOW — `generate-notes` failure aborted the step before the draft was created.** Added a fallback to a bare `## What's Changed` heading so a transient API hiccup still yields a draft with the header.
- **LOW (informational) — `sed` injection.** Confirmed safe: the `v*.*.*` trigger glob can't match `/`, and semver tags carry no `&`/`#`. No change.

Verified locally: YAML lints clean; assembled body for both `v0.4.0` and `v0.5.0-rc.1` stamps the version correctly, leaves no placeholders, keeps the scaffold comment, and (rc) advertises no bogus rolling tag.

Rebased onto @moonming's `aa204eb` (curated-notes scaffold) rather than clobbering it.

PR #802, opened 2026-07-22
test(e2e): pin the proxy health endpoints under admin-off

moonming:
## What

Pin the **proxy listener's health endpoints under admin-off** — the health surface that survives the Admin API's removal. Two e2e legs added to `admin-disabled-e2e` (etcd + file source): wait for `/readyz` to report ready, then assert `/livez` → 200 `ok` and `/readyz?verbose` lists `[+]shutdown ok` + `[+]config ok`, all with **no admin listener bound**.

## Why — a premise correction

The Admin-API-removal roadmap listed "`/readyz` relocation + deployment-probe repointing" as a prerequisite, on the premise that `/readyz` existed only on the admin listener. Investigation shows the premise is wrong:

- The **proxy listener already serves both `/livez` and `/readyz`** (`aisix-proxy` `build_router`), via the same shared `health::livez_response`/`readyz_response` the admin handlers call.
- The proxy `/readyz` carries the **full readiness semantics**: its config-freshness block is wired from the **same** `supervisor.watch_status()` the admin handler reads (etcd and managed modes), and an always-fresh probe in file mode (no watch to go stale) — `crates/aisix-server/src/main.rs` wires `with_config_apply_age` in every mode.
- A sweep of every deployment artifact — this repo's `Dockerfile` (no `HEALTHCHECK`), `docker/entrypoint.sh`, CI docker smoke (already probes the **proxy** `/livez`), and the control-plane repo's entire `deploy/` tree — found **no health probe pointing at the admin listener**. There is nothing to relocate and nothing to repoint.

So the only real gap was a **regression pin**: nothing asserted that the surviving health surface works with the admin listener off. This PR adds that pin, so the eventual admin-listener removal PR cannot silently regress `/livez`/`/readyz` on the proxy.

The operational readiness story after removal is unchanged and now fully covered: proxy `/livez` (liveness) + proxy `/readyz` (readiness incl. config freshness/drain) + metrics listener `/status/ready` (config-only readiness) — the latter already pinned by `status-config-e2e`.

## Verification

`admin-disabled-e2e` 9/9 green (7 existing + 2 new health legs); the new legs do their own readiness wait so they hold under test filtering. `npx tsc --noEmit` clean on the changed file; `health-minimal` (held-back admin health surface) untouched and green.

## Follow-up (not this PR)

The repo README still leads with "configure through the admin API on `:3001`" and points at the rewritten docs quickstart — stale against the deprecation story; belongs with the coexistence-release communication pass. Docs-side, the proxy `/livez`/`/readyz` could be listed on the ports/proxy-api reference pages as the surviving health surface.

## Summary by CodeRabbit

* **Tests**
  * Expanded end-to-end coverage for proxy health checks when the Admin listener is disabled.
  * Verified readiness and liveness endpoints across both etcd-backed and file-based configurations.
  * Confirmed readiness details report successful shutdown and configuration checks, while liveness returns the expected healthy response.

### DIFF
--- a/crates/aisix-proxy/src/lib.rs
+++ b/crates/aisix-proxy/src/lib.rs
@@ -931,6 +931,32 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn readyz_503s_when_the_config_probe_reports_no_apply_yet() {
+        // Pins the config_apply_age plumbing through the router: a wired
+        // probe reporting "no apply yet" must gate readiness with a 503.
+        // Without this, dropping the probe wiring would leave readyz
+        // reporting `[+]config ok` unconditionally (the field-None path),
+        // byte-identical to wired-and-fresh — a silent downgrade of
+        // readiness to shutdown-only that no other test would notice.
+        let hub = Arc::new(Hub::new());
+        let snap = seed_snapshot("my-gpt4", &["my-gpt4"], "http://unused");
+        let state = build_state(snap, hub).with_config_apply_age(Arc::new(|| None));
+        let app = build_router(state);
+
+        let req = Request::builder()
+            .method("GET")
+            .uri("/readyz?verbose")
+            .body(Body::empty())
+            .unwrap();
+
+        let resp = run(app, req).await;
+        assert_eq!(resp.status(), StatusCode::SERVICE_UNAVAILABLE);
+        let bytes = to_bytes(resp.into_body(), 1024).await.unwrap();
+        let text = std::str::from_utf8(&bytes).unwrap();
+        assert!(text.contains("[-]config failed: not ready"));
+    }
+
+    #[tokio::test]
     async fn health_route_is_not_found() {
         let hub = Arc::new(Hub::new());
         let snap = seed_snapshot("my-gpt4", &["my-gpt4"], "http://unused");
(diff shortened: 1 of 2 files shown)


moonming:
@coderabbitai full review

moonming:
**Independent cold audit**: PASS — all four premise-correction claims verified against source and empirically (the auditor re-ran the sweeps, including the AISIX-Cloud helm chart the PR body's sweep description didn't mention: CP-only components, no DP probe — conclusion unchanged and now repo-wide).

**MEDIUM — the pin couldn't catch the config-freshness gate becoming unwired** (field-`None` → `[+]config ok` unconditionally, byte-identical to wired-and-fresh; the likely trigger is the admin-removal PR reworking the adjacent `main.rs` region). **Addressed in `06acf84`**: a router-level unit test (`readyz_503s_when_the_config_probe_reports_no_apply_yet`) pins the plumbing — a wired probe reporting no-apply-yet must 503 with `[-]config failed: not ready`. The `main.rs` wiring itself cannot be e2e'd leak-proof (the pre-first-apply 503 window is milliseconds, and an unreachable etcd fails boot instead of serving 503); `/status/ready` on the metrics listener independently covers config readiness (status-config-e2e). The structural fix — a **fail-closed default** for `ProxyState.config_apply_age` so a dropped wiring turns readyz permanently 503 instead of permanently 200 — is filed as #803 (audit LOW, pre-existing design).

**LOW — comment overstated**: the etcd health leg's comment now says readyz reports ready after the *initial* load applies (an empty prefix counts), matching the supervisor's record-apply-on-first-cycle behavior.

Housekeeping per audit: branch is 1 commit behind main (release-notes CI only — no path overlap, no semantic conflict); CodeRabbit shows a rate-limit false-green with zero comments to triage. Verification at the new head: `aisix-proxy` readyz tests 4/4 (incl. the new pin), clippy clean, `admin-disabled-e2e` 9/9.

PR #804, opened 2026-07-23
fix(anthropic): preserve client cache_control through the OpenAI-to-Anthropic bridge

moonming:
## What

Client-supplied `cache_control` prompt-cache markers now survive the OpenAI→Anthropic bridge translation. Previously the bridge flattened every message through the concatenated-text path (`content_str()`) and rebuilt content as fresh `{"type":"text"}` blocks, silently stripping the markers a caller attached to its system prompt, message blocks, or tool definitions — the upstream cached nothing while the caller believed its caching strategy was active, paying full input price every turn.

Phase 0 of the prompt-caching work tracked in AISIX-Cloud#1110 (Gap A).

## Changes

- **`AnthropicSystem` enum** (`wire.rs`): the top-level `system` field is now `string | block-array`, per Anthropic's documented equivalence ([Messages API — system](https://docs.anthropic.com/en/api/messages)). All-plain-text system messages keep the historical `"\n\n"`-joined **string form byte-for-byte**; the block-array form is emitted only when a caller's system message itself carried typed blocks.
- **Message content blocks map 1:1**: when a caller sent array-form content, its `{type:"text", …}` blocks now map 1:1 onto the Anthropic wire — block-level fields and block *positions* ride along (a marker means "cache through here", so its position is the payload). Two per-block guarantees, preserving what the old flatten path guaranteed implicitly:
  - **Marker fidelity**: the fields Anthropic accepts on a text block (`text`, `cache_control`, `citations`) forward unchanged.
  - **Clean-block guarantee**: everything else is stripped — stray caller metadata (e.g. an OpenAI streaming `index` replayed from assembled history) would 400 at Anthropic's strict request validator, and degenerate blocks (missing or whitespace-only `text`) are rejected per-block upstream; both vanished in the old flatten, so forwarding them would break requests that worked before.

  Non-text blocks (images/audio) are still skipped — the pre-existing, documented cross-provider content limitation, unchanged. Empty survivors degrade to a single empty text block (Anthropic rejects `content: []`).
- **Tool definitions preserve `cache_control`** (`translate_openai_tools_to_anthropic`): callers attach the marker at the tool's top level or inside `function`; either spelling is forwarded. Tool definitions sit first in the prompt-cache prefix hierarchy (tools → system → messages, per the [prompt-caching doc](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching)), so a stripped tool marker disables the caller's whole cache.

**Chokepoint coverage — precise scope**: `split_system` + `build_request` serve the OpenAI-shape bridge to three upstream families — Anthropic direct, Bedrock Anthropic-invoke (`chat_anthropic`), and Vertex Anthropic `:rawPredict`/`:streamRawPredict` — so **OpenAI-shape `/v1/chat/completions` callers** routed to any of them keep their markers. Two entry points do **not** yet benefit, because their ingress flattens block structure *before* the chokepoint (pre-existing, unchanged by this PR): the `/v1/responses` input translation (`responses_bridge.rs` builds every message with `content_blocks: None`) and the native `/v1/messages` inbound parse when bridged cross-provider to Bedrock/Vertex Claude (`parse_inbound_request` rebuilds clean blocks). Both are tracked as Phase-0 follow-ups in AISIX-Cloud#1110. The native `/v1/messages` **passthrough** (direct Anthropic upstream) already preserved these fields byte-for-byte and is untouched.

## Ecosystem comparison (per repo rule 7)

Marker-stripping in the OpenAI→Anthropic transform is a recurring bug class across mainstream AI gateways: among the gateways surveyed, one shipped and fixed exactly this drop on a model-ID edge case, one still drops the caller's TTL on its transformed route (with an open issue for markers vanishing on a second provider path), and one drops markers entirely on its OpenAI-format path while supporting them elsewhere. Gateways that avoid the class do so the way this PR does: forward the caller's typed blocks verbatim instead of rebuilding content from flattened text. Our design lands on verbatim block forwarding for `type:"text"` blocks only, keeping the string-form `system` byte-stable when no blocks were sent — a divergence from gateways that always emit array form, chosen to keep existing traffic's wire bytes (and any provider-side cache prefixes built on them) unchanged.

## Testing

- `wire.rs` unit tests (7 new): system-marker preservation (pure/mixed), user/assistant block preservation incl. positions, image-only degrade invariant, tool markers (both spellings + absence), byte-stability pin for the plain-string `system` form.
- e2e (`anthropic-upstream-e2e.test.ts`, new scenario): black-box wire-shape assertion — OpenAI-shape request in, captured gateway→upstream `/v1/messages` body must carry the markers on system (block-array form), the exact marked message block (with `ttl`), and the translated tool.
- Full workspace `cargo test` + full local e2e suite green; `cargo fmt` + clippy clean.

## Deliberately out of scope

- **Injection** of gateway-authored breakpoints (that's AISIX-Cloud#1110 Phase 1 — this PR is the fidelity prerequisite).
- Markers inside `role:"tool"` (tool_result) message content — no clean OpenAI-shape spelling exists for that position; the flattened-string path there is unchanged.
- Non-text block translation (vision on this bridge) — pre-existing limitation, tracked separately.
- Bedrock Converse `cachePoint` (streaming path) — AISIX-Cloud#1110 Phase 2.
- `/v1/responses` ingress and native-`/v1/messages`-cross-provider ingress marker preservation — flattened upstream of this chokepoint; Phase-0 follow-ups in AISIX-Cloud#1110 (see coverage scope above).
- TTL validity per upstream: a forwarded `ttl:"1h"` marker reaches Bedrock-invoke/Vertex models that may support only the 5-minute tier — the marker is the caller's own request content and per-model TTL gating belongs to the Phase-1 injector (AISIX-Cloud#1110), not to pass-through fidelity.

## Audit round

An independent multi-angle cold audit (12 confirmed findings) drove three changes after the initial push: (1) the per-block whitelist + degenerate-block filter above — the initial verbatim clone would have forwarded empty text blocks and stray caller metadata that Anthropic's strict validator 400s, breaking previously-working array-form traffic; (2) e2e strengthened — exact tool-shape equality (no OpenAI wrapper leakage, `input_schema` presence), markers asserted on every turn incl. replayed-assistant blocks, message-count pin, stray-field strip, and a plain-string `system` byte-stability assertion on the live wire; (3) this coverage-scope section corrected — the initial body overclaimed `/v1/responses` benefit and omitted the Vertex paths.

## Summary by CodeRabbit

- **New Features**
  - Preserved typed content blocks and metadata when translating requests to Anthropic.
  - Improved system prompt handling to support both plain text and structured, typed block formats while keeping plain-text behavior unchanged.
  - Ensured `cache_control` markers are carried through for system content, message blocks, and translated tool/function definitions.

- **Tests**
  - Extended Anthropic upstream E2E coverage to validate structured system content, metadata/TTL preservation, safe handling when no usable text blocks remain, and correct tool translation shape.

### DIFF
--- a/crates/aisix-provider-anthropic/src/wire.rs
+++ b/crates/aisix-provider-anthropic/src/wire.rs
@@ -6,13 +6,14 @@
 //!
 //! - System prompt is a top-level `system` field, not a message with
 //!   `role: "system"` — we collapse all leading system messages into one
-//!   string and forward it there.
+//!   string (or an array of text blocks when the caller sent typed
+//!   blocks, so `cache_control` markers survive) and forward it there.
 //! - Only `user` and `assistant` roles on the wire. `tool` messages from
 //!   ChatFormat are rejected at the bridge boundary rather than being
 //!   silently re-classified.
 //! - Content is an array of blocks — we emit a single `{"type":"text",…}`
-//!   block per message and read the concatenation of text blocks on the
-//!   way back.
+//!   block per message (typed caller blocks forward their text blocks
+//!   verbatim) and read the concatenation of text blocks on the way back.
 //! - `max_tokens` is required by Anthropic. We default to a safe ceiling
 //!   when the client didn't set one, but log the fallback so operators
 //!   can tune the default if desired.
@@ -30,13 +31,30 @@ use serde::{Deserialize, Serialize};
 /// enough that a runaway prompt doesn't burn tokens silently.
 pub const DEFAULT_MAX_TOKENS: u32 = 4096;
 
+/// The top-level `system` field — Anthropic accepts a plain string or an
+/// array of text content blocks, documented as equivalent shapes
+/// (<https://docs.anthropic.com/en/api/messages#parameter-system>).
+///
+/// The gateway emits the string form whenever every system message came
+/// in as plain text (byte-identical to what it has always sent), and
+/// switches to the block-array form only when a caller's system message
+/// itself carried typed blocks — whose block-level fields (notably
+/// `cache_control` prompt-cache markers) must survive translation
+/// (AISIX-Cloud#1110 Gap A).
+#[derive(Debug, Clone, PartialEq, Serialize)]
+#[serde(untagged)]
+pub enum AnthropicSystem {
+    Text(String),
+    Blocks(Vec<serde_json::Value>),
+}
+
 #[derive(Debug, Clone, Serialize)]
 pub struct AnthropicRequest<'a> {
     pub model: &'a str,
     pub messages: Vec<AnthropicMessage<'a>>,
     pub max_tokens: u32,
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub system: Option<String>,
+    pub system: Option<AnthropicSystem>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub temperature: Option<f32>,
     #[serde(skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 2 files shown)


moonming:
## Independent cold-audit triage (merge gate)

A 6-angle cold audit (each finding adversarially re-verified against the head SHA) produced 12 confirmed findings (5 rejected as false positives / out-of-scope). Triage below; all MEDIUMs are either fixed in `ad9a51f` or explicitly justified.

### Fixed in `ad9a51f`

- **MEDIUM (correctness+reliability+breaking, 3 findings, same root):** verbatim block clone forwarded empty/whitespace-only text blocks and stray caller metadata (e.g. a replayed OpenAI streaming `index`) that Anthropic's strict validator 400s per-block — breaking array-form requests that worked before the flatten was removed. → `anthropic_text_blocks` now whitelists `{type,text,cache_control,citations}` and drops degenerate text blocks; 2 new unit tests + e2e stray-`index` assertion.
- **MEDIUM (e2e, 2 findings):** weak tool assertion (`toMatchObject` let an untranslated OpenAI wrapper or dropped `input_schema` pass) and single-message coverage (marker-on-first-turn-only and the assistant-blocks arm were unit-only). → e2e now asserts exact tool shape via `toEqual` (`{name,input_schema,cache_control}`, no wrapper), markers on user **and** replayed-assistant turns, and a message-count pin.
- **LOW (e2e):** plain-string `system` byte-stability was pinned only at unit level. → e2e now asserts `sentBody.system === "you are terse"` on the live wire.

### Fixed by PR-body correction (`gh pr edit`)

- **MEDIUM (breaking) + LOW (correctness) + MEDIUM (e2e), same root:** the body overclaimed `/v1/responses` coverage — that ingress flattens with `content_blocks: None`, so markers can't reach the chokepoint. And it omitted the Vertex Anthropic `:rawPredict`/`:streamRawPredict` paths, which the chokepoint **does** serve. → Coverage-scope section rewritten: OpenAI-shape `/v1/chat/completions` → {Anthropic, Bedrock-invoke, Vertex} covered; `/v1/responses` ingress + native-`/v1/messages`-cross-provider ingress explicitly listed as not-yet-covered Phase-0 follow-ups (`AISIX-Cloud#1110`).
- **LOW (reliability):** native `/v1/messages` → Bedrock/Vertex Claude loses markers in the inbound parse. → Same follow-ups filed in `AISIX-Cloud#1110`; passthrough-to-direct-Anthropic is unaffected and correctly claimed.

### Justified — deferred with tracking (not blocking)

- **LOW (breaking): forwarded `ttl:"1h"` may exceed a Bedrock-invoke model's cache tier.** The marker is the **caller's own request content**, not gateway-authored; pass-through fidelity should forward what the caller sent. Per-model TTL gating belongs to the Phase-1 injector, tracked in `AISIX-Cloud#1110` (now noted there). Pre-PR the marker was stripped, hiding the mismatch; forwarding it surfaces a caller error honestly rather than silently no-op'ing.

### Rejected false positives (recorded for the reviewer)

- **HIGH (security): decoupled `content_blocks` guardrail bypass.** Verifier confirmed the mechanics but ruled it **pre-existing and systemic, not caused by this PR** — the OpenAI-compat bridge already forwards `content_blocks` verbatim with identical logic, so an attacker's capability is unchanged; the root cause + fix live in `aisix-guardrails` (untouched here). Filed separately as **`AISIX-Cloud#1123`** (security), fix scoped to `message_scan_text`/`redact.rs`. Not a blocker for this PR.
- 4 others (Role::Tool markers, Bedrock-Converse streaming drop, `user_turn` fallback, arbitrary-sibling-field forwarding) rejected as either explicitly-out-of-scope-in-body or superseded by the `ad9a51f` whitelist.

CI + full local e2e re-run green after `ad9a51f`.

Issue #449, opened 2026-05-29
security: enforce model ACLs for routing targets and passthrough providers


juzhiyuan:
## Summary

Routing and passthrough currently allow authenticated callers to reach provider credentials or target models outside their API key model ACL. The fix should make passthrough opt-in/allowlisted and authorize every concrete model/provider target, not just the user-supplied router or provider prefix.

## Source

- Imported from Codex Security scan results shared in Slock `#Dev:bc985f48` (`chatgpt.com.har`).
- This issue groups findings with the same engineering boundary to avoid fragmented fixes.
- Triage priority: **P0**.

## Covered scanner findings

- [ ] **#19** `Jina models leak into passthrough without ACL` — Any valid API key can use configured Jina credentials through generic passthrough if `api_base` exists.
- [ ] **#20** `Cohere credentials exposed through passthrough` — Cohere rerank provider credentials become usable through generic passthrough.
- [ ] **#27** `Passthrough bypasses model ACLs` — Generic `/passthrough/:provider/*rest` selects a provider credential without checking the authenticated key can access that model/provider.
- [ ] **#29** `Routing models bypass target model ACL checks` — Access to a virtual router can indirectly invoke target models absent from `allowed_models`.

## Expected fix direction

- [ ] Require API-key authorization against every resolved routing target model before dispatch.
- [ ] Do not treat access to a virtual router as implicit access to all target models.
- [ ] Disable passthrough by default for providers unless explicitly configured as an allowed provider/operation.
- [ ] For passthrough, check the authenticated API key against the selected provider/model/resource boundary before injecting credentials.
- [ ] Do not select credentials via “first configured model with matching provider”; use explicit routing/allowlist metadata.
- [ ] Add regression tests showing low-privilege keys cannot reach premium target models or unrelated provider passthrough credentials.

## Definition of done

- [ ] Fix is covered by regression tests for the affected mode/provider/endpoint.
- [ ] The fix is applied at the shared boundary where possible, not only at one provider call site.
- [ ] Security behavior is documented in code comments or docs where operator expectations change.

PR #805, opened 2026-07-23
fix(passthrough): enforce the body model's rate limit on the raw tunnel

moonming:
## What

`/passthrough/:provider/*rest` now enforces the **model-level** rate-limit layers (a Model's inline `rate_limit` plus `model`-scope rate-limit policies) for JSON bodies whose top-level `model` field names a configured Model of the addressed provider.

Tracking: `api7/AISIX-Cloud#1116` (field report: a video model's configured rate limit was silently ignored on passthrough).

## Why

Pre-fix, the tunnel called `quota::enforce(&state, auth, None)` — only the API-key / team / member layers applied. For provider endpoints with no typed surface (video generation being the reported case), the model cap configured in the dashboard was unenforceable **anywhere in the product**, and nothing surfaced that: the config was accepted and silently inert.

## How

- `body_model_rate_limit()` in `passthrough.rs`: best-effort parse of the buffered request body (already in memory for the input-guardrail scan), exact `display_name` lookup constrained to the addressed provider, then the existing `ModelRateLimit::from_model` → `quota::enforce(Some(&model_rl))` path used by every typed endpoint. Same bucket keys as typed traffic, so passthrough and typed calls to the same model share one budget.
- Fallbacks preserve pre-fix behavior exactly: non-JSON body, no `model` field, unregistered name, or a same-named Model of a *different* provider → request-level layers only, no error.
- The tunnel still parses no usage (tokens stay 0), so only request-count dimensions (rps/rpm/rph) draw from the model buckets — token-per-minute caps remain inert on passthrough, consistent with raw tunnels generally (documented in the tracking issue; docs follow-up noted there).

## Ecosystem comparison (per repo rule 7)

- The provider-native envelope carries the target model in a top-level `model` field — the shape shared by OpenAI-compatible bodies and e.g. Alibaba Model Studio's synthesis APIs (<https://help.aliyun.com/zh/model-studio/text-to-image-v2-api-reference>).
- Established gateways that run a pre-call policy hook on their passthrough routes resolve the target model the same way — from the request body's `model` field — and apply per-model request caps when the name matches a registered model; identity-level (key/team) limits apply on all routes regardless. None skip the model layer wholesale for lack of a "resolved model".
- Raw-tunnel token counting is universally degraded across gateways (post-hoc or next-request accounting at best); this PR deliberately does not attempt it.

## Scope notes

- `a2a.rs` / `mcp.rs` also pass `None`, but have no model concept — intentionally untouched.
- The issue's optional step 4 (preferring the body-named Model as the credential-lending target) is **deferred**; tracked in `api7/AISIX-Cloud#1116`. This PR changes rate limiting only — credential resolution and the #449 ACL semantics are untouched.
- No CP work needed: model `rate_limit` and `model`-scope policies are existing CP surfaces; this is a DP enforcement gap only.

## Audit remediation (cold audit: MERGE-WITH-CHANGES, 3 MEDIUM / 0 HIGH)

- **MEDIUM-1 fixed** — body `model` now also matches the provider-native `model_name` within the addressed provider (exact `display_name` first); the bucket is keyed by the entry's `display_name` either way, so tunnel and typed traffic share one budget. `min_by_key(id)` keeps duplicate-`model_name` picks deterministic.
- **MEDIUM-2 fixed** — full `serde_json::Value` DOM replaced by a `#[derive(Deserialize)]` field probe; unknown fields skip without allocation.
- **MEDIUM-3 fixed** — new unit test pins a `model`-scope `RateLimitPolicy` row gating the tunnel (no inline `rate_limit`).
- LOW-1 fixed (cross-provider same-name test), LOW-3(c) fixed (e2e asserts the 429 produced no upstream round-trip). LOW-2 (typed+tunnel shared-bucket cross-call test) not added: the sharing is now pinned structurally — both surfaces key by `display_name` through the same `ModelRateLimit::from_model` — and the alias test covers the bucket-key path.

**Accepted residuals (by design, per the fallback contract):** (a) a caller that omits or misnames the body `model` field draws only the key/team layers — model caps on the raw tunnel bind well-behaved clients; the polling leg of async APIs is intentionally exempt. (b) A key not ACL'd for model X can still drain X's shared bucket through the tunnel, because credential borrowing picks the first accessible model and the body model is not ACL-checked — pre-existing behavior, tracked with the deferred credential-resolution step in `api7/AISIX-Cloud#1116`.

## Tests

- Unit (`crates/aisix-proxy/src/lib.rs`):
  - `passthrough_enforces_model_rate_limit_from_body_model_field` — rpm=1 model, second tunnel call 429 `rate_limit_exceeded`.
  - `passthrough_unregistered_or_absent_body_model_keeps_key_layer_only` — unregistered name and non-JSON body pass; key-level cap still gates (pre-fix behavior pinned).
- E2E (`tests/e2e/src/cases/passthrough-model-rate-limit-e2e.test.ts`, source-blind, real gateway binary + etcd + mock upstream):
  - registered body model: 2nd call 429 with `Retry-After` and standard envelope;
  - unregistered body model: repeated calls keep flowing.
- Full `aisix-proxy` suite: 660 passed. Existing passthrough e2e family: green (one unrelated `/v1/responses` least-busy case flaked with ECONNRESET in a parallel run and passes alone; it does not traverse the changed code).

### DIFF
--- a/crates/aisix-proxy/src/lib.rs
+++ b/crates/aisix-proxy/src/lib.rs
@@ -2565,6 +2565,152 @@ data: [DONE]\n\n"
         assert_eq!(v["error"]["type"], "rate_limit_exceeded");
     }
 
+    // ---- regression coverage for api7/AISIX-Cloud#1116 --------------
+    // Pre-fix the passthrough tunnel passed `None` to quota::enforce, so
+    // a Model's inline rate_limit (and model-scope policies) never
+    // applied to passthrough traffic — for provider endpoints with no
+    // typed surface (e.g. video generation) the model limit was
+    // unenforceable everywhere. Post-fix the top-level `model` field of
+    // a JSON passthrough body is matched against the addressed
+    // provider's configured Models and its limits reserved like the
+    // typed endpoints.
+
+    fn model_entry_with_rate_limit(
+        name: &str,
+        rate_limit: serde_json::Value,
+    ) -> ResourceEntry<Model> {
+        let cfg = format!(
+            r#"{{
+                "display_name": "{name}",
+                "provider": "openai",
+                "model_name": "gpt-4o",
+                "provider_key_id": "{PK_ID}",
+                "rate_limit": {rate_limit}
+            }}"#
+        );
+        let model: Model = serde_json::from_str(&cfg).unwrap();
+        ResourceEntry::new("model-id-1", model, 1)
+    }
+
+    #[tokio::test]
+    async fn passthrough_enforces_model_rate_limit_from_body_model_field() {
+        let upstream = MockServer::start().await;
+        Mock::given(method("POST"))
+            .and(path(
+                "/api/v1/services/aigc/video-generation/video-synthesis",
+            ))
+            .respond_with(ResponseTemplate::new(200).set_body_json(serde_json::json!({
+                "output": {"task_id": "t-1", "task_status": "PENDING"}
+            })))
+            .mount(&upstream)
+            .await;
+
+        let hub = Arc::new(Hub::new());
+        hub.register_specialized("openai", Arc::new(openai_test_bridge()));
+        let snap = new_snap(&upstream.uri());
+        snap.models.insert(model_entry_with_rate_limit(
+            "my-video-model",
+            serde_json::json!({"rpm": 1}),
+        ));
+        snap.apikeys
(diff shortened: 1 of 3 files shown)


moonming:
@coderabbitai review

PR #807, opened 2026-07-23
fix(routing): honour routing.retries on the streaming chat path

jarvis9443:
## Problem

`/v1/chat/completions` dispatches streaming and non-streaming requests through two separately-written loops in `crates/aisix-proxy/src/chat.rs`.

The non-streaming loop wraps each target in `for attempt_idx in 0..=retries` with exponential backoff. The streaming loop never read `routing.retries` at all — it walked `attempt_models` once and, on a retryable failure, went straight to the next target:

```
target A initial fails -> fallback to target B     (actual)
target A initial fails -> A retry #1 -> fallback   (configured)
```

With a **single** target there is nothing to fall over to, so one transient upstream error failed the request outright — `retries` was the only thing that could have saved it.

`/v1/messages` and `/v1/responses` don't have this gap: both run streaming and non-streaming through one loop (`dispatch_to_target` branches internally) that already honours `retries`. The streaming chat path was the only one that had drifted.

## Change

Nest the same-target retry loop inside the streaming target walk, mirroring the non-streaming loop:

- read `routing.retries` (default `0`, so behaviour is unchanged for anyone not configuring it);
- retryable failure → back off (`retry_backoff`, exponential + jitter) and re-hit the same target until `retries` is exhausted, then fall over;
- non-retryable failure → stop immediately, as before;
- target-invariant setup (`model_arc` / `pk_arc` / `BridgeContext` / stream budget) hoists above the retry loop; the per-attempt rate-limit reservation, telemetry record, and dispatch stay inside it.

## Behaviour change

Per-attempt telemetry now emits a `retry`-kind record for each extra same-target attempt, so the Logs view shows `initial → retry #1 → fallback` instead of `initial → fallback`. The attempt-failed warning gains `target_attempt`, matching the non-streaming log line.

Only requests on a routing model with `retries > 0` are affected. `retries` is read from the routing block, so a direct (non-group) model is still a single attempt.

## Tests

- `streaming_routing_honors_same_target_retries` — multi-target group, `retries=1`, always-502 primary: asserts the primary is hit twice and the attempt sequence is `initial → retry → fallback`.
- `streaming_routing_retries_single_target_before_failing` — single target, `retries=2`: asserts three attempts all on that target before the request fails.
- `retry-backoff-e2e.test.ts` gains a `stream: true` case asserting the same upstream hit count (3) and backoff floor as the existing non-streaming case.

All three fail before this change and pass after. Full `aisix-proxy` suite: 662 passed.

Fixes api7/AISIX-Cloud#1119
Fixes api7/AISIX-Cloud#1122

## Summary by CodeRabbit

* **New Features**
  * Streaming requests now honor configured per-model retry counts before failing over to another target.
  * Retry attempts use exponential backoff and preserve accurate attempt and usage tracking.

* **Bug Fixes**
  * Streaming timeouts are applied consistently when connecting and reading response chunks.
  * Retry behavior now correctly handles single-target and fallback scenarios.

### DIFF
--- a/crates/aisix-proxy/src/chat.rs
+++ b/crates/aisix-proxy/src/chat.rs
@@ -1155,6 +1155,19 @@ async fn dispatch(
     // the remaining chunks, so a mid-stream stall terminates the response
     // (no fallback once the 200 is committed).
     if req.is_streaming() {
+        // `routing.retries` — how many times to re-hit the SAME target (with
+        // backoff) on a retryable failure before failing over to the next one.
+        // Read here for the same reason the non-streaming loop below reads it:
+        // the two loops are written separately, and streaming used to skip
+        // `retries` entirely (AISIX-Cloud#1119), so a retryable first-chunk
+        // failure fell straight over to the next target — or, with a single
+        // target, failed the request outright.
+        let retries = virtual_entry
+            .value
+            .routing
+            .as_ref()
+            .map(|r| r.retries_or_default())
+            .unwrap_or(0);
         let retry_on_429 = virtual_entry
             .value
             .routing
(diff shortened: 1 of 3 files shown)


jarvis9443:
@coderabbitai review

jarvis9443:
@coderabbitai full review

jarvis9443:
@coderabbitai full review

PR #809, opened 2026-07-23
fix(azure): apply stream_timeout per chunk, not as a whole-response ceiling

jarvis9443:
## Problem

`Model::stream_timeout` is defined as *"Maximum gap in milliseconds between upstream streaming chunks"* (`crates/aisix-core/src/models/model.rs`). The proxy's `with_read_timeout` wrapper and the OpenAI bridge both implement it that way — the budget bounds each `next()` and resets after every successful read.

The Azure bridge did something different:

```rust
let stream_deadline = ctx.deadline.map(|d| started + d);   // one absolute instant
...
tokio::time::timeout_at(d.into(), stream.next())           // same instant, every chunk
```

Computing the instant once and reusing it turns the budget into a ceiling on the **entire response**. A long but perfectly healthy stream was cut off with `BridgeError::Timeout` as soon as its total duration passed one gap's budget — so the more output a model produced, the more likely it was to be killed mid-answer. Reasoning models and long completions are the worst case.

This is not limited to deployments that set `stream_timeout`: `stream_timeout_effective()` falls back to `timeout`, so a model configured with only `timeout: 30000` had its Azure streams capped at 30s of total wall-clock.

## Fix

Pass the budget as a `Duration` and use `tokio::time::timeout(d, …)`, which restarts on every chunk. `elapsed_ms` on the timeout error now reports the gap budget, matching what `with_read_timeout` reports on the same condition.

The Audit H2 protection this code was added for is unchanged — a hung upstream body still can't wedge the connection after headers arrive, because a single gap exceeding the budget still fails.

## Test

`stream_budget_bounds_each_chunk_gap_not_the_whole_response` drives `build_chunk_stream` with three chunks 80ms apart under a 250ms budget: every gap is comfortably inside the budget while the ~320ms total is past it, so the semantics are distinguishable. The whole stream must be delivered.

Verified it fails under the old absolute deadline (`Timeout` partway) and passes per-chunk.

The existing `chat_stream_enforces_per_chunk_deadline` test uses wiremock's `set_delay`, which delays the response as a whole and so can't tell the two semantics apart — it passes either way. No E2E case is added: reproducing this needs precise per-chunk delivery timing from the upstream, which the unit test drives directly and an E2E mock would only approximate.

Fixes api7/AISIX-Cloud#1122

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Azure OpenAI streaming reliability by applying the timeout separately between streamed chunks.
  * Healthy streams with multiple gradual responses can now continue successfully, even when their total duration exceeds the timeout window.
  * Timeout errors now accurately reflect the configured wait period between chunks.

### DIFF
--- a/crates/aisix-provider-azure-openai/Cargo.toml
+++ b/crates/aisix-provider-azure-openai/Cargo.toml
@@ -30,5 +30,9 @@ bytes.workspace = true
 http.workspace = true
 
 [dev-dependencies]
-tokio = { workspace = true, features = ["macros", "rt", "time"] }
+# `net` + `io-util` back the hand-rolled SSE upstream in
+# `chat_stream_delivers_a_long_stream_whose_gaps_stay_within_budget`, which
+# needs to emit frames on a schedule — wiremock can only delay a response as
+# a whole, so it cannot distinguish per-chunk from total-duration timeouts.
+tokio = { workspace = true, features = ["macros", "rt", "time", "net", "io-util"] }
 wiremock.workspace = true
--- a/crates/aisix-provider-azure-openai/src/bridge.rs
+++ b/crates/aisix-provider-azure-openai/src/bridge.rs
@@ -2067,6 +2067,124 @@ mod tests {
         );
     }
 
+    /// The same contract as the test above, but driven through the public
+    /// `chat_stream` entry point instead of `build_chunk_stream`, so it
+    /// also covers `chat_stream` actually forwarding `ctx.deadline` — the
+    /// helper-level test alone would still pass if that wiring were
+    /// dropped.
+    ///
+    /// Needs an upstream that emits frames on a schedule, which wiremock
+    /// cannot do (`set_delay` delays the response as a whole, which is
+    /// exactly the distinction under test), so this hand-rolls a one-shot
+    /// SSE server: no `content-length`, `connection: close`, body framed
+    /// by EOF.
+    #[tokio::test]
+    async fn chat_stream_delivers_a_long_stream_whose_gaps_stay_within_budget() {
+        use tokio::io::{AsyncReadExt as _, AsyncWriteExt as _};
+
+        let gap = Duration::from_millis(80);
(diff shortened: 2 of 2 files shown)


jarvis9443:
@coderabbitai review

jarvis9443:
@coderabbitai full review

jarvis9443:
@coderabbitai full review

jarvis9443:
Valid point, addressed in 0b82b50 — with one correction to the proposed shape.

You're right that the helper-level test cannot see whether `chat_stream` forwards `ctx.deadline`. But the suggested test doesn't close that gap either: "gaps within budget, total beyond it, all content arrives" passes just as well when *no* deadline is forwarded, because then nothing can time out. It pins the per-chunk semantics, not the wiring.

Closing it needs the opposite assertion, so both are now there:

- `chat_stream_delivers_a_long_stream_whose_gaps_stay_within_budget` — the shape you suggested, driving per-chunk semantics through the public entry point.
- `chat_stream_times_out_when_one_gap_exceeds_the_budget` — headers flush immediately so the connect-phase `with_deadline` is already satisfied, and only then does a single gap exceed the budget. The per-chunk timeout is the only thing that can fire, which makes this the test that fails if the wiring is dropped.

Verified by mutation rather than by inspection, and the two failure modes turn out to be caught by different tests:

| mutation | helper test | delivery test | timeout test |
|---|---|---|---|
| absolute deadline (the original bug) | **FAIL** | **FAIL** | pass |
| `ctx.deadline` not forwarded | pass | pass | **FAIL** |

The middle column is the empirical version of the point above — the suggested test alone would not have caught the regression it was aimed at.

One implementation note: both need an upstream that emits frames on a schedule, which wiremock cannot express (`set_delay` delays the response as a whole, which is exactly the distinction under test). They hand-roll a one-shot SSE server instead — no `content-length`, `connection: close`, body framed by EOF — which is why tokio's `net` + `io-util` features join dev-dependencies.

PR #810, opened 2026-07-23
feat(anthropic): auto-inject prompt-cache breakpoints when enabled on a model

moonming:
## What

When a direct Anthropic model has `auto_prompt_caching` enabled, the gateway now **injects prompt-cache breakpoints** into requests that carry none of their own — one on the last system block, one on the last content block of the final message. Callers get provider-side prompt-cache discounts (0.1× input on cache reads) and faster TTFT with **zero client changes**. Callers that set their own `cache_control` markers are forwarded unchanged (stand-down).

Phase 1 (DP half) of AISIX-Cloud#1110. Builds on #804 (Phase 0 — marker preservation), which this PR is stacked on.

## Why two breakpoints, why these positions

Per the Anthropic prompt-cache prefix hierarchy (tools → system → messages, [docs](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching)):
- **Last system block** → caches the stable tools+system prefix.
- **Last content block of the final message** → caches the whole conversation prefix; because it rides the final turn, it advances as the conversation grows, so each turn re-caches incrementally.

This is the convergent strategy across the mainstream gateways that auto-inject (a system+trailing pair, trailing marker advancing with the dialogue). We inject at most two markers and only into a request that had zero, so Anthropic's 4-breakpoint cap can never be breached.

## Scope — direct Anthropic only

Injection is applied **only in the Anthropic bridge** (`chat` + `chat_stream`), not in the shared `build_request` — so the Bedrock-invoke and Vertex `:rawPredict` paths (which also call `build_request`) do **not** inject. That's deliberate: gateway-authored markers on Bedrock/Vertex carry provider-specific hazards (per-model 5m-only TTL tiers on Bedrock; Vertex's per-project "explicit caching disabled → reject" mode) that pass-through fidelity doesn't. Bedrock injection is AISIX-Cloud#1110 Phase 2; Vertex is deferred there. Marker **preservation** (Phase 0, #804) already covers all three.

## Changes

**DP config (`aisix-core`)**
- `Model.auto_prompt_caching: Option<AutoPromptCaching>` — nested `{ enabled: bool, ttl: Option<CacheTtl> }`, following the `CooldownConfig` house pattern. `CacheTtl` = `5m | 1h` (default `5m`). Regenerated `schemas/resources/model.schema.json`.

**DP injection (`aisix-provider-anthropic`)**
- `inject_cache_breakpoints(req, ttl)` in `wire.rs`: stand-down scan (any `cache_control` on system/messages/tools → inject nothing), else mark the last system block (promoting a string-form system to a one-block array so the marker has somewhere to live) and the last content block of the final message. `5m` emits the bare `{"type":"ephemeral"}` Anthropic treats as the 5-minute default; `1h` carries the explicit `ttl`.
- `maybe_inject_cache_breakpoints(body, ctx)` in `bridge.rs`: reads `ctx.model.auto_prompt_caching`, gates on `enabled`, shared by the streaming and non-streaming paths so they can't drift.

No min-token gate: below-minimum-*length* prompts are a documented silent no-op upstream (no error, normal billing, both cache counters 0), so speculative injection on short prompts is harmless. An *empty* text block is a different case — Anthropic 400s on `cache_control` attached to one — so injection skips empty/whitespace-only blocks (reachable via the image-only-turn degrade shape and a blank system prompt); see `is_empty_text_block`.

## Billing / observability — already wired

No usage-plumbing change needed: the response path already parses `cache_creation_input_tokens` / `cache_read_input_tokens` (`wire.rs`), and the control plane already prices them at distinct `cache_read` / `cache_write` rates — so budgets, spend reports, and the dashboard reflect injected-cache economics for free.

## Paired CP work (required before this is user-reachable) — AISIX-Cloud#1110

Per the repo's config-knob rule, this DP PR is **one half**. The `auto_prompt_caching` knob is not reachable until the control plane exposes it: `cp-admin.yaml` schema + Go model/validation/kine projection + dashboard form + en/zh i18n. That is a **paired CP PR**, tracked in AISIX-Cloud#1110.

**Rollout ordering (blocking constraint):** DP `Model` is `#[serde(deny_unknown_fields)]`, so this DP PR — which teaches the DP to *read* `auto_prompt_caching` — **must deploy before** the CP PR that *writes* it. If the CP shipped the field first, any model carrying it would be silently dropped from the DP snapshot and 404. This PR is the safe-to-land-first half by construction.

## Testing

- `wire.rs` unit tests (9): system+trailing marking, 1h ttl, no-system case, stand-down on a client message marker, stand-down on a client tool marker, in-place marking of a client block-form system, **empty-text final block skipped, blank system skipped, only-last-of-multi-block marked**.
- `bridge.rs` tests (3): non-streaming injects when enabled, **streaming injects when enabled** (captures the outbound body, closing the chat_stream coverage gap), does-not-inject when `enabled:false`.
- e2e (`anthropic-upstream-e2e.test.ts`, 2 new): black-box wire-shape — an `auto_prompt_caching`-enabled model injects markers on system (block-array form) + final message at the configured `1h` ttl with earlier turns clean and a message-count pin; and a stand-down case where a caller-marked request comes out with exactly its own single marker.
- Full workspace `cargo test` + Phase-0 e2e green (injection e2e run 3×, stable); `cargo fmt` + clippy clean.

## Audit round

Independent 6-angle cold audit (5 confirmed / 2 rejected), all addressed in the follow-up commit:
- **HIGH (spec):** injection stamped `cache_control` onto an empty text block (image-only/caption-less final turn, or blank system) → Anthropic 400. Fixed with `is_empty_text_block` guard on both injection sites + 2 unit tests.
- **MEDIUM ×2 (reliability + e2e):** `chat_stream` injection had zero test coverage. Added a bridge-level streaming test capturing the outbound body.
- **MEDIUM (e2e):** only-last-block-marked was asserted only against single-block final messages. Added a 3-block unit test.
- **LOW (e2e):** `enabled:false` gate untested. Added a bridge test.
- **Rejected:** an "empty block is a silent no-op not a 400" LOW (superseded — the guard is correct either way) and a default-on cache-tenancy side-channel LOW (feature is opt-in default-off; the provider-credential cache boundary is already noted in AISIX-Cloud#1110 for the customer docs).

## Deliberately out of scope

- CP knob exposure — paired PR, AISIX-Cloud#1110 (see above).
- Bedrock (Phase 2) / Vertex (deferred) injection — AISIX-Cloud#1110.
- Native `/v1/messages` passthrough injection: those callers are cache-aware and send their own markers (→ stand-down), so injection targets the OpenAI-shape callers who don't; tracked in AISIX-Cloud#1110 if a gap emerges.
- The newer `usage.cache_creation.{ephemeral_5m,ephemeral_1h}_input_tokens` breakdown (verifies 5m-vs-1h write billing) — AISIX-Cloud#1110 follow-up.

### DIFF
--- a/crates/aisix-core/src/models/model.rs
+++ b/crates/aisix-core/src/models/model.rs
@@ -158,6 +158,47 @@ impl CooldownConfig {
     }
 }
 
+/// Cache lifetime for gateway-injected prompt-cache breakpoints.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, schemars::JsonSchema)]
+pub enum CacheTtl {
+    #[serde(rename = "5m")]
+    FiveMinutes,
+    #[serde(rename = "1h")]
+    OneHour,
+}
+
+impl CacheTtl {
+    /// The value emitted on the upstream `cache_control.ttl` field.
+    pub fn as_wire_str(self) -> &'static str {
+        match self {
+            CacheTtl::FiveMinutes => "5m",
+            CacheTtl::OneHour => "1h",
+        }
+    }
+}
+
+/// Automatic prompt-cache breakpoint injection for a direct Anthropic
+/// model. When enabled, the gateway adds cache-control markers to
+/// requests that carry none of their own, so callers get provider-side
+/// prompt-cache discounts without changing their requests. Requests that
+/// already set their own cache-control markers are forwarded unchanged.
+#[derive(Debug, Clone, Serialize, Deserialize, schemars::JsonSchema, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
+pub struct AutoPromptCaching {
+    /// Whether automatic prompt-cache injection is active for this model.
+    pub enabled: bool,
+    /// Cache lifetime for injected breakpoints: `5m` (default when omitted) or `1h`. A `1h` cache write costs 2x the base input rate versus 1.25x for `5m`, so it pays off only when the cached prefix is reused across a longer session.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub ttl: Option<CacheTtl>,
+}
+
+impl AutoPromptCaching {
+    /// Effective TTL — operator override or the built-in 5-minute default.
+    pub fn ttl_or_default(&self) -> CacheTtl {
+        self.ttl.unwrap_or(CacheTtl::FiveMinutes)
+    }
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize, schemars::JsonSchema)]
 #[serde(deny_unknown_fields)]
 pub struct Model {
@@ -237,6 +278,10 @@ pub struct Model {
     /// Direct-model-only request-path cooldown configuration. Omit this field to use the built-in cooldown behavior.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub cooldown: Option<CooldownConfig>,
+
+    /// Automatic prompt-cache breakpoint injection for direct Anthropic models. Omit to leave injection off.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub auto_prompt_caching: Option<AutoPromptCaching>,
 
     /// Non-schema runtime id. Not part of the JSON payload — filled in by
     /// the snapshot loader from the etcd key path. Kept here so `Resource`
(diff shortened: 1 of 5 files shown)


moonming:
@coderabbitai review

moonming:
Merging: full CI green (build, e2e, rust unit, coverage ≥90%, lint, schema drift, generate-openapi). The independent 6-angle cold audit (see the triage above) is the substantive review gate here and is fully satisfied — it caught and fixed a HIGH (cache_control on an empty text block → Anthropic 400). CodeRabbit could not run (plan rate-limited; its check is a false-green, not a review), so it is not standing in as the review — the cold audit is.

Issue #465, opened 2026-06-01
guardrails: text in content_blocks bypasses moderation (only flat content is scanned)


moonming:
## Summary

Guardrail input/output moderation reads only the flat `ChatMessage.content` string. A message whose text lives in the structured `content_blocks` array (with an empty top-level `content`) bypasses moderation entirely.

This affects every text-scanning guardrail — `kind=keyword`, `kind=azure_content_safety` (P1), and `kind=azure_content_safety_text_moderation` (P2) — plus the bedrock dispatcher's input collection. It is a **pre-existing** shape limitation, not introduced by any single PR; surfaced by the PR2 self-audit of #461.

## Evidence

- `crates/aisix-guardrails/src/text_moderation.rs` `collect_input_text` — reads `m.content` only.
- `crates/aisix-guardrails/src/prompt_shield.rs` `collect_input_text` — same.
- `ChatMessage.content_blocks` (`crates/aisix-gateway/src/chat.rs`) holds the structured content array that bridges forward verbatim; the guardrail layer never inspects it.

## Impact

A caller that sends `content: [{ "type": "text", "text": "..." }]` (OpenAI content-block form) with no top-level string `content` gets empty scanned text → the guardrail allows the request. A moderation bypass for the content-block request shape.

## Options

- Add a shared helper (used by all guardrail dispatchers) that, when `content` is empty, concatenates the `text`-type entries from `content_blocks` into the scanned text.
- Scope decision: text-type blocks now; image/audio block moderation is a separate multimodal feature.

Severity: LOW–MEDIUM — a real bypass, but only for the content-block shape; most traffic uses the flat string today.

PR #815, opened 2026-07-24
fix(guardrails): scan the union of content and content_blocks

moonming:
## What

Fixes an input-guardrail bypass: the guardrail scan read only a message's flattened `content` string, while the provider bridges forward the raw `content_blocks` array upstream when it is present. A caller could put benign text in `content` and a real payload in `content_blocks`, slipping the payload past **every** input guardrail while the model still received it.

The scan now covers the **union** of `content` and the text of `content_blocks`, so the scanned text is a superset of everything a bridge can forward.

## The bug

`content` and `content_blocks` are independent wire fields on a message (no reconciliation — a client can send both with different text, and the parse deliberately prefers a client-supplied `content_blocks`). The shared scan helper `message_scan_text` read `content` and only fell back to `content_blocks` when `content` was empty. The bridges do the inverse: the OpenAI-compat and Anthropic bridges forward `content_blocks` verbatim when present. So:

```json
{
  "model": "guardrailed-model",
  "messages": [{
    "role": "user",
    "content": "what is a good banana bread recipe?",
    "content_blocks": [{ "type": "text", "text": "<banned keyword / jailbreak payload>" }]
  }]
}
```

The scan sees only the benign recipe question → Allow; the OpenAI/Anthropic upstream receives the `content_blocks` payload. This defeats keyword blocklists, PII-block, Aliyun/Bedrock content policy, Azure Prompt Shield, Lakera, and OpenAI moderation alike, for any valid key holder.

## Fix

One change, one choke point: `message_scan_text` (`crates/aisix-guardrails/src/lib.rs`) now scans **every text surface the bridges forward** — the flat `content`, the `text`-type `content_blocks` entries, **and** `extra["tool_calls"]** (serialized whole, so neither a function name nor an argument can hide a token). Because every guardrail kind's input/output collector routes through this single helper, the fix covers all kinds at once.

The `tool_calls` surface was added after an independent audit found the same bypass class survived through it: history-replay tool calls travel upstream verbatim through `extra` (the OpenAI bridge flattens them; the Anthropic bridge translates them into `tool_use` blocks), and the codebase already scans them on the **output** side (`guardrail_output_text`) and masks them in **redaction** (`redact_chat_format`) — only the input scan had missed them. The fix now aligns all three (input-scan / output-scan / redaction) on the same three surfaces.

Cost: in the legitimate vision round-trip where `content` was *derived* from `content_blocks`, the same text is scanned twice — harmless for keyword/regex/PII/moderation (no false negatives, no change to a block/allow decision). Serializing `tool_calls` whole (mirroring `guardrail_output_text`) can surface JSON structural tokens to the scanner; that is the same accepted tradeoff the output scan already makes and is the only variant that also closes a payload hidden in a function name. The `#465` empty-`content` fallback is preserved as a subset.

## Scope

- The exposure is on `/v1/chat/completions` with an OpenAI-compat, Azure-OpenAI, or Anthropic upstream (the bridges that forward `content_blocks`). Bedrock and Vertex read the flattened text and drop `content_blocks`, so they were never exploitable. The native `/v1/messages` and `/v1/responses` paths reconcile the fields at ingress and were not exploitable.
- The fix is upstream-independent: the block now happens **before** dispatch, so no bridge is reached once a payload is caught — which is why the e2e below covers it with a single upstream family.
- Redaction (`redact.rs`) already covered both fields; no change needed there.

## Testing

- **Unit** (`aisix-guardrails/src/lib.rs`): `message_scan_text_scans_content_blocks_even_when_flat_content_is_nonempty` and `message_scan_text_scans_tool_call_payload` — assert the union covers `content_blocks` and `tool_calls` (name + arguments) respectively. The existing empty-`content` fallback test still passes.
- **Unit** (`keyword.rs`): `blocks_keyword_hidden_in_content_blocks_behind_benign_flat_content` and `blocks_keyword_hidden_in_tool_call_arguments` — the security contract at the guardrail level for both surfaces.
- **E2E** (`guardrail-keyword-message-locations-e2e.test.ts`, 2 new cases): black-box — a raw request hiding the forbidden literal in (1) a `content_blocks` text entry and (2) a replayed `tool_calls` argument, each behind benign flat `content`, returns `422 content_filter`, the upstream is untouched (block short-circuits before dispatch), and the literal does not leak into the envelope.
- Full `aisix-guardrails` (244) + `aisix-proxy` (704) crate tests and the sibling keyword/output/PII-redaction e2e suites pass; `cargo fmt` + clippy clean.

TDD: the two unit tests were written first and observed failing (scan returned only the benign `content`), then the fix made them pass.

### DIFF
--- a/crates/aisix-guardrails/src/keyword.rs
+++ b/crates/aisix-guardrails/src/keyword.rs
@@ -213,6 +213,32 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn blocks_keyword_hidden_in_tool_call_arguments() {
+        // Same bypass via `extra["tool_calls"]`: a history-replay tool
+        // call carries the banned keyword in its arguments, which the
+        // bridges forward upstream. Must be scanned and blocked.
+        let msg: ChatMessage = serde_json::from_value(serde_json::json!({
+            "role": "assistant",
+            "content": null,
+            "tool_calls": [{
+                "id": "c1",
+                "type": "function",
+                "function": {
+                    "name": "lookup",
+                    "arguments": "{\"q\":\"the FORBIDDEN word\"}"
+                }
+            }]
+        }))
+        .unwrap();
+        let g = KeywordBlocklist::new(vec![KeywordRule::literal("Forbidden")]);
+        let v = g.check_input(&ChatFormat::new("m", vec![msg])).await;
+        assert!(
+            v.is_block(),
+            "keyword hidden in tool_call arguments must be blocked"
+        );
+    }
+
+    #[tokio::test]
     async fn empty_literal_pattern_never_matches() {
         let g = KeywordBlocklist::new(vec![KeywordRule::literal("")]);
         let v = g.check_input(&req(&[("user", "anything")])).await;
--- a/crates/aisix-guardrails/src/keyword.rs
+++ b/crates/aisix-guardrails/src/keyword.rs
@@ -192,6 +192,27 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn blocks_keyword_hidden_in_content_blocks_behind_benign_flat_content() {
+        // Guardrail bypass: a caller sends benign top-level `content` but
+        // hides the banned keyword in a `content_blocks` text entry, which
+        // the provider bridges forward upstream. The scan must catch it —
+        // otherwise every input guardrail is trivially bypassed by a valid
+        // key holder.
+        let msg: ChatMessage = serde_json::from_value(serde_json::json!({
+            "role": "user",
+            "content": "what is a good banana bread recipe?",
+            "content_blocks": [{"type": "text", "text": "the FORBIDDEN word"}]
+        }))
+        .unwrap();
+        let g = KeywordBlocklist::new(vec![KeywordRule::literal("Forbidden")]);
+        let v = g.check_input(&ChatFormat::new("m", vec![msg])).await;
+        assert!(
+            v.is_block(),
+            "keyword hidden in content_blocks must still be blocked"
+        );
+    }
+
+    #[tokio::test]
     async fn empty_literal_pattern_never_matches() {
         let g = KeywordBlocklist::new(vec![KeywordRule::literal("")]);
         let v = g.check_input(&req(&[("user", "anything")])).await;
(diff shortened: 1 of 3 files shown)


moonming:
## Independent cold-audit triage (merge gate)

An independent auditor reviewed this fix cold (correctness, completeness/missed-siblings, is-the-bypass-actually-closed, test quality, public-repo hygiene). Verdict was GO-WITH-CHANGES; all findings are resolved in `c934cde`.

### Fixed

- **HIGH — the same bypass class survived through `extra["tool_calls"]`.** History-replay tool calls are forwarded upstream verbatim (the OpenAI bridge flattens them; the Anthropic bridge translates them into `tool_use` blocks), but the input scan only read `content` + `content_blocks`. The codebase already scanned `tool_calls` on the **output** side (`guardrail_output_text`) and masked them in **redaction** (`redact_chat_format`) — only the input scan had missed them. So a caller could hide a payload in a replayed `tool_calls` argument (or function name) exactly as with `content_blocks`. → `message_scan_text` now serializes the whole `tool_calls` payload too (mirroring `guardrail_output_text`, so neither name nor argument can hide a token), plus a unit test and an e2e case for the tool_calls shape.
- **MEDIUM — the docstring overclaimed "a superset of everything a bridge can forward."** With the `tool_calls` gap that was false. → Docstring rewritten to enumerate the exact three forwarded surfaces (`content`, `content_blocks` text, `extra["tool_calls"]`) and to point at the two siblings (`guardrail_output_text`, `redact_chat_format`) that must stay in lockstep.

### Noted (not blocking)

- **LOW — only `type == "text"` content blocks are scanned.** An alternate text-bearing block shape (e.g. `input_text`) would not be scanned. Theoretical today: the OpenAI chat upstream rejects non-`text`/`image_url` block types, so no bridge forwards such a block as model-visible text. If the accepted block vocabulary widens, revisit here. No code change now.
- **LOW (accepted) — double-scan of derived vision content.** For the legitimate array-form path (`content` derived from `content_blocks`) the same text is scanned twice; harmless for keyword/regex/PII (no false negatives, no block/allow change) — the same cost the output scan already accepts.

The three text-collection paths (input scan / output scan / redaction) now agree on the same surface set, so this class can't drift between them silently. Full `aisix-guardrails` (244) + `aisix-proxy` (704) crate tests, the extended message-locations e2e (5 cases), and the sibling guardrail e2e suites pass; `cargo fmt` + clippy clean.

Issue #554, opened 2026-06-08
/v1/completions: add output guardrails on generated choices[].text


moonming:
## Summary
#545 (PR #551) added **input** guardrails to `/v1/completions`. Unlike `/v1/images` (image), `/v1/rerank` (scores), and `/v1/audio/speech` (audio), legacy `/v1/completions` returns **generated text** (`choices[].text`), so a configured **output** content/DLP block can still be bypassed on this surface.

## Scope
- [ ] Scan `choices[].text` (non-streaming) and the streamed `choices[].text` deltas (streaming, buffer-then-scan) on `/v1/completions`, mirroring the `/v1/responses` output work (#544): synthesize a ChatResponse, run `check_output`, Block -> 422; streaming uses BufferFull hold-back with the cap.

Deferred from #551 to keep that PR focused on the demonstrated input bypass. Refs #545, #544, #719.

jarvis9443:
Implemented across two PRs, neither of which linked back here.

The non-streaming ask landed in #683 (`3b3fe67c`, 2026-07-02), as finding [23] of a separate security review. The response is scanned through a synthetic `ChatResponse`; `choices[].text` is joined for the scan and masked for mask actions across every choice; a block returns 422 with the already-billed usage still recorded. Covered by `completions-output-guardrail-e2e.test.ts`, which drives an upstream emitting a forbidden word and asserts a `content_filter` 422 that never carries the word.

The streaming ask was closed differently from what this issue specified, and that is worth stating plainly. This issue asked for buffer-then-scan over streamed `choices[].text` deltas with a BufferFull hold-back. That was never built. Instead #1099 (`83ef4a83`, 2026-09-02) makes the route refuse `stream: true` outright with a 400 before the provider is contacted, pinned by `completions-legacy-e2e.test.ts`. The bypass this issue named is gone, but by removing the surface rather than by scanning it.

Closing on that basis. If streaming support on this route is still wanted, that is a new issue rather than this one.

PR #818, opened 2026-07-24
feat(downstream): configurable idle timeout and SSE heartbeat on the inbound side

jarvis9443:
## Problem

`upstream:` (#808) bounded the connections the gateway *dials out* with. The inbound half had nothing:

1. **No downstream idle timeout at all.** hyper 1.x defaults `header_read_timeout` to 30s, but `Time::check` silently drops a defaulted duration when no timer is installed — and neither `axum::serve` nor `axum_server` installs one. So an accepted connection was held open forever once it went idle.
2. **The SSE heartbeat only existed on `/v1/chat/completions`**, at a hardcoded 15s. `/v1/messages` and `/v1/responses` streams sent nothing while the model was silent.
3. **Half the outbound clients never got #808's settings.** Only the four provider bridges and the proxy's shared client used `client_builder()`; the 7 guardrail clients, MCP OAuth, A2A, telemetry, heartbeat and the OTLP exporter were still on reqwest's defaults (no connect timeout, TCP keepalive off, 90s pooled-connection lifetime).

## Implementation

New `downstream:` block, the mirror of `upstream:`:

```yaml
downstream:
  idle_timeout_secs: 0          # 0 = never close an idle connection
  # sse_keepalive_interval_secs: 15
```

**`idle_timeout_secs`** goes on hyper's header-read timer. That timer arms only while hyper is waiting for a request head, and hyper only waits for one once the previous response has been fully written (`Conn::can_read_head` needs the read half back at `Init`, which `try_keep_alive` reaches only when reading *and* writing are done). So it covers exactly the between-requests window — a slow model or a long SSE stream is never interrupted. Reaching the builder means the plain-HTTP listener moves from `axum::serve` to `axum_server`, which the TLS listener already used; `serve_connection_with_upgrades` and `ConnectInfo<SocketAddr>` are preserved, so WebSocket upgrades and real-ip resolution are unaffected. HTTP/1.1 only — hyper has no h2 equivalent.

Default `0` = today's behaviour. Closing first is what hands the node in front a connection it still considers usable — the same failure this gateway avoids upstream by keeping `pool_idle_timeout_secs` low — and a fronting Envoy pools for an hour by default. Set it *above* the pool idle timeout of whatever sits in front.

**`sse_keepalive_interval_secs`** (default 15, `0` disables) now drives every SSE surface: axum's `Sse` keep-alive on chat, and a `with_heartbeat` stream wrapper on bridged `/v1/messages`, bridged `/v1/responses`, and the `/v1/responses` byte passthrough. The frame is `:\n\n` — the same comment axum writes, ignored by every conforming SSE parser. Not applied to opaque binary passthroughs (audio, images), where it would corrupt the body.

**Uniform connection layer**: guardrails, MCP OAuth, A2A, telemetry, heartbeat and the OTLP exporter now build from `aisix_gateway::client_builder()`. A test walks every crate's production source and fails on a bare `reqwest::Client::builder()` / `::new()`, so the next client can't drift back (verified it catches a reintroduced offender).

## Which surfaces the heartbeat covers

Audited every response-body outlet in `aisix-proxy`, since the issue asks whether the raw passthrough paths need one too:

| Surface | Shape | Heartbeat |
| --- | --- | --- |
| `/v1/chat/completions` (both stream sites) | axum `Sse` | yes, now configurable |
| `/v1/messages` bridged + native passthrough | SSE byte stream | yes, new |
| `/v1/responses` bridged + native passthrough | SSE byte stream | yes, new |
| `/passthrough/:provider/*rest` | **not streaming** — buffers the whole upstream body (`bytes().await`) for the output-guardrail scan | n/a |
| `/v1/completions` | non-streaming, always `Json(...)` | n/a |
| audio / images / videos | binary or non-streaming | deliberately not — a comment frame would corrupt the body |
| `/v1/realtime` | WebSocket, has its own idle semantics (`stream_timeout`) and would need WS ping frames, not SSE comments | not covered |

Worth flagging from that audit: the raw tunnel buffers, so a slow provider call through `/passthrough/...` puts **zero** bytes on the client connection until the whole body has arrived — the same "looks idle" exposure the heartbeat fixes for SSE, and one that can't be fixed the same way because the content type is unknown. Out of scope here; noting it so it isn't assumed covered.

## Not implemented, deliberately

- **Per-provider overrides of the pool settings.** The pool idle timeout only has to be below the *shortest* hop on the path, so a single global value is sufficient; per-provider pools multiply FDs and memory for no reachability gain. LiteLLM and Portkey expose no equivalent either.
- **`max_requests_per_connection` / `max_connection_lifetime`.** reqwest exposes neither.
- **A separate upstream stream-idle timeout.** Per-model `stream_timeout` already bounds each inter-chunk gap (#554, corrected for Azure in #809); this would be a second name for it.

## Behaviour changes

- Graceful shutdown now drains without a deadline on both listeners (the TLS one previously capped at 10s). An in-flight stream can run for minutes and the platform already caps total shutdown time; idle connections are closed immediately either way.
- SSE responses on `/v1/messages` and `/v1/responses` now carry heartbeat comments during upstream silence.
- Guardrail/MCP/A2A/telemetry/exporter calls get a 5s connect timeout, TCP keepalive, and a 30s pool idle timeout.

## Tests

- `downstream-connection-e2e.test.ts`: an idle keep-alive connection is closed at the deadline; **a request in flight past the deadline still completes**; heartbeats appear on all three SSE code paths while the upstream is silent, without disturbing the payload. Mutation-checked — with the idle timeout un-applied and the interval hardcoded, 4 of the 5 fail (the in-flight case is the control and keeps passing).
- Unit tests for the heartbeat combinator (silence heartbeats, busy stream untouched, disabled = pass-through, terminal error preserved) and for the `downstream:` config defaults/overrides.
- `upstream-pool-idle-e2e.test.ts`: `pool_idle_timeout_secs` shipped in #808 without behavioural coverage. Against an upstream that never closes first, a request after the deadline opens a new TCP connection; the same gap with the knob off reuses the pooled one. The two cases are each other's control.
- Re-ran the listener-sensitive suites: TLS, real-ip/XFF, WebSocket realtime, SIGHUP reload, SIGTERM shutdown, admin-off, `#554` timeout fallback, audio timeout, streaming usage, guardrails.

Fixes api7/AISIX-Cloud#1126

## Summary by CodeRabbit

* **New Features**
  * Added configurable downstream idle timeouts for HTTP connections.
  * Added configurable SSE heartbeats, enabled by default at 15 seconds, to keep quiet streams active.
  * Applied SSE keep-alives across supported chat, messages, and responses streaming endpoints.
  * Standardized outbound connection handling for improved reuse and timeout consistency.

* **Tests**
  * Added end-to-end coverage for downstream timeouts, SSE heartbeats, and upstream connection pooling.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -36,6 +36,7 @@ name = "aisix-a2a"
 version = "0.3.0"
 dependencies = [
  "aisix-core",
+ "aisix-gateway",
  "async-trait",
  "axum",
  "futures",
@@ -182,6 +183,7 @@ name = "aisix-mcp"
 version = "0.3.0"
 dependencies = [
  "aisix-core",
+ "aisix-gateway",
  "async-trait",
  "axum",
  "futures",
@@ -203,6 +205,7 @@ name = "aisix-obs"
 version = "0.3.0"
 dependencies = [
  "aisix-core",
+ "aisix-gateway",
  "aliyun-log-sdk-protobuf",
  "aliyun-log-sdk-sign",
  "async-trait",
@@ -423,6 +426,7 @@ dependencies = [
  "config",
  "etcd-client",
  "hostname",
+ "hyper-util",
  "rcgen",
  "reqwest 0.12.28",
  "rustls 0.23.38",
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -184,6 +184,29 @@ upstream:
   # Cap on idle connections kept per upstream host. Unset = unbounded.
   # pool_max_idle_per_host: 32
 
+# Connection-layer settings for the inbound side — the clients, or the
+# gateway in front of this one, talking to the listeners above. The
+# mirror image of `upstream:`.
+downstream:
+  # How long an accepted connection may sit idle — response written, no
+  # next request started — before the gateway closes it. HTTP/1.1 only.
+  # An in-flight request or SSE stream is never interrupted, however long
+  # it runs; the timer only arms between requests.
+  #
+  # 0 (the default) never closes an idle connection and leaves that to
+  # the peer. Whatever sits in front pools its own connections, and
+  # closing first is what hands *it* a connection it thinks is still
+  # usable — the same failure this gateway avoids upstream by keeping
+  # pool_idle_timeout_secs low. If you set this, keep it ABOVE the pool
+  # idle timeout of the node in front.
+  idle_timeout_secs: 0
+
+  # Interval between SSE heartbeat comments (":\n\n") sent while a
+  # streaming response has produced nothing. Keeps a proxy in front from
+  # treating a model that is slow to its first token as an abandoned
+  # connection. 0 disables the heartbeat.
+  # sse_keepalive_interval_secs: 15
+
 # Models, API keys, provider keys, guardrails, cache policies, and
 # observability exporters are NOT defined in this file. They are stored
 # in etcd and managed via the Admin API (see docs/api-admin.md). This
--- a/config.managed.yaml
+++ b/config.managed.yaml
@@ -113,3 +113,11 @@ upstream:
   # tcp_keepalive_retries: 5
   pool_idle_timeout_secs: 30
   # pool_max_idle_per_host: 32
+
+# Connection-layer settings for the inbound side. `idle_timeout_secs: 0`
+# (the default) never closes an idle client connection, leaving that to
+# the peer — set it only when a node in front pools connections for a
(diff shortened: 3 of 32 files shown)


jarvis9443:
@coderabbitai review

jarvis9443:
@coderabbitai full review

jarvis9443:
@coderabbitai full review

PR #840, opened 2026-07-29
feat(mcp): expose REST APIs as MCP tools from an OpenAPI spec (type: openapi)

jarvis9443:
## What

A registered `mcp_server` can now be backed by a plain REST API instead of a real MCP upstream. Setting `type: openapi` plus an OpenAPI 3.x document under `spec` makes the gateway generate one MCP tool per operation and execute `tools/call` as an HTTP request against the resource's `url` (the REST base URL). The gateway-held credential — `bearer`, `api_key` (default `x-api-key` header, overridable via the new `api_key_header` field), or `oauth2` client credentials through the existing token cache — is injected on every outbound request and never exposed to the calling agent.

## How

- `aisix-core`: `McpServer` gains `type` (`mcp` | `openapi`, default `mcp` — existing rows are untouched), `spec`, and `api_key_header`; resource JSON schema regenerated.
- `aisix-mcp`: new `openapi` module with an `OpenApiBridge` implementing the existing `McpBridge` trait; `McpGateway::from_snapshot` picks the bridge per `type`. Because the bridge sits behind the same surface, existing governance — approval gating, per-tool ACLs / MCP policies, per-(key × server) rate limits, guardrails, usage events — applies to generated tools unchanged.
- Tool generation follows LiteLLM's `openapi_to_mcp_generator` (sanitized `operationId` names capped at 128 chars, fallback `method_path`, path/query params as schema properties, JSON request body as one `body` property), with deliberate improvements:
  - bounded local `$ref` resolution (depth 16, 256 expansions, cycle-safe) so referenced schemas keep their shape;
  - non-2xx responses and argument mistakes (e.g. a missing path parameter) return tool-level `isError` results the agent can react to;
  - operations whose request body has no `application/json` variant are skipped rather than emitting a tool that cannot succeed;
  - header/cookie parameters are not exposed to the agent (upstream headers are the gateway's to set);
  - path parameter values are rejected on traversal (`/`, `\`, `.`, `..`) and percent-encoded.
- Admin API writes validate strictly via the new `aisix_mcp::validate_spec`: missing/non-object spec, Swagger 2.0 (with a conversion hint), zero generatable operations, duplicate post-sanitization tool names, and `api_key_header` coupling all fail with a message naming the problem. Snapshot loading stays permissive: a broken row degrades like an unreachable upstream (tools absent, warning logged) instead of poisoning the snapshot.

## Behavior notes

- Redirects are not followed on generated tool calls, so the gateway-held credential cannot be re-sent to a host the operator never configured.
- An upstream 401 under `oauth2` invalidates the cached token so the next call re-mints (mirrors the MCP connect path).

## Tests

- Unit: generator (names, collisions, $ref/cycles, skips, schemas), URL building/traversal, query/body coercion, strict `validate_spec`, admin coupling.
- Crate integration (`tests/openapi_tool_roundtrip.rs`): real axum REST API + full gateway over Streamable HTTP driven by an rmcp client — list/call, auth injection for both bearer and custom-header api_key, error surfacing, broken-spec degradation next to a healthy server.
- E2E (`tests/e2e/src/cases/mcp-openapi-e2e.test.ts`): real `aisix` binary + etcd + fake ERP REST upstream — generated tool discovery with spec-derived schemas, GET/POST execution with gateway-held auth, non-2xx and missing-param error results, and per-tool ACL enforcement on generated tools.

Fixes api7/AISIX-Cloud#1077 (control-plane exposure ships as a paired AISIX-Cloud PR; docs as a paired api7/docs PR)


jarvis9443:
@coderabbitai review

PR #844, opened 2026-07-30
feat(config): deployment-wide default for upstream request/stream timeouts

jarvis9443:
## Problem

A Model that sets neither `timeout` nor `stream_timeout` has **no upstream deadline at all**: an upstream that accepts the connection and then goes silent forever holds the request open indefinitely. TCP keepalive (already on by default) only detects a dead peer — it cannot detect a live-but-silent one, so the only bound left is the client's own patience. The mainstream LLM-proxy baseline never runs unbounded: its router falls back to a global `request_timeout` of **6000 s** when no deployment-level timeout is set, and its per-deployment `stream_timeout` falls back to the non-stream timeout the same way.

Two smaller gaps in the same area:

- The `timeout`/`stream_timeout` fields on a model group (a Model carrying `routing`) were **dead fields** — members only ever used their own values, unlike `retries`, which already resolves member → group → deployment default.
- There was no config-file surface to set a deployment-wide timeout at all.

## Change

- New `upstream.timeout_ms` (default `6_000_000` = 6000 s, matching the reference proxy's `request_timeout`) and `upstream.stream_timeout_ms` (default `0` = fall back to `timeout_ms`). Env overrides work like every other `upstream.*` knob (`AISIX_UPSTREAM__TIMEOUT_MS`, `AISIX_UPSTREAM__STREAM_TIMEOUT_MS`).
- New `routing::effective_timeouts` resolves deadlines **target → group → deployment default**, mirroring `effective_retries`, and is applied uniformly across every dispatch family: chat, messages (passthrough + cross-provider), responses (both), count_tokens, completions, embeddings, rerank, images, audio (transcription + speech), videos, jobs, passthrough, ensemble (panel + judge), and the realtime idle cap.
- `timeout: 0` at model or group level opts that resource out of the backstop (unchanged for existing users of `0`); `upstream.timeout_ms: 0` restores the old unbounded behaviour deployment-wide. `stream_timeout: 0` keeps its historical "defer" semantics at every level.
- Streaming precedence follows the baseline: resource levels first (`stream_timeout`, then the resource-resolved `timeout`), deployment defaults last. Within that, the dedicated stream knob outranks the generic one at every level — a group's `stream_timeout` beats a member's own `timeout` for the streaming budget, and supplies one even to a member whose `timeout: 0` opted out of the request deadline (unit-tested; same shape as the reference router, whose stream chain is exhausted before its non-stream chain starts).
- **The deployment-default streaming budget does not arm the pre-200 first-chunk peek** (`TimeoutBudget::stream_configured`). Peeking withholds the 200 until the first token — that is the #554 slow-first-token failover contract and stays exactly as-is for resource-configured budgets — but doing it for the deployment default would withhold headers for every model and silence the #818 SSE heartbeats whose whole purpose is covering a slow first token. With only the default budget, a first-chunk stall surfaces as an in-band timeout after the 200 instead of failing over. Same shape as `RetryBudget::covers`: explicit config opts into the sharper behaviour.

## Behaviour changes

1. Models with no `timeout` configured anywhere are now bounded at 6000 s end-to-end (non-streaming) / 6000 s inter-chunk (streaming). This is a backstop against infinite hangs, deliberately generous so it can never cut a legitimate long request.
2. Group-level `timeout`/`stream_timeout` now apply to members that don't set their own. A pre-existing group-level `stream_timeout` (previously a dead field) therefore also arms the pre-200 first-chunk peek for members without their own streaming config.
3. For CP-managed deployments, the control plane currently normalizes an explicit `timeout: 0` to unset — after this change those models pick up the 6000 s backstop instead of staying unbounded. A companion CP PR (preserve explicit `0`, expose the group-level fields) is queued to restore the opt-out there; standalone/file-based deployments are unaffected.

## Tests

- `routing::effective_timeouts` unit tests: three-level precedence, explicit-zero opt-out at each level, stream fallback order, peek arming.
- New E2E `upstream-timeout-defaults-e2e.test.ts` (real binary + etcd + mock upstream): default applies (504 at the deadline), `timeout: 0` opt-out, explicit model timeout beats the default, streaming inherits the default as its chunk budget without suppressing the 200/heartbeats, group timeout applies to members.
- `timeout-fallback-e2e` (the #554 contract) and `downstream-connection-e2e` (the #818 heartbeat contract) pass unchanged.
- CP contract-check pre-run against the regenerated schemas: clean (the `timeout`/`stream_timeout` schema delta is description-only).

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -160,6 +160,21 @@ ratelimit:
 # The values below are the defaults; uncomment to override. Every
 # duration accepts 0 to switch that knob off.
 upstream:
+  # Deployment-wide default for `Model.timeout`: the end-to-end deadline
+  # for non-streaming calls, and the fallback streaming budget (below),
+  # for every model that does not set its own. A backstop against an
+  # upstream that accepted the connection and then goes silent forever —
+  # not a responsiveness target (set per-model `timeout` for that), so it
+  # is deliberately generous: deep-reasoning calls can run past 10
+  # minutes. A model opts out with `timeout: 0`; setting 0 here restores
+  # the old "no deadline unless configured" behaviour.
+  # timeout_ms: 6000000
+
+  # Deployment-wide default for `Model.stream_timeout`: the maximum gap
+  # between streaming chunks. 0 falls back to `timeout_ms`, mirroring how
+  # an unset `Model.stream_timeout` falls back to `Model.timeout`.
+  # stream_timeout_ms: 0
+
   # Max time for DNS + TCP + TLS before an attempt fails. Without it a
   # black-holed upstream is bounded only by the model's own timeout.
   # connect_timeout_ms: 5000
--- a/config.managed.yaml
+++ b/config.managed.yaml
@@ -107,6 +107,8 @@ cache:
 # the gateway expires them — the symptom is intermittent transport
 # errors against an otherwise healthy upstream.
 upstream:
+  # timeout_ms: 6000000
+  # stream_timeout_ms: 0
   # connect_timeout_ms: 5000
   # tcp_keepalive_secs: 60
   # tcp_keepalive_interval_secs: 30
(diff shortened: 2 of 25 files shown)


jarvis9443:
@coderabbitai review

PR #845, opened 2026-07-30
fix(bedrock): enforce the model deadline and connect timeout on the SDK client

jarvis9443:
## Problem

The Bedrock bridge was the only dispatch path where a model's `timeout` did not actually bound the call. The SDK client is built with **no `timeout_config` at all**, and `ctx.deadline` was consulted only inside `map_sdk_error` — i.e. after a call had already failed for some other reason, purely to relabel the error as `Timeout`. A Bedrock upstream that accepted the connection and then went silent held the request open indefinitely, while every other bridge cancels at the deadline via its `with_deadline` / `tokio::time::timeout` wrapper. (The connect phase itself was bounded only by the SDK default plugins' own 3.1 s — `upstream.connect_timeout_ms`, which reaches every reqwest-based bridge through the shared client builder, never reached the AWS SDK client, including its documented `0` = disabled convention.)

## Change

`build_client` now installs a `TimeoutConfig` on the SDK client:

- `connect_timeout` ← the shared `upstream.connect_timeout_ms` (same knob every other bridge honors);
- `operation_timeout` ← the per-request `ctx.deadline` (the model/group/deployment-resolved timeout). An elapsed deadline surfaces as `SdkError::TimeoutError`, which `map_sdk_error` already maps to `BridgeError::Timeout`, so retry / fallback / cooldown classification is untouched. The per-request client construction the bridge already does (stateless by design) is what makes a per-request `operation_timeout` possible.

`converse_stream` bodies stay bounded per-chunk by the proxy-side read-timeout wrapper, exactly like every other bridge's streams; `operation_timeout` covers send + response headers only.

SDK-internal retries are disabled (`RetryConfig::disabled()`): retries belong to the gateway's `routing::effective_retries` budget, which emits per-attempt telemetry and honors per-model config. Left at its default, the SDK adds a hidden standard-mode retry layer (up to 3 attempts) underneath the gateway's own budget, grinding the upstream invisibly — no other bridge's HTTP client retries on its own.

Not in scope: the Bedrock **guardrail** client builds its own SDK client but wraps every call in `tokio::time::timeout(guardrail.timeout_ms)` (default 5 s), so it is already bounded end-to-end.

## Behaviour changes

1. Bedrock models now time out at their resolved deadline like every other provider (previously: unbounded on a silent upstream).
2. Hidden SDK-level retries (up to 3 transparent attempts on retryable failures) no longer happen; retry behaviour is governed solely by `Model.retries` / `routing.retries` / `upstream.retries`.

## Tests

`chat_deadline_cancels_a_silent_upstream` drives the real `bridge.chat()` through wiremock with an 8 s response delay and a 300 ms deadline, asserting `BridgeError::Timeout` **and** an elapsed time far below the delay (cancellation, not relabelling). Verified to fail without the fix: the call sits through the full 8 s and returns `Ok`.

`sdk_does_not_retry_on_its_own` pins the single-attempt semantics with a wiremock hit counter: a 500 reaches the gateway after exactly one wire attempt. Verified to fail without `RetryConfig::disabled()` — the SDK hits the upstream 3 times.

### DIFF
--- a/crates/aisix-provider-bedrock/src/bridge.rs
+++ b/crates/aisix-provider-bedrock/src/bridge.rs
@@ -355,17 +355,21 @@ fn build_client(
     );
     // The SDK enforces the deadlines the other bridges get from
     // `tokio::time::timeout` wrappers: the shared `upstream.connect_timeout`
-    // bounds the dial, and the resolved per-request deadline cancels the
-    // whole operation (SdkError::TimeoutError → BridgeError::Timeout in
-    // `map_sdk_error`). Without these the SDK client has NO timeouts and a
-    // silent upstream holds the request open past the model's `timeout` —
-    // the deadline used to matter only for post-hoc error relabelling.
-    // `operation_timeout` covers send + response headers; a converse_stream
-    // body is bounded per-chunk by the proxy's read-timeout wrapper, same
-    // as every other bridge.
+    // bounds the dial (explicitly disabled when the operator set `0`, or
+    // the SDK's default plugins would quietly restore their own 3.1 s),
+    // and the resolved per-request deadline cancels the whole operation
+    // (SdkError::TimeoutError → BridgeError::Timeout in `map_sdk_error`).
+    // Without an operation timeout nothing bounds the call after connect —
+    // a silent upstream holds the request open past the model's `timeout`,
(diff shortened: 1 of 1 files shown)


jarvis9443:
@coderabbitai review

PR #846, opened 2026-07-30
fix(mcp): apply the upstream.* connection settings to the rmcp transport

jarvis9443:
## Problem

The rmcp streamable-http transport's HTTP client was outside the `upstream.*` connection-settings contract: rmcp pins its own reqwest line (0.13, vs the workspace's 0.12), so the `from_uri`/`from_config` constructors built rmcp's `default_http_client()` — which sets **no connect timeout**. A black-holed MCP upstream (SYN swallowed) was bounded only by the coarse per-operation `McpServer.timeout_ms` (default 30 s), instead of failing at `upstream.connect_timeout_ms` like every other outbound path. This is the same class the no-bare-client guard exists for (#1122/#1126); it missed this site because the client is built inside rmcp.

## Change

- New `shared_http_client()` in `aisix-mcp`: one client shared by every MCP upstream, built against rmcp's reqwest line with the `upstream_http::config()` values applied — `connect_timeout_ms`, the TCP keepalive triple, `pool_idle_timeout_secs`, `pool_max_idle_per_host`. All four transport constructions (none / bearer / api-key / oauth2) now go through `StreamableHttpClientTransport::with_client`. Auth stays per-request (transport-injected), so sharing one client across differently-authenticated upstreams is safe — same posture as the provider bridges.
- The `no_production_code_builds_a_bare_reqwest_client` guard names `rmcp_reqwest::Client::` in `aisix-mcp/src/bridge.rs` as the sanctioned second construction site (scoped to that file; a bare rmcp client anywhere else still gets flagged).

## Behaviour changes (the honest delta vs rmcp's default client)

1. **A connect timeout exists at all** (default 5 s) — the actual fix.
2. **Connection pooling turns ON.** rmcp deliberately disables idle pooling (`pool_max_idle_per_host(0)`) to dodge ~40 ms delayed-ACK stalls on reused connections. Here reuse wins: every other outbound path pools under `upstream.*` management (30 s idle expiry), per-call TCP+TLS handshakes to a remote MCP server cost far more than the stall rmcp avoids, and the reference proxy's MCP path (python-sdk over httpx) pools by default too. `upstream.pool_max_idle_per_host: 0` restores rmcp's exact behaviour.
3. The TCP keepalive triple moves from reqwest's default 15 s/15 s/3 to the deployment's 60 s/30 s/5.

Operation-level deadlines are unchanged: `McpServer.timeout_ms` still bounds every MCP operation end-to-end.

## Tests

`connect_timeout_bounds_an_unreachable_upstream` dials a black-holed upstream (TEST-NET-3) with a 12 s operation timeout and asserts the dial is cut well below it (the 5 s connect timeout). Verified to fail without the fix: the dial hangs the full 12 s and surfaces the outer-timeout error instead.

### DIFF
--- a/crates/aisix-gateway/src/upstream_http.rs
+++ b/crates/aisix-gateway/src/upstream_http.rs
@@ -260,6 +260,14 @@ mod tests {
                 None => &src[..],
             };
             for (n, line) in production.lines().enumerate() {
+                // `rmcp_reqwest::Client::*` is aisix-mcp's sanctioned
+                // counterpart of `client_builder()`: rmcp pins its own
+                // reqwest line, so `shared_http_client()` there applies
+                // the same `upstream_http::config()` values to that crate
+                // version instead of building on defaults.
+                if line.contains("rmcp_reqwest::Client::") {
+                    continue;
+                }
                 if line.contains("reqwest::Client::builder()")
                     || line.contains("reqwest::Client::new()")
                 {
--- a/crates/aisix-gateway/src/upstream_http.rs
+++ b/crates/aisix-gateway/src/upstream_http.rs
@@ -259,13 +259,15 @@ mod tests {
                 Some(i) => &src[..i],
                 None => &src[..],
             };
+            // `shared_http_client()` in aisix-mcp is the sanctioned
+            // counterpart of `client_builder()` for rmcp's own reqwest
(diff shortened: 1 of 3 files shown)


jarvis9443:
@coderabbitai review

PR #848, opened 2026-07-30
fix(core): enforce the a2a/mcp name and credential rules on every config path

moonming:
Prerequisite for removing the Admin API write path — and a bug fix in its own right.

## The bug

The per-`auth_type` credential coupling, the name-shape rules, and the `type`/`spec`/`api_key_header` coupling are enforced **only** in this gateway's Admin API write handlers. A declaratively configured gateway never sees them. On `main` today:

```yaml
# resources.yaml
a2a_agents:
  - name: bad/name          # the name IS the /a2a/<name> path segment
    url: https://agents.example.com/a2a
    auth_type: bearer       # …with no secret to send
```

```
$ aisix validate --resources resources.yaml
OK: resources.yaml loaded 1 resource(s)
```

Both problems are real: a slash splits the agent's route into two segments, and `bearer` with no secret means the gateway authenticates upstream with nothing. The struct comments said this was deliberate — *"this gateway's own Admin API re-checks it on write"* — but that design leans on a write path that is about to be removed.

## What is lifted

Twelve rules, defined once and injected into the generated schema following the existing `model_one_of` / `mcp_policy_root_schema` pattern, so the **published JSON Schema and every runtime validator share one definition**:

| Rule | Why it matters |
|---|---|
| a2a `name` excludes `/?#%`, whitespace, control chars | the name is interpolated into the advertised agent-card URL **unencoded** (`a2a.rs:237`), so `a?b` advertises a URL whose path is just `/a2a/a` |
| mcp `name` excludes `__` **and a trailing `_`** | tools are advertised as `<name>__<tool>` and parsed with `split_once("__")`; `github_`+`list` → `github___list` → resolves to the non-existent server `github`, and `gh_`+`x` collides with `gh`+`_x` |
| a2a/mcp `bearer`/`api_key` ⇒ non-empty `secret` | otherwise the gateway sends an empty credential |
| mcp `oauth2` ⇒ `secret` + `client_id` + `token_url` | incomplete credential exchange fails at first tool call instead of at load |
| mcp `spec` required for, and limited to, `type: openapi`; must be an object; must not be Swagger 2.0 | |
| mcp `api_key_header` limited to `type: openapi` + `auth_type: api_key`; must be a legal RFC 7230 header name | |

Cross-field coupling is an injected `allOf` of `if`/`then` subschemas, so the resources stay flat — no `oneOf` restructuring, which the original comments gave as the reason for leaving the rules out.

## Not lifted (3 rules)

`aisix_mcp::validate_spec`, the `http::HeaderName` typing, and duplicate-tool-name detection all need to parse the OpenAPI document, and `aisix-mcp` depends on `aisix-core` — lifting them would be a dependency cycle. They stay on the write path for now; their disposition belongs to the removal PR. (The header-name **shape** is covered by the RFC 7230 pattern above; only the exact `HeaderName` parse is not.)

## ⚠️ Behavior changes

**One assertion is deliberately inverted.** `mcp_server_schema_stays_permissive_on_credential_coupling` asserted that an incomplete `oauth2` row still validates, so the loader keeps it and the runtime degrades that server gracefully. That reasoning depended on a write path existing to catch it. Rejecting at load is also the more diagnosable failure — a rejected row is named in `GET /status/config`'s `rejected[]`, whereas a loaded-but-degraded server silently serves no tools. The rationale is recorded at the test.

**Upgrade impact differs by configuration source:**

- **etcd:** a now-invalid row is dropped by the loader and named in `/status/config`'s `rejected[]`; the gateway keeps serving (`degraded`).
- **`resources.yaml`:** the file source is **all-or-nothing** (`filesource/status.rs:6`). A single now-invalid `a2a_agents`/`mcp_servers` entry rejects the whole file, so a gateway that boots with one **exits at startup** (`aisix-server/src/main.rs:493`) rather than starting degraded. Run `aisix validate --resources <file>` before rolling the image. On `SIGHUP` the last-good snapshot is retained and the entry appears in `rejected[]`.

**One diagnosability regression, accepted and recorded at the constraint.** Rejecting Swagger 2.0 in the schema loses the write path's targeted *"convert the spec to OpenAPI 3.x"* hint — a validator reports a pointer and a constraint, not advice. The rule is intact on every path. A post-schema semantic hook in the loaders would let the rule and the advice live together; worth doing, out of scope here.

## Also

The write handlers' own checks now sit behind the schema gate, so their tests observe `AdminError::Schema` instead of `BadRequest`. Both map to `400` (`error.rs:46`), so the wire contract is unchanged; those assertions now check the status and the failing pointer. Those handler branches are consequently unreachable — they are the thing being lifted, and the removal PR deletes them.

## Verification

`cargo fmt`, `clippy --workspace --all-targets` clean, full workspace tests pass, schemas regenerated so the drift gate passes.

Every rule was then exercised against a locally built `aisix validate`, not just unit tests — which mattered: the first draft of the `api_key_header` coupling used `dependentSchemas` (draft 2019-09) in a schema declaring draft-07, so the validator **silently ignored it** while every unit test passed. Running the nine documents through the binary caught it. It is now `dependencies`, with a regression test that says why.

## Summary by CodeRabbit

* **New Features**
  * Added stricter validation for agent and MCP server names, including reserved characters and trailing underscores.
  * Added authentication-based credential requirements, including OAuth2 fields.
  * Added validation for MCP server configuration fields, including OpenAPI settings and API key headers.

* **Bug Fixes**
  * Exports now fail when rejected entries produce incomplete files.
  * Standardized invalid configuration responses to HTTP 400.

* **Documentation**
  * Updated validation guidance and error-handling comments.

### DIFF
--- a/crates/aisix-admin/src/a2a_agents_handlers.rs
+++ b/crates/aisix-admin/src/a2a_agents_handlers.rs
@@ -137,6 +137,12 @@ fn assert_unique_name(
 
 #[cfg(test)]
 mod tests {
+    // These rules moved into the canonical schema (see
+    // `a2a_agent_credential_coupling` / `mcp_server_credential_coupling` and the
+    // `name` patterns), so the schema gate now rejects these payloads before
+    // `decode`'s own checks run. The variant is `Schema` rather than
+    // `BadRequest`; both map to 400, so the wire contract is unchanged. The
+    // assertions below check the status-bearing outcome, not the variant.
     use super::*;
     use serde_json::json;
 
@@ -144,7 +150,7 @@ mod tests {
     fn decode_rejects_slash_in_name() {
         let err = decode(&json!({"display_name": "a/b", "url": "https://x/a2a"}))
             .expect_err("`/` in the agent name must be rejected");
-        assert!(matches!(err, AdminError::BadRequest(_)));
+        assert_eq!(err.status(), axum::http::StatusCode::BAD_REQUEST);
     }
 
     #[test]
@@ -155,7 +161,7 @@ mod tests {
             "auth_type": "bearer"
         }))
         .expect_err("bearer auth without a secret must be rejected");
-        assert!(matches!(err, AdminError::BadRequest(_)));
+        assert_eq!(err.status(), axum::http::StatusCode::BAD_REQUEST);
     }
 
     #[test]
--- a/crates/aisix-admin/src/a2a_agents_handlers.rs
+++ b/crates/aisix-admin/src/a2a_agents_handlers.rs
@@ -4,8 +4,9 @@
 //! reject duplicate names (409), generate a uuid v4 on POST, bump revision on
 //! PUT. The name is the path segment under which the agent is exposed
 //! (`/a2a/<name>`), so it must be a single URL path segment (no `/`). The
-//! per-auth_type credential coupling is enforced here too, since the flat
-//! schema stays permissive on it.
+//! per-auth_type credential coupling is enforced by the canonical schema, so
+//! every configuration path rejects an incomplete credential set; the checks
+//! below are defense in depth.
 
 use aisix_core::models::validate_a2a_agent;
 use aisix_core::resource::ResourceEntry;
--- a/crates/aisix-admin/src/mcp_servers_handlers.rs
+++ b/crates/aisix-admin/src/mcp_servers_handlers.rs
@@ -197,6 +197,12 @@ fn assert_unique_name(
 
 #[cfg(test)]
 mod tests {
+    // These rules moved into the canonical schema (see
+    // `a2a_agent_credential_coupling` / `mcp_server_credential_coupling` and the
+    // `name` patterns), so the schema gate now rejects these payloads before
+    // `decode`'s own checks run. The variant is `Schema` rather than
+    // `BadRequest`; both map to 400, so the wire contract is unchanged. The
+    // assertions below check the status-bearing outcome, not the variant.
     use super::*;
     use serde_json::json;
 
@@ -204,7 +210,7 @@ mod tests {
     fn decode_rejects_separator_in_name() {
         let err = decode(&json!({"display_name": "a__b", "url": "https://x/mcp"}))
             .expect_err("`__` in the server name must be rejected");
-        assert!(matches!(err, AdminError::BadRequest(_)));
+        assert_eq!(err.status(), axum::http::StatusCode::BAD_REQUEST);
     }
 
     #[test]
@@ -215,7 +221,7 @@ mod tests {
             "auth_type": "bearer"
         }))
         .expect_err("bearer auth without a secret must be rejected");
-        assert!(matches!(err, AdminError::BadRequest(_)));
+        assert_eq!(err.status(), axum::http::StatusCode::BAD_REQUEST);
     }
 
     #[test]
@@ -226,7 +232,7 @@ mod tests {
             "auth_type": "api_key"
         }))
         .expect_err("api_key auth without a secret must be rejected");
-        assert!(matches!(err, AdminError::BadRequest(_)));
+        assert_eq!(err.status(), axum::http::StatusCode::BAD_REQUEST);
     }
 
     #[test]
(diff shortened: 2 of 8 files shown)


moonming:
## Second review pass — findings and disposition

A second reviewer went at this from angles the first pass did not cover: revert-fails-test (would deleting each constraint actually fail a test), the full ingestion lifecycle, export round-trip, and upgrade hazard. It found two things the first pass missed. I reproduced every claim before acting on it.

### Fixed in `9d79dbc`

**`aisix export` silently dropped newly-invalid rows and exited 0.** Export is the migration path off a store-backed deployment. A row the loader rejects is absent from the output, the file still loads, so `document.blocking` is empty and the command succeeds — a scripted migration treats an incomplete file as finished and loses those resources. Rejected rows now fail the export the same way a non-loadable file does; the written file remains for inspection.

**`api_key_header: null` was rejected.** The field is `Option<String>` and an explicit null means absent — the comment directly above my own constraint said exactly that — but `dependencies` triggers on key presence, null included. Now a value-sensitive `if`/`then` on `type: string`. Verified both directions: null loads; a real header with non-`api_key` auth still fails.

**Only 5 of the 12 rules were pinned at single-constraint granularity.** `spec` must-be-an-object was pinned by nothing at all: deleting it left the string fixture rejected by the sibling swagger check, so its test passed either way. Added mutation-pinning tests that start from a fully valid document and violate exactly one obligation — each oauth2 field independently, empty *and* null for every credential, an explicit `type: mcp` carrying openapi-only fields, a header with `auth_type: none`, and every excluded a2a character under both the `name` and `display_name` spellings.

**A fixture called `{"openapi": "3.0.0"}` a complete spec**, but tool generation needs a `paths` object; it now uses a spec that actually generates a tool.

### Confirmed, no action needed

The canonical schema does run on both supported ingestion paths — `resources.yaml` (`filesource/mod.rs:403`) and etcd/kine including initial load, cache restore, watch PUT and resync (`loader.rs:123/249/312`) — and `aisix export` goes through `build_snapshot`, so it does not bypass validation either.

### ⚠️ Open question for the maintainer

The review flagged that this PR mixes two different things:

- **lifting** the rules the old write handlers already enforced (`__` for mcp, `/` for a2a) — no upgrade risk, purely moving enforcement to where both paths see it;
- **tightening** beyond them (trailing `_`, and `?#%`/whitespace for a2a) — the old handlers accepted these, so an existing deployment may legitimately hold such a row, created through the Admin API. In file mode the load is all-or-nothing, so such a gateway would **fail to start** after the upgrade.

I verified the premise against the base commit: `22b8353` mcp checked only `contains("__")`, a2a only `contains('/')`.

Three ways forward:

1. **Ship as-is with a breaking-changes note.** `gh_` is already broken at runtime — every tool call against it resolves to the wrong server — and `a?b` breaks the advertised agent-card URL. Keeping them loadable only defers the failure, and `aisix validate --resources <file>` detects them before an upgrade.
2. **Narrow this PR to strict parity with the old handlers**, and do the extra tightening in its own PR with a migration story.
3. **Two-phase**: report-only first, hard rejection a release later.

I lean 1 and the PR is written that way, but it is a compatibility call rather than a technical one — happy to switch to 2 or 3.

moonming:
**Maintainer decision on the open question: option 1 — ship as-is with a breaking-changes note.**

Rationale: the newly rejected shapes are already broken at runtime, not merely untidy. An mcp server named `gh_` fails every `tools/call` against it (`gh___x` parses back to the non-existent server `gh`), and an a2a agent named `a?b` advertises a card URL whose path truncates at the `?`. Keeping them loadable only defers the failure to first use; `aisix validate --resources <file>` detects them before an upgrade.

Release-notes entry for the next release's ⚠️ Breaking changes section:

> **Stricter validation for `mcp_servers` and `a2a_agents` names and credentials.** The canonical schemas now reject: mcp server names containing `__` or ending in `_`; a2a agent names containing `/ ? # %`, whitespace, or control characters; `bearer`/`api_key` auth without a non-empty `secret`; mcp `oauth2` without `client_id` + `token_url`; and inconsistent `type`/`spec`/`api_key_header` combinations. These shapes either already misbehaved at runtime or authenticated upstream with an empty credential. **Before upgrading**, run `aisix validate --resources <file>` on a declarative gateway (an invalid file now fails startup, all-or-nothing), and check `GET /status/config`'s `rejected` array on a store-backed gateway (invalid rows are dropped individually and named there). `aisix export` now exits non-zero if any stored row was rejected, so an incomplete migration cannot pass silently.

PR #853, opened 2026-07-30
feat(config): request_body_limit_bytes defaults to 0 = no cap; uniform 413 surface

jarvis9443:
## Problem

`proxy.request_body_limit_bytes` defaulted to 10 MiB. The reference LLM proxy ships with **no** request-body cap at all (its request-size guard defaults to off), and providers accept larger requests than any fixed gateway default — Anthropic's Messages API takes requests up to 32 MB. A 10 MiB gateway default therefore rejects requests the upstream would have served (base64 PDFs and images hit this in practice), and a client that works against the provider directly breaks the moment it goes through the gateway.

Two adjacent defects in the same family, found while aligning:

1. `0` was accepted but meant "reject every request with a body" — `DefaultBodyLimit::max(0)` plus a `declared > 0` check — instead of "no cap". There was no way to express "unlimited" at all (omitting the layer entirely would still leave axum's built-in 2 MiB extractor default).
2. The 413 surface was inconsistent on the chunked/no-Content-Length path: five bare `Json<Value>` handlers (completions, images, rerank, responses, audio/speech) and the two raw-`Bytes` handlers (batches, fine-tuning) leaked axum's stock `text/plain` rejection instead of the OpenAI envelope; the multipart handlers (audio transcriptions/translations, files upload) folded axum's correctly-classified 413 into a generic 400; `/mcp` and `/a2a` returned a bare 400 for an over-cap body.

## Change

- **Default `request_body_limit_bytes` → `0` = no cap**, matching the reference proxy's out-of-box behaviour. `0` now installs `DefaultBodyLimit::disable()` (not merely a skipped layer — axum would otherwise fall back to its built-in 2 MiB), skips the Content-Length pre-check, and widens the manual `to_bytes` reads (`/mcp`, `/a2a`, `/passthrough`) accordingly. (The passthrough site was missed in the first cut — every passthrough POST got a 413 on the new default; caught by the independent audit, fixed with regression tests.) The duplicate-Content-Length rejection (request-smuggling hygiene) keeps firing regardless of the cap setting.
- **Configured limits behave exactly as before** — both enforcement layers, the drain-before-413, and the per-path envelopes are unchanged — and the 413 surface is now uniform across every inbound-body route: JSON handlers map extractor-layer 413s through `proxy_error_from_json_rejection`, `Bytes` handlers through a new `proxy_error_from_bytes_rejection`, multipart handlers preserve axum's 413 classification via `proxy_error_from_multipart`, and `/mcp` + `/a2a` discriminate the length-limit error into the standard 413 envelope.

## Behaviour changes

1. Deployments **without** an explicit `request_body_limit_bytes` in their config no longer cap request bodies (previously 10 MiB). The shipped `config.example.yaml` / `config.managed.yaml` previously carried the value explicitly, so file-based deployments that started from them keep whatever their file says; the compiled default only applies when the key is absent. Operators who want a cap set the value — everything about configured behaviour is unchanged.
2. With a configured cap, chunked over-limit requests on the seven previously-bare handlers now get the enveloped 413 (previously axum's `text/plain`), multipart over-limit uploads get 413 (previously 400), and `/mcp`/`/a2a` over-limit bodies get the enveloped 413 (previously a bare 400). Malformed JSON on the five converted handlers now also gets the enveloped 400 instead of axum's stock rejection.

## Tests

- Unit: `zero_limit_admits_bodies_over_axums_builtin_cap` (a 2.5 MiB body — over axum's built-in 2 MiB — reaches the handler on both the declared-length and chunked paths), `zero_limit_still_rejects_duplicate_content_length`, `chunked_oversize_on_v1_completions_returns_openai_envelope`, `chunked_oversize_on_v1_batches_returns_openai_envelope`, `chunked_oversize_multipart_returns_413_envelope`. The three envelope/multipart tests fail against the previous handlers (verified).
- Chunked-oversize coverage now also exists for `/passthrough`, `/mcp` and `/a2a` (the length-limit discrimination path), and `/passthrough` no longer mislabels transport faults as `RequestTooLarge`.
- E2E: new `body edges: unlimited default` suite spawns the gateway with `request_body_limit_bytes: 0` and drives a 12 MiB request through to the mock upstream; the existing 413 suite keeps its 10 MiB subject via a new dedicated harness override (the harness pins the old value so every existing with-cap assertion still tests what it always tested).

## Summary by CodeRabbit

* **New Features**
  * Request-body size limits can now be disabled by setting the limit to `0`.
  * Requests exceeding configured limits consistently return HTTP 413 responses using the standard API error format.

* **Bug Fixes**
  * Improved handling of oversized, malformed, chunked, JSON, and multipart requests across supported endpoints.
  * Non-size-related request-reading failures now return more accurate bad-request errors.

* **Tests**
  * Added coverage for unlimited large requests and standardized oversized-request responses.

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -192,24 +192,6 @@ jobs:
           docker logs smoke-aisix
           test -n "$ok"
 
-      # The profiling contract from #847: the shipped binary must stay
-      # profileable in the field, meaning the symbol table survives into
-      # the image (function-level flame graphs). A future strip step,
-      # RUSTFLAGS change, or base-image swap would break flame graphs
-      # silently, so pin it here — and print the size so every PR
-      # records the real Linux artifact cost.
-      - name: Verify shipped binary keeps its symbol table (#847)
-        if: github.event_name == 'pull_request'
-        run: |
-          set -eux
-          IMAGE="$(printf '%s\n' "${{ steps.meta.outputs.tags }}" | head -n1)"
-          cid="$(docker create "$IMAGE")"
-          docker cp "$cid:/usr/local/bin/aisix" /tmp/aisix-shipped
-          docker rm "$cid"
-          ls -l /tmp/aisix-shipped
-          test "$(nm /tmp/aisix-shipped | grep -c ' [tT] ')" -gt 1000
-          readelf -S /tmp/aisix-shipped | grep -E '\.symtab|\.eh_frame'
-
       - name: Install cosign
         if: github.event_name != 'pull_request'
         uses: sigstore/cosign-installer@v3
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -192,6 +192,24 @@ jobs:
           docker logs smoke-aisix
           test -n "$ok"
 
+      # The profiling contract from #847: the shipped binary must stay
+      # profileable in the field, meaning the symbol table survives into
+      # the image (function-level flame graphs). A future strip step,
+      # RUSTFLAGS change, or base-image swap would break flame graphs
+      # silently, so pin it here — and print the size so every PR
+      # records the real Linux artifact cost.
+      - name: Verify shipped binary keeps its symbol table (#847)
+        if: github.event_name == 'pull_request'
+        run: |
+          set -eux
+          IMAGE="$(printf '%s\n' "${{ steps.meta.outputs.tags }}" | head -n1)"
+          cid="$(docker create "$IMAGE")"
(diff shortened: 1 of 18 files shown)


jarvis9443:
@coderabbitai review

jarvis9443:
Addressed the release-note/backstop point in the paired docs PR (api7/docs#1987): the reference page now carries an explicit upgrade note (earlier releases defaulted to 10 MiB) and tells operators accepting untrusted clients directly to either set a value or enforce the limit at the LB/ingress in front. The PR description's Behaviour changes section covers the same for release notes.

PR #872, opened 2026-08-04
feat(config): lenient etcd parsing with tri-state compatibility reporting

### LINKED ISSUE #871: Config forward-compat: unknown fields from a newer CP whole-row-reject resources on older DPs — lenient parse + tri-state compat status
## Description

The DP parses every etcd resource document strictly: each resource struct carries `#[serde(deny_unknown_fields)]` and the runtime JSON Schema generated from the same struct carries `additionalProperties: false` — both the runtime validator and `schemas/resources/*.json` build from `struct_root_schema` (`crates/aisix-core/src/models/schema.rs:172`), so serde strictness and schema strictness are one artifact. A document containing any field this DP version does not know is rejected **whole-row** at the single decode chokepoint `validate_and_parse<T>` (`crates/aisix-etcd/src/loader.rs:312-352`).

That turns every additive schema change into a breaking change under the normal upgrade order (CP first, DPs rolling behind):

1. The upgraded CP adds a field to a resource; the admin edits any document of that kind.
2. An older DP receives the watch Put, fails schema validation, and **keeps serving the stale pre-edit value** — `Supervisor::apply_put` returns before mutating the snapshot (`crates/aisix-etcd/src/supervisor.rs:423-438`). The admin's change silently does not apply.
3. On the next resync or restart the snapshot is rebuilt wholesale from accepted rows only (`apply_resync`, `supervisor.rs:628-664`) and the row **vanishes**. For an `api_key` the credential stops authenticating, with a 401 byte-identical to "no such key", logged only at DEBUG (`crates/aisix-proxy/src/auth.rs:118-136`).

The failure is therefore delayed and decoupled from the action that caused it: everything looks green after the CP upgrade, then a routine DP restart days later takes the resource out. The behavior is pinned — knowingly — by the `provider_key_*_payload_currently_rejected` tests (`loader.rs:521-575`), whose names anticipate this change.

This affects **all resource kinds** dispatched by `build_snapshot` (models, api_keys, provider_keys, guardrails, cache_policies, ...), not one resource. `CachePolicy` is the one existing exception: it already parses leniently, without incident.

## Proposed design

Uniform tri-state compatibility, derived mechanically at the existing chokepoint — no per-resource rules:

- **RED — incompatible.** Lenient-schema or serde deserialization fails: field type conflict, missing required field, constraint violation (the generated schemas already carry `required` / `minimum` / `maximum`), unknown enum value. Whole-row reject as today, but: log at **ERROR** (today `warn`, `loader.rs:324,342`), report `incompatible` in the heartbeat, and retain the **last known good** typed value across resync/restart for as long as the etcd key still exists (xDS-NACK-style; removes the step-3 cliff above), with the staleness age reported every cycle.
- **YELLOW — partially compatible.** Struct parses and constraints pass but unknown fields are present: load and run with the extra fields ignored, collect their exact paths (`serde_ignored` wrapped around the existing `from_value` — one wrapper covers all kinds, nesting included), log WARN deduplicated per (kind, field-set), report `partially_compatible` **aggregated per kind + field with counts**, in a buffer separate from `RejectedEntry` so YELLOW volume cannot evict RED entries.
- **GREEN — fully compatible.** Exact match; unchanged.

Constraints that are part of the design, not optional hardening:

- **Strict write / lenient read.** Only the etcd loader becomes lenient. The self-hosted Admin API write path keeps rejecting unknown fields with 400 (apply an `additionalProperties: false`-injecting transform to the write-path validators only).
- **Deliberate closures stay closed.** The hand-injected `additionalProperties: false` on the `observability_exporter` oneOf branches (`schema.rs:816-866`) guards against plaintext-secret smuggling and must survive the migration.
- **`config_hash` must not lie.** A YELLOW row must not produce a "fully synced" hash match while enforcement differs — either exclude YELLOW rows from the hash or ship an explicit `partially_compatible[]` list next to `rejected[]` in the status surface.
- **Same-version drift stays a test failure.** CP-DP e2e at equal versions asserts zero YELLOW after convergence, preserving the typo-catching `deny_unknown_fields` performs today (`provider_key.rs:654`).
- **No silent window.** The YELLOW reporting machinery lands in the same PR as the leniency flip; there must be no interim state where unknown fields are ignored without a signal.
- **Known-load-bearing unknowns get case-by-case decisions.** The three pinned `*_currently_rejected` payloads (`aws_region` / `gcp_project_id` / `azure_resource_name`) become YELLOW loads whose runtime behavior is *not* "the old behavior" — flipping each test is a per-case decision in the same PR (prefer adding the field to the model where it is load-bearing for a provider family).

Paired CP work — **non-blocking follow-up**, filed as api7/AISIX-Cloud#1227. The CP needs no code change for this issue to ship: its heartbeat handler uses plain `json.Unmarshal` (no `DisallowUnknownFields`), so the new tri-state fields from a newer DP are tolerated as-is. The follow-up covers consuming/surfacing them:

- heartbeat consumption of the tri-state + dashboard surface + i18n;
- **save-time fleet-version warnings** — the heartbeat already carries `BUILD_VERSION`, so cp-api can warn "this environment has DPs at vX which will not enforce field Y" at the moment the admin saves. This is the only mitigation that acts *before* an admin relies on an unenforced restriction-type field. This class is real in-repo: `ApiKey.disabled` / `expires_at` were restriction fields added after first ship; under lenient parsing alone, an old DP would keep authenticating a key the dashboard shows as disabled;
- **a written policy for new enum values.** A new routing strategy / adapter / guardrail kind is *not* an unknown field — it stays RED (correctly: there is no old behavior to fall back to for a value the DP cannot interpret), and enum extension is routine roadmap work here, not a rare break. Each new value needs an explicit decision: version-gate it at the CP, or ship a `#[serde(other)]`-degradable variant where a fallback is semantically safe;
- the CP's heartbeat endpoint must itself accept unknown fields from newer DPs (the same skew problem in reverse).

## Delivery plan

Two PRs (sized against this repo's history — feature PRs here routinely land at +500~1600 lines with e2e included, e.g. the outbound-TLS and OpenAPI-to-MCP work):

**PR 1 — lenient parse + tri-state + full reporting (one atomic change).**
The leniency flip and its signal must land together (the no-silent-window constraint), and the repo norm is to ship a feature with its e2e in one PR:
- `schema.rs`: strict-write/lenient-read split (loader validates the opened schema; Admin API keeps 400 via an `additionalProperties: false`-injecting transform; the `observability_exporter` closed branches stay closed);
- drop `#[serde(deny_unknown_fields)]` (~30 sites, mechanical) + regenerate `schemas/resources/`;
- `validate_and_parse`: `serde_ignored` collects unknown-field paths -> GREEN/YELLOW/RED; YELLOW in its own buffer (never sharing the `RejectedEntry` cap); RED logs at ERROR, YELLOW at WARN deduped per (kind, field-set);
- reporting in the same PR: `/status/config` `partially_compatible[]` (aggregated per kind + field + count), heartbeat fields (the CP tolerates unknown fields today — verified), Prometheus counter, `config_hash` companion list;
- tests: characterization flips, the three `*_currently_rejected` payloads decided case by case, e2e for unknown-field-goes-YELLOW plus the zero-YELLOW-at-equal-versions invariant;
- rider: the written enum-addition policy (version-gate at CP vs `#[serde(other)]` vs accept RED) goes into the contributor guidelines.

Estimated +1200~1800 — top of the normal range, but one coherent reviewable concept; the attribute sweep inflates the count mechanically.

**PR 2 — RED last-known-good across resync/restart.**
Separable on purpose: it changes serving semantics (supervisor resync/cache/delete interplay), has its own rollback risk, and fixes a cliff that already exists today independent of the leniency flip. e2e: rejected update -> stale value keeps serving -> survives resync -> survives restart -> disappears on delete, with staleness age reported.

User-facing docs go to `api7/docs` (separate repo, not a step here). CP-side surfacing is api7/AISIX-Cloud#1227, non-blocking.

## Why

CP-before-DP is the supported rolling-upgrade order, and additive fields are the most common schema change — the current strict reader makes the common case a breaking one, with the worst failure surfacing days later at an unrelated restart. Whole-row rejection remains correct for genuine contract breaks (type changes, removed required fields), which the tri-state keeps as RED.

## Priority

High. Triggered by every CP-first upgrade that adds any resource field, across every resource kind; the failure mode is delayed (post-restart cliff), and at request time indistinguishable from an invalid credential.

## Prior art

| System | Reader behavior on unknown config fields | Reference |
|---|---|---|
| Envoy | Lenient by default for xDS-delivered config — ignore + one deduped WARN + `server.dynamic_unknown_fields` counter; strict is opt-in via `--reject-unknown-dynamic-fields` ("This allows newer xDS configurations to be delivered to older Envoys"). Hard-invalid updates are NACKed with `error_detail` and the last valid config keeps serving. | [CLI flags](https://www.envoyproxy.io/docs/envoy/latest/operations/cli), [xDS ACK/NACK](https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol) |
| Kubernetes | Readers lenient by design: kubelet may run up to 3 minors behind the API server and decodes newer objects non-strictly; strictness is write-side only (`fieldValidation=Strict\|Warn\|Ignore`, CRD pruning at admission). API rule: adding a field is compatible; changing a type requires a new API version. | [version skew](https://kubernetes.io/releases/version-skew-policy/), [field validation](https://kubernetes.io/docs/reference/using-api/api-concepts/#setting-the-field-validation-level), [API change rules](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md) |
| Kong (hybrid) | The counter-example: CP-side per-DP-version field stripping — `removed_fields.lua` carries 9 version keys / 103 field removals and grows every minor; viable only because the CP holds a websocket to each DP and knows its version. The DP itself stays strict with last-known-good fallback and reports a per-DP sync status enum (Compatible / Compatible with limitations / Incompatible). | [removed_fields.lua](https://github.com/Kong/kong/blob/master/kong/clustering/compat/removed_fields.lua), [version compatibility](https://docs.konghq.com/gateway/latest/production/deployment-topologies/hybrid-mode/#version-compatibility) |
| protobuf | proto3 restored unknown-field preservation as the default in v3.5.0 specifically for uncoordinated rolling upgrades and intermediaries; ignoring unknown fields is the documented forward-compat mechanism. (ProtoJSON, by contrast, is strict by default with an opt-in ignore.) | [updating message types](https://protobuf.dev/programming-guides/proto3/#updating), [v3.5.0 release](https://github.com/protocolbuffers/protobuf/releases/tag/v3.5.0) |
| LiteLLM | Fully lenient: router config models set `ConfigDict(extra="allow")`; invalid deployment rows are logged and skipped. | [types/router.py](https://github.com/BerriAI/litellm/blob/main/litellm/types/router.py) |
| Apache APISIX | Row-strict at the DP (`additionalProperties: false`) but with per-key last-known-good retention; unknown plugin names arriving from etcd are warn-and-skip. | [schema_def.lua](https://github.com/apache/apisix/blob/master/apisix/schema_def.lua) |
| Tyk | Fully lenient: plain `encoding/json` decode ignores unknown fields; a definition that fails decoding is logged and skipped. | [tyk](https://github.com/TykTechnologies/tyk) |

Current behavior (whole-row reject, no post-restart fallback) is stricter than every surveyed system.

<sub>Surveyed 2026-08-04; claims verified against the linked docs/source.</sub>

membphis:
Closes #871 (PR 1 of 2 — the lenient parse + tri-state + full reporting half; PR 2 covers RED last-known-good retention across resync/restart).

## Problem

Every resource struct carried `#[serde(deny_unknown_fields)]`, which also generated `additionalProperties: false` in the runtime schemas — so a document containing any field this DP version does not know was rejected whole-row at the etcd loader. Under the supported rolling-upgrade order (control plane first, data planes behind), every additive schema change became a breaking one: the admin edits a resource, an older DP rejects the watch event and keeps serving the stale value, and the next resync or restart silently drops the row entirely. For an `api_key` that means the credential stops authenticating, byte-identical to "no such key".

## Design

Uniform tri-state compatibility, derived mechanically at the existing `validate_and_parse` chokepoint — no per-resource rules:

- **RED — incompatible.** Lenient-schema or serde deserialization fails: type conflict, missing required field, constraint violation, unknown enum value, or an unknown field inside a closed tagged-enum shape. Whole-row reject as before, now logged at ERROR and reported as `rejected[]` (unchanged surface).
- **YELLOW — partially compatible.** The struct parses and all constraints pass, but unknown fields are present: the row loads and serves with those fields ignored, and `serde_ignored` collects the exact paths (array indices normalized to `[]`, `Option` layers stripped). Logged WARN, deduplicated per (kind, field-set).
- **GREEN — fully compatible.** Exact match; unchanged.

### Strict write / lenient read

Only the etcd loader becomes lenient. Strictness moved out of the structs into a mechanical `close_unknown_fields` pass over the shared schema producers: the write paths (Admin API, standalone file source) compile the closed set and keep returning 400 on unknown fields, while the loader compiles the open set. Both sets build from the same producers, so they cannot drift field-wise.

**The published `schemas/resources/*.json` are byte-identical after regeneration** — the closing transform provably reproduces what `deny_unknown_fields` used to generate, and the CI schema-drift gate doubles as a write-contract-unchanged check. The published files deliberately keep the strict shape: they feed the Admin API reference, which documents the write contract.

### Deliberate closures stay closed

The hand-injected `additionalProperties: false` on the observability-exporter branches and the guardrail tagged sub-enums survives in **both** validator sets. Two reasons, verified empirically: those branches guard the `credential_ref` indirection against plaintext-secret smuggling, and serde silently swallows unknown fields inside inline-tagged enum content (`serde_ignored` cannot see through the Content buffering) — an open schema there would be an unreportable, silent tolerance, violating the no-silent-window constraint. Unknown fields in those subtrees therefore stay RED, same stance as unknown enum values.

### Reporting (landed atomically with the flip — no silent window)

- `GET /status/config` gains `partially_compatible[]` next to `rejected[]`, aggregated per (kind, field) with row counts. This is the `config_hash` companion the issue requires: YELLOW rows are served (so the hash covers them and the state stays `synced`), and the explicit list is what distinguishes "fully synced" from "synced with fields this version does not enforce".
- The managed-mode heartbeat gains `partially_compatible_resources` (omitted when empty; the CP heartbeat handler was verified to tolerate unknown fields, so no paired CP change is needed to ship — surfacing is api7/AISIX-Cloud#1227).
- Prometheus gains `aisix_config_partially_compatible_resources{kind}`, mirroring the rejected-resources gauge including stale-label zeroing. Field paths stay off the labels (not a bounded set); per-field detail lives in `/status/config`.
- The supervisor retains YELLOW state per etcd key in its own capped buffer (1024, WARN on truncation) — never sharing the 256-cap rejection buffer, so YELLOW volume cannot evict RED entries. A rejected update for a YELLOW key keeps the signal (the previously loaded value still serves); a delete or an exact-match re-put clears it; a resync replaces it wholesale.

### The three pinned `*_currently_rejected` payloads

Flipped to partially-compatible loads, per-case: the Bedrock and Vertex bridges read region/project from the credential JSON inside `api_key`, and the Azure bridge derives its resource from `api_base` with a pinned API version — none of the top-level `aws_region` / `gcp_project` / `gcp_region` / `azure_resource_name` / `api_version` fields is consumed by this DP, so a dead model field would be strictly worse than an ignored-and-reported one.

### Enum-addition policy (rider)

CONTRIBUTING.md now documents the rollout rule: added fields are the safe, reported-as-partial change; a new enum value is **not** forward compatible by design (no old behavior exists for a value the DP cannot interpret) and each addition needs an explicit decision — version-gate at the CP, a `#[serde(other)]` fallback only where semantically safe, or an accepted loud rejection.

## Ecosystem alignment

Surveyed mainstream implementations (primary sources linked from #871): service proxies consuming dynamically-delivered config default to ignoring unknown fields with a deduplicated warning and a counter, keeping strictness opt-in; the major container orchestrator makes readers lenient by several minor versions and puts strictness on the write side only; hybrid-mode API gateways either report a per-DP compatibility status enum or strip unknown fields CP-side per DP version; the dominant serialization framework restored unknown-field preservation as its default specifically for uncoordinated rolling upgrades. The previous behavior here (whole-row reject, no reporting) was stricter than every surveyed system; this PR lands on the common ground — lenient read, loud reporting, write-side strictness — plus the per-DP compatibility status idea for the heartbeat.

## Scope notes

- The standalone file source stays strict: a hand-edited resources file has no version-skew writer, so strictness there is pure typo protection.
- `unknown_kind` (an entirely new resource kind from a newer CP) keeps its existing behavior (skipped + reported), unchanged by this PR.
- No CP change required to merge: the CP tolerates the new heartbeat field today; consuming/surfacing it is tracked in api7/AISIX-Cloud#1227.

## Testing

- The reproducing test landed first (red on the old behavior): an `api_key` document with an unknown field must be accepted and reported partially compatible.
- Loader unit tests: root and nested unknown-field YELLOW with exact dotted paths, cross-row aggregation, unknown-enum-value stays RED, the three provider_key flips.
- Supervisor unit tests: YELLOW retention across put/replace/clear/delete/resync, and survival when an update for the same key is rejected.
- Heartbeat wiremock tests: field present when wired, omitted when empty (historical body shape preserved).
- Metrics tests: gauge rendering and stale-label zeroing.
- E2E against the real binary + etcd (new `config-forward-compat-e2e.test.ts`): the unknown-field `api_key` authenticates a real chat; `partially_compatible[]` + gauge report it; an unknown routing strategy stays rejected; deleting the row clears the report and zeroes the gauge; a converged same-version config reports zero YELLOW (typo-catching preserved); the Admin API still 400s on unknown fields.
- Full suite: `cargo test --workspace` green; full e2e suite green (175 files, 461 tests); `cargo fmt --check`, `clippy --workspace`, `dump-schema` (zero diff), and `dump-openapi` all clean.

## Summary by CodeRabbit

* **New Features**
  * Configuration reads now tolerate unknown fields for forward compatibility while continuing to reject invalid values and unsupported enum entries.
  * Partial compatibility is reported through configuration status, metrics, and heartbeat data.
  * Strict validation remains enforced for configuration writes.
* **Documentation**
  * Added guidance on schema evolution, compatibility policies, and enum rollout strategies.
* **Tests**
  * Added end-to-end coverage for compatible reads, status reporting, metrics, cleanup, and rejected writes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,7 +59,11 @@ different discipline:
   `aisix_config_partially_compatible_resources` metric). Never assume a new
   field is enforced fleet-wide until every data plane runs a version that
   knows it — this matters most for restriction-type fields (an old gateway
-  keeps allowing what the new field would forbid).
+  keeps allowing what the new field would forbid). Two kinds are exempt from
+  this tolerance: `guardrail` and `observability_exporter` documents flatten
+  their fields into closed tagged shapes, so ANY new field there still
+  whole-row rejects on older gateways — treat additions to those two like
+  enum values below.
 - **Adding an enum value** (a routing strategy, an adapter, a guardrail
   `kind`, …) is NOT forward compatible, by design: a value the gateway cannot
   interpret has no old behavior to fall back to, so the whole document stays
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,6 +46,40 @@ still build and pass its structural checks
    implementation details.
 4. Make sure the checks above pass locally before pushing.
 
+### Changing resource models: unknown fields vs. new enum values
+
+The gateway reads its resources leniently from etcd and strictly on write
+(issue #871). That split makes two kinds of schema change behave very
+differently on a data plane that has not been upgraded yet, and each needs a
+different discipline:
+
+- **Adding a field** is the safe, expected change. An older gateway loads the
+  document with the new field ignored and reports it as partially compatible
+  (`GET /status/config` `partially_compatible[]`, the heartbeat, and the
+  `aisix_config_partially_compatible_resources` metric). Never assume a new
+  field is enforced fleet-wide until every data plane runs a version that
+  knows it — this matters most for restriction-type fields (an old gateway
+  keeps allowing what the new field would forbid).
+- **Adding an enum value** (a routing strategy, an adapter, a guardrail
+  `kind`, …) is NOT forward compatible, by design: a value the gateway cannot
+  interpret has no old behavior to fall back to, so the whole document stays
+  rejected on older versions. Do not "fix" that by opening the enum. Every
+  new enum value needs an explicit rollout decision, made in the PR that adds
+  it:
+  1. **Version-gate at the control plane** (preferred for values that change
+     serving behavior): the control plane only offers the value once the
+     environment's data planes are on a version that knows it, using the
+     version the heartbeat already reports.
+  2. **Ship a degradable fallback** via `#[serde(other)]` — only when a
+     fallback is semantically safe (e.g. an advisory label where "unknown"
+     is a reasonable interpretation). Never for values that select serving
+     behavior: silently running a different routing strategy than configured
+     is worse than rejecting the document.
+  3. **Accept the rejection** for values that are new capabilities: an old
+     gateway that cannot serve the capability rejecting the row loudly (the
+     rejection reaches `rejected[]` and the heartbeat) can be the correct
+     outcome — state which of the three you chose and why in the PR.
+
 ### Commit and PR style
 
 Commit subjects follow Conventional Commits, matching the existing history:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -126,6 +126,7 @@ dependencies = [
  "etcd-client",
  "futures",
  "serde",
+ "serde_ignored",
  "serde_json",
  "tempfile",
  "thiserror 1.0.69",
@@ -4374,6 +4375,16 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn",
+]
+
+[[package]]
+name = "serde_ignored"
+version = "0.1.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "115dffd5f3853e06e746965a20dcbae6ee747ae30b543d91b0e089668bb07798"
+dependencies = [
+ "serde",
+ "serde_core",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -81,6 +81,10 @@ webpki-roots = "0.26"
 # Serialization / validation
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
+# Collects the paths of fields serde ignored during deserialization —
+# how the etcd loader reports unknown fields from a newer control plane
+# as "partially compatible" instead of dropping them silently (#871).
+serde_ignored = "0.1"
 jsonschema = "0.28"
 schemars = "0.8"
 
--- a/crates/aisix-admin/src/lib.rs
+++ b/crates/aisix-admin/src/lib.rs
@@ -843,9 +843,7 @@ mod tests {
                 field: "quota_profile".into(),
                 count: 2,
             }],
-            partially_compatible_rows_by_kind: [("api_keys".to_string(), 2)]
-                .into_iter()
-                .collect(),
+            partially_compatible_rows_by_kind: [("api_keys".to_string(), 2)].into_iter().collect(),
             is_reload: true,
             wholly_rejected: false,
         });
--- a/crates/aisix-admin/src/lib.rs
+++ b/crates/aisix-admin/src/lib.rs
@@ -800,6 +800,8 @@ mod tests {
                 resource_counts: Default::default(),
             }),
             rejected: vec![],
+            partially_compatible: Vec::new(),
+            partially_compatible_rows_by_kind: Default::default(),
             is_reload: true,
             wholly_rejected: false,
         });
@@ -836,6 +838,14 @@ mod tests {
                 last_error: "schema validation failed at `/display_name`".into(),
                 seen_at: chrono::Utc::now(),
             }],
+            partially_compatible: vec![aisix_core::config_status::PartialCompatResource {
+                resource_kind: "api_keys".into(),
+                field: "quota_profile".into(),
+                count: 2,
+            }],
+            partially_compatible_rows_by_kind: [("api_keys".to_string(), 2)]
+                .into_iter()
+                .collect(),
             is_reload: true,
             wholly_rejected: false,
         });
@@ -866,6 +876,12 @@ mod tests {
         assert_eq!(v["applied"]["resource_counts"]["models"], 1);
         assert_eq!(v["rejected"][0]["resource_kind"], "models");
         assert_eq!(v["rejected"][0]["last_error_kind"], "schema_failed");
+        // The partially-compatible companion list (#871) rides next to
+        // rejected[] so a matching config_hash can't hide that some
+        // served rows carry fields this DP does not enforce.
+        assert_eq!(v["partially_compatible"][0]["resource_kind"], "api_keys");
+        assert_eq!(v["partially_compatible"][0]["field"], "quota_profile");
+        assert_eq!(v["partially_compatible"][0]["count"], 2);
     }
 
     #[tokio::test]
@@ -883,6 +899,8 @@ mod tests {
                 resource_counts: [("models".to_string(), 2)].into_iter().collect(),
             }),
             rejected: vec![],
+            partially_compatible: Vec::new(),
+            partially_compatible_rows_by_kind: Default::default(),
             is_reload: true,
             wholly_rejected: false,
         });
--- a/crates/aisix-core/src/bin/dump-schema.rs
+++ b/crates/aisix-core/src/bin/dump-schema.rs
@@ -97,8 +97,21 @@ fn dump<T: JsonSchema>(out_dir: &Path, name: &str) {
 }
 
 /// Insert `additionalProperties: false` on a struct-shaped schema object
-/// (one that lists `properties`), unless it already pins a value.
+/// (one that lists `properties`), unless it already pins a value. Recurses
+/// into `anyOf` branches so an untagged enum's object variant closes too —
+/// serde silently swallows unknown fields inside untagged content, so the
+/// schema closure is the only non-silent guard there (the resource
+/// producers apply the same rule, e.g. `OnEmbeddingFailure` in `model`).
 fn close_object_schema(schema: &mut schemars::schema::SchemaObject) {
+    if let Some(sub) = schema.subschemas.as_deref_mut() {
+        if let Some(any_of) = sub.any_of.as_mut() {
+            for branch in any_of.iter_mut() {
+                if let schemars::schema::Schema::Object(b) = branch {
+                    close_object_schema(b);
+                }
+            }
+        }
+    }
     let Some(object) = schema.object.as_deref_mut() else {
         return;
     };
--- a/crates/aisix-core/src/bin/dump-schema.rs
+++ b/crates/aisix-core/src/bin/dump-schema.rs
@@ -38,38 +38,35 @@ fn main() {
     fs::create_dir_all(&out_dir).expect("create schemas/resources dir");
 
     // Every resource with a runtime validator goes through the SAME
-    // `*_root_schema()` producer the validator uses, so the published schema ==
-    // the enforced schema by construction. `ensemble`/`rate_limit`/`routing`
-    // have no standalone validator (they are nested struct types) so they dump
-    // straight from the struct via `schema_for!`.
-    dump_value(&out_dir, "api_key", schema::apikey_root_schema());
-    dump_value(&out_dir, "cache_policy", schema::cache_policy_root_schema());
-    dump_value(&out_dir, "model", schema::model_root_schema());
-    dump_value(
-        &out_dir,
+    // `resource_root_schema(name, strict: true)` producer the write-path
+    // validators compile, so the published schema == the enforced write
+    // contract by construction. The published files deliberately carry the
+    // STRICT shape: they document the Admin API write contract (unknown
+    // fields are a 400) and the etcd loader's lenient read tolerance is a
+    // runtime behavior, not a contract callers may write against.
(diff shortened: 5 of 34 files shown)


membphis:
## Independent audit — outcome

A fresh, zero-context audit agent reviewed this PR per the repository merge gate. Verdict: no HIGH findings; all core claims verified (lenient parse mechanics, strict-write preservation via schema regeneration, separate YELLOW retention, all three reporting surfaces, CP-side verification that the flipped provider_key fields are genuinely unconsumed and that the CP heartbeat handler tolerates the new field). 4 MEDIUM and 4 LOW findings — **all fixed in code** in 7f3bc88:

| Finding | Disposition |
|---|---|
| MEDIUM-1: CP writes `env_id` on every guardrail_attachment → permanent false YELLOW on same-version managed fleets | Fixed: `env_id` declared on the model as known-and-ignored (documented in the published schema). A full CP marshal-site sweep confirmed this is the **only** CP-written field the DP models did not know, across all 12 kinds. Loader test added with the real CP projection shape. |
| MEDIUM-2: untagged `OnEmbeddingFailure` object variant silently tolerated unknown fields on BOTH paths (pre-existing) | Fixed: object branch closed in the producer for both validator sets, plus the standalone nested-schema dump; tests pin strict AND lenient rejection. This is a deliberate write-path tightening (a typo there was previously accepted and silently dropped). |
| MEDIUM-3: no test pinned the deliberate closures on the lenient set | Fixed: `lenient_set_keeps_deliberate_closures_closed` + strict/lenient split tests added. |
| MEDIUM-4: unbounded per-row field reporting (logs / status JSON / heartbeat body) | Fixed: 64-field cap per row with a visible `...truncated` sentinel; test added. |
| LOW-1 poisonable WARN-dedup lock | Fixed: poison-tolerant lock recovery. |
| LOW-2 path-normalization aliasing | Documented on the normalizer (lossy by design; per-kind field name is the actionable signal). |
| LOW-3 CONTRIBUTING overstated "adding a field is safe" | Fixed: `guardrail` / `observability_exporter` named as exempt kinds (closed tagged shapes — treat additions like enum values). |
| LOW-4 YELLOW journey e2e covered one kind | Fixed: e2e now also serves chat through a model document carrying an unknown field. |

Note on the "byte-identical schemas" claim in the description: it held exactly for the leniency flip itself (proving the closing transform reproduces `deny_unknown_fields` output). The audit-fix commit then makes three **deliberate** schema changes: the documented `env_id` on `guardrail_attachment` (open resource — write behavior unchanged) and the `OnEmbeddingFailure` object-branch closure in `model` + `semantic` (the MEDIUM-2 tightening).

Verification after fixes: `cargo test --workspace` green, clippy/fmt clean, forward-compat + status-config + guardrail e2e suites green against real etcd.

PR #874, opened 2026-08-04
feat(config): serve last-known-good values for rejected keys across resync and restart

### LINKED ISSUE #871: Config forward-compat: unknown fields from a newer CP whole-row-reject resources on older DPs — lenient parse + tri-state compat status

membphis:
Part 2 of #871 (the delivery plan's "PR 2 — RED last-known-good across resync/restart"). PR 1 (lenient parse + tri-state reporting) landed as #872.

## Problem

A watch put that fails validation already leaves the previous good value serving — but only until the next full resync or restart, which rebuild the snapshot from accepted rows only. A key whose latest etcd bytes are rejected then **vanishes**: an `api_key` stops authenticating with a 401 byte-identical to "no such key", days after and decoupled from the write that broke it. This is the step-3 "cliff" in #871, present for every resource kind.

## Design

**Retention (xDS-NACK style).** The supervisor pins the serving bytes as the key's *last known good* the moment a replacement is rejected (`capture_last_good` on the rejected-put path; the same capture runs inside `apply_resync` for keys first observed rejected there). On every resync the pinned bytes re-build through the normal loader and merge into the fresh snapshot, so the typed value and every derived signal stay consistent with what actually serves. A pinned value the running build can no longer parse (a DP downgrade) drops its retention with an ERROR — the same contract as any RED row.

**Retention ends exactly when the etcd key stops existing or heals.** A successful put for the key, a watch delete, or a resync that no longer carries the key (deletion observed via full re-read after reconnect/compaction) all remove the pin — no zombie config can outlive its key. Pinned values and their staleness clocks ride in the snapshot cache (additive `stale` section, format version unchanged: old DPs ignore it, new DPs default it), so retention survives restarts.

**Design decision 1 — what `config_hash` / `resource_counts` say about a rejected-but-serving-stale row.** `config_hash` covers the bytes each key *actually serves*: observed etcd bytes for accepted keys, the **pinned last-known-good bytes** for stale-serving keys, and nothing for a rejected key with no last good. The hash therefore never claims the rejected bytes applied (a CP diffing its expected hash sees divergence) and never claims the row stopped serving (two DPs — one serving a pin, one that never had the value — report different hashes). The substitution is decided by the stale map, **not** the capped rejection buffer, so buffer overflow cannot flip a served key's hash contribution. `resource_counts` counts served rows, pins included — they are serving — with the stale detail carried separately (`rejected[]` per-row + a per-kind gauge), and `state` stays `degraded` while any rejection exists. Pre-existing boundary, unchanged in scope: a rejected key with no pin that has been evicted from the 256-cap rejection buffer folds its observed bytes back into `config_hash`; this predates this PR and now applies uniformly to put- and resync-observed rejections.

**Design decision 2 — coexistence with PR 1's YELLOW retention and the 256-cap rejection buffer.** The same key can simultaneously be "new value rejected (RED)" and "serving old value that is itself YELLOW". Both signals coexist and describe different bytes: `rejected[]` describes the observed etcd value, `partially_compatible[]` describes the served pin — on injection the pin re-runs `serde_ignored`, so its ignored-field paths are regenerated from the same bytes that serve (pinned by `stale_last_good_that_was_yellow_keeps_its_partial_compat_signal`). The stale map is a third, deliberately **uncapped** store, independent of both the 256-cap RED buffer and the 1024-cap YELLOW buffer: evicting there truncates a report; evicting a pin would take a served resource offline. Its size is bounded by rows that ever loaded successfully — a subset of the (uncapped) served snapshot, one extra copy of the raw bytes per rejected row.

**Staleness is reported every cycle, on the PR 1 surfaces.** `GET /status/config` `rejected[]` entries gain `serving_stale_since` (RFC3339) and `serving_stale_age_seconds` (recomputed per read); absent both when nothing serves for the row, so consumers can tell "serving stale" from "actually down". Heartbeat `rejected_resources[]` entries gain `stale_serving_since_unix_secs` (omitted when absent; the CP heartbeat handler tolerates unknown fields — verified in #872). Metrics gain `aisix_config_stale_served_resources{kind}` with the same stale-label zeroing as the PR 1 gauge. The `since` instant persists in the cache, so the age is continuous across restarts instead of resetting at boot.

**Behavior change: a rejected live put now enters the observed-state map.** Previously the bytes only reached `source_hash` at the next resync; now `source_hash` reflects the observed etcd state immediately and — the load-bearing part — a restart *inside* the rejected-put window restores the same rejected-bytes + pinned-value cache shape a post-resync restart would, keeping the staleness clock continuous. The e2e restart leg caught this gap live (the clock reset by 2s across the restart before the fix).

**Config: `managed.snapshot_cache_path` becomes optional.** Omitted keeps today's defaults exactly (managed on at `/var/lib/aisix/config_cache.json`, self-hosted etcd off); an explicit path now enables the cache in self-hosted etcd mode too (previously silently ignored there — anyone who set it wanted it); `""` still disables everywhere. This is DP-local `config.yaml`, not a CP-managed resource, so no `cp-admin.yaml` counterpart applies.

## Testing

Reproducing tests were written first and failed on the pre-PR code at the resync cliff (row count 1 → 0), then flipped green:

- unit: rejected update keeps serving across resync (with staleness fields asserted via the JSON view), across restart (cache round-trip), dies on watch delete AND on resync-observed absence, YELLOW pin keeps its ignored-field report, `config_hash` equals the served-bytes hash and diverges from `source_hash`, pin persists for an immediate restart with a continuous clock, cache format round-trips the stale section and still loads legacy files.
- e2e (real etcd + real chat traffic): seeded model serves → schema-invalid replacement lands in etcd → chat still 200 with `serving_stale_since`/age reported and the gauge at 1 → full process restart on the same prefix + cache (a sentinel write proves the live load_all/resync completed against the still-rejected bytes) → chat still 200, `serving_stale_since` byte-identical across the restart → etcd delete → chat 404, `rejected[]` empty, gauge zeroed.
- regressions: full workspace suite, PR 1's forward-compat e2e, and the status-surface e2e files all pass; clippy and fmt clean.

## Prior art

The issue's prior-art table (#871) surveys mainstream proxies/gateways and config-delivery systems: the dominant pattern for hard-invalid dynamic config is NACK-with-last-valid-config-retained (the xDS ACK/NACK model), and per-key last-known-good retention at the data plane appears in the surveyed gateway with hybrid CP/DP deployment and in the row-strict gateway of the same family. This PR lands that shape: whole-row rejection stays (RED is still RED), but rejection of an *update* no longer takes down the running value, and the divergence is continuously observable.

## Out of scope / follow-ups

- CP-side surfacing of the tri-state + staleness fields remains api7/AISIX-Cloud#1227 (non-blocking; the CP tolerates the new fields today).
- User-facing docs go to the docs repository per repo policy.

Closes #871.

## Summary by CodeRabbit

* **New Features**
  * Rejected configuration updates now continue serving the last known good values until the affected entries are removed or successfully updated.
  * Stale-serving status includes start time, elapsed age, and affected resource counts.
  * Stale configuration state persists across resynchronization and application restarts.
  * Snapshot caching can now be enabled, disabled, or configured explicitly across deployment modes.

* **Bug Fixes**
  * Configuration hashes now accurately reflect the values currently being served.
  * Heartbeat and status responses report stale-serving details for rejected resources.

### DIFF
--- a/crates/aisix-admin/src/lib.rs
+++ b/crates/aisix-admin/src/lib.rs
@@ -802,6 +802,7 @@ mod tests {
             rejected: vec![],
             partially_compatible: Vec::new(),
             partially_compatible_rows_by_kind: Default::default(),
+            stale_served_rows_by_kind: Default::default(),
             is_reload: true,
             wholly_rejected: false,
         });
@@ -837,6 +838,7 @@ mod tests {
                 last_error_kind: "schema_failed".into(),
                 last_error: "schema validation failed at `/display_name`".into(),
                 seen_at: chrono::Utc::now(),
+                serving_stale_since: None,
             }],
             partially_compatible: vec![aisix_core::config_status::PartialCompatResource {
                 resource_kind: "api_keys".into(),
@@ -844,6 +846,7 @@ mod tests {
                 count: 2,
             }],
             partially_compatible_rows_by_kind: [("api_keys".to_string(), 2)].into_iter().collect(),
+            stale_served_rows_by_kind: Default::default(),
             is_reload: true,
             wholly_rejected: false,
         });
@@ -899,6 +902,7 @@ mod tests {
             rejected: vec![],
             partially_compatible: Vec::new(),
             partially_compatible_rows_by_kind: Default::default(),
+            stale_served_rows_by_kind: Default::default(),
             is_reload: true,
             wholly_rejected: false,
         });
--- a/crates/aisix-core/src/config.rs
+++ b/crates/aisix-core/src/config.rs
@@ -264,7 +264,8 @@ pub struct ManagedConfig {
     /// enables the cache in either mode — self-hosted etcd deployments
     /// gain the same offline resilience by opting in. Empty string
     /// disables persistence everywhere — useful for ephemeral test runs
-    /// where you don't want a stale cache to mask a real failure.
+    /// where you don't want a stale cache to mask a real failure. A
+    /// bare `snapshot_cache_path:` (YAML null) is treated as omitted.
     #[serde(default)]
     pub snapshot_cache_path: Option<String>,
 
--- a/crates/aisix-core/src/config.rs
+++ b/crates/aisix-core/src/config.rs
@@ -258,10 +258,15 @@ pub struct ManagedConfig {
     /// so the proxy can serve traffic from cached config across CP
     /// outages and full container restarts.
     ///
-    /// Empty string disables persistence — useful for ephemeral test
-    /// runs where you don't want a stale cache to mask a real failure.
-    #[serde(default = "ManagedConfig::default_snapshot_cache_path")]
-    pub snapshot_cache_path: String,
+    /// When the field is omitted, managed mode uses
+    /// `/var/lib/aisix/config_cache.json` and self-hosted etcd mode
+    /// leaves persistence off (unchanged defaults). Setting a path
+    /// enables the cache in either mode — self-hosted etcd deployments
+    /// gain the same offline resilience by opting in. Empty string
+    /// disables persistence everywhere — useful for ephemeral test runs
+    /// where you don't want a stale cache to mask a real failure.
+    #[serde(default)]
+    pub snapshot_cache_path: Option<String>,
 
     /// Heartbeat interval, in seconds. The DP POSTs a heartbeat to
     /// dp-manager every `heartbeat_interval_secs`; CP surfaces a DP as
@@ -296,14 +301,27 @@ impl ManagedConfig {
         has_pem || has_file
     }
 
+    /// Resolve the snapshot-cache path per the field docs: an explicit
+    /// path wins in any mode, an explicit empty string disables, and an
+    /// omitted field means "the default path in managed mode, disabled
+    /// in self-hosted etcd mode".
+    pub fn effective_snapshot_cache_path(&self) -> Option<&str> {
+        match self.snapshot_cache_path.as_deref() {
+            Some("") => None,
+            Some(path) => Some(path),
+            None if self.is_managed() => Some(Self::DEFAULT_SNAPSHOT_CACHE_PATH),
+            None => None,
+        }
+    }
+
+    /// Default on-disk snapshot cache location for managed mode.
+    pub const DEFAULT_SNAPSHOT_CACHE_PATH: &'static str = "/var/lib/aisix/config_cache.json";
+
     fn default_mtls_dir() -> String {
         "/var/lib/aisix/mtls".into()
     }
     fn default_dp_id_file() -> String {
         "/var/lib/aisix/dp_id".into()
-    }
-    fn default_snapshot_cache_path() -> String {
-        "/var/lib/aisix/config_cache.json".into()
     }
     const fn default_heartbeat_interval_secs() -> u64 {
         15
@@ -2091,13 +2109,49 @@ managed:
         assert_eq!(cfg.managed.mtls_dir, "/var/lib/aisix/mtls");
         assert_eq!(cfg.managed.dp_id_file, "/var/lib/aisix/dp_id");
         // Default snapshot cache path keeps offline-resilience on by
-        // default; operators opt out by setting the field to "".
+        // default in managed mode; operators opt out by setting the
+        // field to "".
         assert_eq!(
-            cfg.managed.snapshot_cache_path,
-            "/var/lib/aisix/config_cache.json",
+            cfg.managed.effective_snapshot_cache_path(),
+            Some("/var/lib/aisix/config_cache.json"),
         );
         // CP URL comes from env at runtime — empty here is fine.
         assert!(cfg.managed.cp_base_url.is_none());
+    }
+
+    /// #871: the snapshot cache resolves per mode — managed defaults on,
+    /// self-hosted etcd defaults off, an explicit path enables either,
+    /// an explicit "" disables either.
+    #[test]
+    fn snapshot_cache_path_resolution_per_mode() {
+        let mut managed = ManagedConfig {
+            enabled: true,
+            ..Default::default()
+        };
+        assert_eq!(
+            managed.effective_snapshot_cache_path(),
+            Some(ManagedConfig::DEFAULT_SNAPSHOT_CACHE_PATH),
+        );
+        managed.snapshot_cache_path = Some(String::new());
+        assert_eq!(managed.effective_snapshot_cache_path(), None);
+        managed.snapshot_cache_path = Some("/tmp/cache.json".into());
+        assert_eq!(
+            managed.effective_snapshot_cache_path(),
+            Some("/tmp/cache.json"),
+        );
+
+        let mut self_hosted = ManagedConfig {
+            enabled: false,
+            ..Default::default()
+        };
+        assert_eq!(self_hosted.effective_snapshot_cache_path(), None);
+        self_hosted.snapshot_cache_path = Some("/tmp/cache.json".into());
+        assert_eq!(
+            self_hosted.effective_snapshot_cache_path(),
+            Some("/tmp/cache.json"),
+        );
+        self_hosted.snapshot_cache_path = Some(String::new());
+        assert_eq!(self_hosted.effective_snapshot_cache_path(), None);
     }
 
     #[test]
--- a/crates/aisix-core/src/config_status.rs
+++ b/crates/aisix-core/src/config_status.rs
@@ -142,6 +142,17 @@ pub struct RejectedResource {
     pub first_seen_at: String,
     /// RFC3339 UTC timestamp the rejection was most recently observed.
     pub last_seen_at: String,
+    /// RFC3339 UTC timestamp since when this resource has been serving its
+    /// last known good value instead of the rejected bytes (issue #871).
+    /// Absent when nothing is serving for this resource — the row either
+    /// never loaded successfully or its retention was dropped.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub serving_stale_since: Option<String>,
+    /// Seconds elapsed since `serving_stale_since`, recomputed on every
+    /// read so the staleness age is reported every cycle. Absent together
+    /// with `serving_stale_since`.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub serving_stale_age_seconds: Option<u64>,
 }
 
 /// One rejected entry handed to [`ConfigStatus`] by a load path. `identity`
@@ -155,6 +166,10 @@ pub struct IncomingRejection {
     pub last_error_kind: String,
     pub last_error: String,
     pub seen_at: DateTime<Utc>,
+    /// When set, the resource is still serving its last known good value
+    /// (pinned before this rejection) and this is the instant that stale
+    /// serving began (#871). `None` means nothing serves for this row.
+    pub serving_stale_since: Option<DateTime<Utc>>,
 }
 
 /// One partially compatible observation, aggregated per (kind, field):
@@ -207,6 +222,11 @@ pub struct LoadObservation {
     /// Row-based (a resource with two unknown fields counts once), for
     /// the `aisix_config_partially_compatible_resources` gauge.
     pub partially_compatible_rows_by_kind: BTreeMap<String, usize>,
+    /// Served resources per kind whose latest source bytes are rejected
+    /// and whose last known good value serves instead (#871), for the
+    /// `aisix_config_stale_served_resources` gauge. The per-resource
+    /// detail (which row, since when) rides on `rejected[]`.
+    pub stale_served_rows_by_kind: BTreeMap<String, usize>,
     /// Whether this load counts as a config reload for
     /// `aisix_config_reloads_total` (full (re)syncs and file loads do;
     /// incremental etcd events do not).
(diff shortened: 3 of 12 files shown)


membphis:
## Independent audit (repo guideline 8)

A fresh, zero-context audit agent reviewed this PR against the six contract claims in the description (correctness, reliability, security, sensitive-info leakage, breaking changes, e2e coverage), re-running the unit suites and the e2e against a live etcd. Findings and disposition:

**MEDIUM — deleted never-serving key leaked into observed state (FIXED in 480dbca).** Because a rejected put now mirrors its bytes into the observed-state map, a watch delete for a key that never served took the not-present early return in `apply_delete`, which cleared the rejection and pin but left the key's bytes in `state` and skipped the cache flush: `source_hash` kept a phantom key until the next resync, the deleted document persisted in the cache file (secret-bearing for `provider_keys`), and a cache-restored boot could resurrect the rejection. The state removal now runs on both branches and the not-present branch syncs + flushes when it removed anything; pinned by `deleting_a_rejected_never_serving_key_clears_observed_state`.

**LOW — kind-dispatch drift guard (FIXED in 480dbca).** `merge_snapshot` and `snapshot_has` (both introduced by this PR) now exhaustively destructure `AisixSnapshot`, so adding a resource kind without wiring the pin/inject paths is a compile error instead of a silently unpinnable kind. The pre-existing dispatch sites (`apply_delete`'s remove match, `resource_counts`) are unchanged to keep this PR's diff scoped; they are covered by the existing every-kind regression tests.

**LOW — `snapshot_cache_path:` YAML null (FIXED in 480dbca, docs).** A bare key previously failed deserialization and now reads as omitted; the field docs state this explicitly. Behavior direction is benign (managed default on, self-hosted off).

**LOW — unordered spawned cache flushes (JUSTIFIED, not fixed).** `flush_cache` spawns each write and the cache's write lock serialises but does not order them, so on a slow disk an older snapshot can land last; a crash in that window restores slightly stale state, now including a possibly missing pin (the staleness clock would then restart at the next resync). This mechanism predates this PR, is self-healing at the first live cycle, and an ordering fix (flush sequence numbers checked under the write lock) is a standalone change to the pre-existing cache writer — deferring rather than growing this PR's blast radius.

All other angles came back clean: the capture invariant held against constructed adversarial sequences, single-lock discipline verified, the cache's `stale` section is the same data class in the same file as the pre-existing entries (no new secret exposure), new wire fields are timestamps only and strictly additive, cache format compatibility verified in both directions, and the e2e restart leg proves a live resync (sentinel write) rather than cache replay. Audit verdict after fixes: no open HIGH/MEDIUM.

Issue #451, opened 2026-05-29
security: prevent unauthenticated metric-label cardinality DoS


juzhiyuan:
## Summary

Proxy metrics currently use the raw request path as the `endpoint` label before authentication. Unauthenticated callers can create unbounded time series through unique passthrough paths.

## Source

- Imported from Codex Security scan results shared in Slock `#Dev:bc985f48` (`chatgpt.com.har`).
- This issue groups findings with the same engineering boundary to avoid fragmented fixes.
- Triage priority: **P1**.

## Covered scanner findings

- [ ] **#11** `Unbounded metrics labels allow unauthenticated proxy DoS` — `request.uri().path()` is used as a metrics label before auth and includes attacker-controlled passthrough suffixes.

## Expected fix direction

- [ ] Normalize metric endpoint labels to a fixed route template such as `/passthrough/:provider/*rest` before recording.
- [ ] Avoid recording high-cardinality labels derived from unauthenticated raw paths.
- [ ] Verify zero-value gauge series are removed or bounded if the metrics backend retains label sets.
- [ ] Add a regression test that many unique unauthenticated passthrough paths produce bounded metric series.

## Definition of done

- [ ] Fix is covered by regression tests for the affected mode/provider/endpoint.
- [ ] The fix is applied at the shared boundary where possible, not only at one provider call site.
- [ ] Security behavior is documented in code comments or docs where operator expectations change.

PR #888, opened 2026-08-04
fix(obs): emit the detailed request metrics on every endpoint, not just chat and messages

jarvis9443:
Fixes api7/AISIX-Cloud#1234

## Problem

`aisix_llm_requests_total` / `aisix_proxy_requests_total` / `aisix_proxy_failed_requests_total` and their duration histograms were emitted by the chat and messages handlers **only**. Every other endpoint recorded just the legacy `aisix_requests_total`.

So `/v1/responses` traffic (Codex and friends) never appeared in any request-count or success-rate query built on the detailed families — while still showing up in the legacy series, which made the gap read as a bad query rather than missing instrumentation.

The reported endpoint was `/v1/responses`, but it is the whole handler family: `/v1/completions`, `/v1/embeddings`, `/v1/rerank`, `/v1/messages/count_tokens`, the three `/v1/audio/*` routes, `/v1/images/generations` and `/v1/videos*` had the identical gap, as did the non-inference surfaces (`/mcp`, `/a2a`, `/v1/realtime`, passthrough, files/batches/fine-tuning) and the pre-dispatch rejections.

`Metrics::record_proxy_request` — the proxy-only variant, no LLM series — had no call sites at all, which is the tell that the two-tier design was intended from the start and simply never wired past chat and messages.

## Implementation

Every handler now emits through one chokepoint, `request_metrics::record`, which writes the legacy series and the detailed families from a single call. Calling `Metrics::record_request` directly is what produced a request present in one family and absent from the others, so nothing does that any more (outside the admin API, which is not proxy traffic).

The LLM-vs-proxy split is a property of the **route** (`LLM_ENDPOINTS`), not of the call site. That matters for correctness, not just tidiness: a 413 refused before dispatch now lands in the same denominator as the model-not-found the handler itself records, instead of the two disagreeing about whether the endpoint had a failure.

Non-inference surfaces get the proxy families only — counting an MCP tool call, a batch-file upload or a 413 as an LLM request would corrupt every per-request token and cost average. `/v1/realtime` stays out for the same reason: it does reach a model but feeds none of the `aisix_llm_*_tokens_total` families, so counting it would inflate the denominator without contributing tokens.

chat and messages move onto the same helper so the family cannot drift apart again. Their emitted labels are unchanged.

## Fixed alongside

All in the code paths this change touches:

- `normalize_endpoint_label` was missing `/v1/videos*`, so all video traffic reported `endpoint="other"` on the in-flight gauge.
- The passthrough **error** path used the caller-supplied `:provider` path segment verbatim as the `provider` label, so `/passthrough/<random>/x` minted one series per random value — #451's unbounded-cardinality hole on the provider axis. It now collapses to `unresolved` unless a configured model uses that provider. (The `AisixPath` rejection path passes the raw `parts.uri.path()`, so normalizing the endpoint label in `reject` is load-bearing too, not just defensive.)
- `/v1/embeddings` and `/v1/images/generations` hardcoded `status = 200` on a success arm that also carries the 501 NotImplemented response — mislabelling it in the access log and booking it as `outcome="success"`. Same fix #426 already made for completions/responses/rerank.
- `/v1/responses` and the rest of the family now report a real `upstream_model` label instead of leaving it unresolved, matching chat and messages.

## Behavior change

New series appear for the previously-missing endpoints. Existing chat/messages series are byte-identical. Dashboards that group `aisix_proxy_requests_total` by `endpoint` will now see the tool/management/tunnel surfaces too.

## Baseline

LiteLLM increments `litellm_proxy_total_requests_metric` / `litellm_proxy_failed_requests_metric` from a single central callback (`async_log_success_event` + `async_post_call_failure_hook`) carrying a `route` label, so all routes are covered uniformly by construction. This change converges on that shape; no divergence.

## Tests

`tests/e2e/src/cases/request-metrics-endpoint-coverage-e2e.test.ts` — verified failing before the fix and passing after:

- `/v1/responses` reaches `aisix_llm_requests_total` (empty before), with the detailed labels actually populated (`upstream_model="gpt-4o-mini"`, not `unknown`).
- a failed `/v1/responses` lands in the same denominator, with `model="unresolved"` rather than the caller's text.
- passthrough is counted as a proxy request but never an LLM one, and the bogus provider name does not reach a label.

Rust unit tests in `request_metrics` pin the classification: no route falls through to `"other"`, no `LLM_ENDPOINTS` entry is unreachable (a typo there fails silently — the endpoint just stops appearing), and the tier split holds for both sides.

Full DP E2E suite passes (483 tests).

### DIFF
--- a/crates/aisix-proxy/AGENTS.md
+++ b/crates/aisix-proxy/AGENTS.md
@@ -30,11 +30,12 @@ opaque binary passthrough (audio, images) corrupts it.
 
 ## Every terminal path emits the access log — including the ones that give up early
 
-The access log and `record_request` are emitted **by the handler**, at the end of
-dispatch, because that is the only place that knows the provider, model and token
-counts. A path that returns before reaching that tail therefore logs nothing, and
-nothing errors: the caller gets a correct status while the gateway keeps no record
-of the request, which is indistinguishable from the request never arriving.
+The access log and `request_metrics::record` are emitted **by the handler**, at
+the end of dispatch, because that is the only place that knows the provider, model
+and token counts. A path that returns before reaching that tail therefore logs
+nothing, and nothing errors: the caller gets a correct status while the gateway
+keeps no record of the request, which is indistinguishable from the request never
+arriving.
 
 Two shapes give up early, and both must answer through
 `reject::reject_before_dispatch` (it renders the envelope *and* emits the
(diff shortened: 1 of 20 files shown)


jarvis9443:
@coderabbitai review

PR #893, opened 2026-08-05
feat(proxy): mid-stream fallback for /v1/messages — both dispatch legs

jarvis9443:
Extends mid-stream fallback (`routing.stream_failure`, AISIX-Cloud#1222) from `/v1/chat/completions` to `/v1/messages`, on both of its dispatch legs. Phase 2 of the feature: the endpoint most Anthropic-SDK/Claude-Code traffic actually uses can now recover a committed (already-200) stream on the remaining fallback targets instead of truncating.

## What changes

**Cross-provider leg** (non-Anthropic targets, `ChatChunkStream` + `AnthropicSseEncoder`): the existing `stream_failover::wrap` combinator is inserted between the first-chunk peek and the SSE pump. Because the encoder lives outside the combinator, its message envelope and content-block state survive the switch — the spliced fallback chunks continue the same wire message with no client-visible seam.

**Anthropic passthrough leg** (byte-verbatim forwarding): a new byte-level combinator (`wrap_anthropic_passthrough`). Armed, it forwards complete SSE frames verbatim while side-channel tracking the wire state (`message.id`, open text block, next block index, delivered partial). On a qualifying failure it dispatches the remaining targets through the chat bridges and re-encodes their chunks onto the client's committed envelope via a resumed `AnthropicSseEncoder` (`AnthropicSseEncoder::resume`) — block indices continue, no second `message_start`. The trailing assistant partial becomes native prefill on Anthropic-wire fallback targets (`split_system` already maps it). Unarmed streams keep today's byte path bit-for-bit.

**Failure classes on the passthrough leg** (previously undetectable or silent):
- in-band `error` frames are withheld (never partially forwarded — forwarding is frame-granular when armed) and either recovered or released verbatim;
- EOF without `message_stop` classifies as `transport_error`;
- the armed combinator owns the per-chunk read timeout, so a stall classifies as `read_timeout` (the silent-truncate byte wrapper stays on unarmed streams).

**Safety gates** (same set as chat): tool_use / thinking blocks on the wire, a `message_delta` already delivered (stop reason on the client's wire), partial text past the 1 MiB accumulation cap, or a structured-output request — all disarm continuation; the failure then terminates exactly as before.

**Telemetry parity with chat**:
- failed serving attempts emit per-attempt UsageEvents with estimated partial spend, stamped `inbound_protocol: anthropic` / sink label `messages` (`emit_usage_event` grew a stamped inner variant for this);
- terminal events now set `stream_outcome` (`success` / `partial_failed` / `partial_recovered`) on this endpoint, are attributed to the *serving* target (id, provider, attempt kind `mid_stream_fallback`, attempt-scoped latency), and fallback targets' TPM is billed post-stream;
- `aisix_mid_stream_fallbacks_total` records recoveries for this endpoint too;
- mid-stream upstream errors on the SSE pump now report status 200 + `stream_outcome: partial_failed` instead of being conflated with client aborts (499) — the 499 label stays reserved for actual client abandons.

**Continuation prompt**: switched to LiteLLM's current instruction verbatim ("The previous assistant response was interrupted mid-stream. Continue exactly where it stopped — do not repeat any of its content. Your response must read as a seamless continuation."). Upstream recently removed its old chat-completions continuation prompt; the Responses-API continuation is the one surviving baseline, so all our endpoints now share its text. Sent as a system turn; on Anthropic-wire targets the mid-conversation instruction becomes a user turn and the partial stays native prefill.

## LiteLLM baseline

LiteLLM has **no** mid-stream fallback for its `/v1/messages` surface (`anthropic_messages` routes through `_ageneric_api_call_with_fallbacks`, which never wraps the stream) — a stream failure there emits one error frame and dies. So there is no behavior to align with on this endpoint; the design transplants our chat-completions semantics (which the #1222 review already settled) onto the Anthropic wire. Where LiteLLM's *Responses* continuation does something comparable, we match its mechanics (continuation input shape, usage merging into the terminal frame).

## Tests

Rust integration (wiremock, `crates/aisix-proxy/src/lib.rs`):
- passthrough leg: in-band failure resumes the client envelope (single `message_start`, single `content_block_start`, no error frame, native-prefill continuation body asserted);
- cross-provider leg: same-stream recovery + OpenAI-wire continuation body;
- default (no `stream_failure`): upstream error frame forwards verbatim, fallback never contacted;
- thinking-block safety gate: withheld error frame released byte-for-byte, no dispatch;
- telemetry: failed-attempt event (estimated partial, `anthropic` stamp) + terminal `partial_recovered` attributed to the fallback target.

TS e2e (`tests/e2e`, real `aisix` binary + real sockets):
- passthrough mid-body connection drop → same-envelope recovery;
- passthrough inter-frame stall → `read_timeout`-classified recovery (armed-combinator-owned timeout);
- cross-protocol: Anthropic-wire head resumed by an OpenAI-protocol fallback target;
- client cancel mid-stream never dispatches a fallback.
The mock upstream harness gained `rawStreamFrames` (verbatim SSE frames incl. `event:` lines) to express Anthropic-wire streams.

`/v1/responses` follows in a separate PR on top of this one.

Ref api7/AISIX-Cloud#1222

## Summary by CodeRabbit

* **New Features**
  * Added mid-stream failover for Anthropic Messages streaming, including passthrough and cross-provider recovery.
  * Preserved streamed content, usage accounting, rate-limit billing, and continuation state during eligible recovery attempts.
  * Added safeguards for unsafe output, malformed responses, structured output, client cancellation, and exhausted fallbacks.
  * Improved streaming telemetry with endpoint, protocol, serving-target, recovery, and terminal-failure details.

* **Bug Fixes**
  * Prevented repeated continuation instructions and ensured accurate usage reporting after failed attempts.

* **Tests**
  * Added end-to-end coverage for connection drops, timeouts, cross-provider recovery, and cancellation behavior.

### DIFF
--- a/crates/aisix-provider-anthropic/src/lib.rs
+++ b/crates/aisix-provider-anthropic/src/lib.rs
@@ -29,8 +29,8 @@ pub use bridge::{AnthropicBridge, ANTHROPIC_DEFAULT_BASE, ANTHROPIC_VERSION};
 /// - [`AnthropicSseEncoder`] re-encodes the bridge's `ChatChunk`
 ///   stream as Anthropic typed SSE events.
 pub use wire::{
-    chat_response_into_anthropic_json, parse_inbound_request,
+    chat_response_into_anthropic_json, parse_inbound_request, stream_error_into_bridge_error,
     translate_anthropic_tool_choice_to_openai, translate_anthropic_tools_to_openai,
     translate_extras_to_openai_shape, AnthropicInboundError, AnthropicSseEncoder,
-    AnthropicSseEvent,
+    AnthropicSseEvent, AnthropicStreamErrorBody,
 };
--- a/crates/aisix-provider-anthropic/src/wire.rs
+++ b/crates/aisix-provider-anthropic/src/wire.rs
@@ -1744,6 +1744,13 @@ pub struct AnthropicSseEncoder {
     seen_input_tokens: u32,
     seen_output_tokens: u32,
     usage_seen: bool,
+    /// Usage folded ADDITIVELY into the closing pair on top of the
+    /// `seen_*` counters (mid-stream failover, AISIX-Cloud#1222): the
+    /// estimated spend of failed partial attempts. Anthropic wire usage
+    /// is cumulative per message, so the terminal counts must also
+    /// cover the partial text the client received before the switch.
+    extra_input_tokens: u32,
+    extra_output_tokens: u32,
 }
 
 impl AnthropicSseEncoder {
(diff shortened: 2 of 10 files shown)


jarvis9443:
@coderabbitai review

jarvis9443:
关闭：mid-stream 续写需求整体暂停待重新评估（见 api7/AISIX-Cloud#1222），一期实现也已 revert。分支 feat/mid-stream-fallback-messages 保留，需求若重启可直接 reopen 或以此为基础接续。

PR #915, opened 2026-08-10
feat(admin)!: remove the Admin API resource write path

moonming:
Removes the Admin API resource write path. The admin listener (`:3001`) keeps its read surface — lists/gets for all 8 resource kinds (including the former `apikeys` spelling), `/admin/v1/models/status`, `/admin/v1/health`, OpenAPI + Scalar UI, playground, `/livez`, `/readyz` — but resources are now managed exclusively through the declarative paths: a `resources_file` (`resources.yaml`, reloaded on SIGHUP) or direct etcd writes.

This is the final step of the deprecation announced in v0.4.0 (RFC 9745 `Deprecation` headers) and executes the removal scheduled for v0.5.0+ after #848 lifted the write-path-exclusive validations into the canonical schemas so the declarative paths enforce them.

## ⚠️ Breaking changes

| Before | After |
| --- | --- |
| `POST /admin/v1/<kind>`, `PUT/DELETE /admin/v1/<kind>/{id}` (deprecated, functional) | **405** with `Allow: GET` |
| `POST /admin/v1/api_keys/{id}/rotate` (and `apikeys` spelling) | **404** — the route is gone |
| File mode: writes rejected with `409` naming the resources file | **405**, same as every other mode |
| Rotate returned a fresh plaintext key | No DP-side plaintext rotation. Rotate declaratively: write the same resource id with a new `key_hash` — the old plaintext stops authenticating as soon as the write propagates (pinned in `apikey-lifecycle-e2e`) |

What to update:
- Scripts that created/updated/deleted resources via `:3001` → write `resources.yaml` (validate offline with `aisix validate --resources <file>`, reload with SIGHUP) or write entity-value JSON to etcd at `{prefix}/{kind}/{id}`.
- Scripts that rotated caller keys via `/rotate` → hash the new secret client-side and update the resource's `key_hash`.
- The published OpenAPI (`/admin/openapi.json`) no longer documents write operations; the write-only component schemas (`ApiKeyRequest`, `ApiKeyEntry`, `ApiKeyRotateResponse`, `DeleteResponse`) are gone from `components.schemas`.

## What changed

**Router / handlers** — every `/admin/v1/*` resource route serves `get(...)` only; both rotate routes deleted; the file-managed write guard and the RFC 9745 deprecation-header middleware deleted (nothing left to mark). 43 write/rotate/uniqueness handler fns removed across the 8 handler modules.

**Store layer** — `ConfigStore` is now a read-only trait (16 `put_*`/`delete_*` methods removed, `StoreError::ReadOnly` gone). `EtcdConfigStore` keeps only reads (module doc rewritten: resources reach etcd through the declarative paths). `FileManagedStore::new(snapshot)` drops the path parameter — read-only by construction. `InMemoryStore` keeps `#[cfg(test)]` inherent write methods used by unit-test seeding.

**OpenAPI** — 24 write operations and the rotate path removed from the base document; the no-op write-deprecation marker pass deleted; unreachable component schemas pruned via a reachability walk from `paths`. New gate test pins that the published reference documents zero non-GET operations under `/admin/v1/` and zero `deprecated` marks anywhere.

**Tests** — write-path unit tests (CRUD flows, rotate atomicity, write-auth, write-validation) deleted; read tests re-seeded through `InMemoryStore`; new contract tests pin 405 + `Allow: GET` on every collection/`:id` route (auth'd and unauthenticated) and 404 on both rotate spellings. The etcd integration test now seeds via direct `etcd_client` puts — the path operators actually use.

**e2e** — `AdminClient` write helpers removed (reads stay; `SeedClient` is the write front door). `file-resource-source-e2e` pins the new file-mode contract (reads serve the file, writes 405, rotate 404, unauthenticated write 405 with no file-path leak). `apikey-lifecycle-e2e`'s rotate coverage became a declarative secret-swap test (old plaintext dies immediately, id unchanged). Deleted: `seed-vs-admin-characterization-e2e` (its own comment scheduled retirement once the seed migration completed), `apikey-budget-e2e` (tested write-path 400s), the sdk-compat deprecation-header test, and forward-compat's admin strict-write test.

**Docs** — README, `config.example.yaml`, crate module docs updated to the read-only story.

## Non-goals / follow-ups (filed internally)

- 3 OpenAPI-parsing validations (mcp `validate_spec`, HeaderName typing, duplicate tool names) still lack declarative-path enforcement — blocked on a dependency-direction extraction, tracked internally.
- etcd watch keeps last-good on an invalid PUT while resync drops the row — policy decision tracked internally.

## Verification

- `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` — all green.
- Real binary, file mode: boots with `resources.yaml`, GET list serves the file, `POST /admin/v1/models` → 405 + `Allow: GET`, rotate → 404, `/admin/openapi.json` documents no admin writes.
- Local e2e stack run before push.

## Summary by CodeRabbit

* **Changes**
  * The Admin API is now read-only for managed resources.
  * Resource listing and detail views remain available through GET requests.
  * Creation, updates, deletion, and API-key rotation through the Admin API are no longer supported.
  * Unsupported write requests return `405` responses, while rotation routes return `404`.
  * Manage resources through `resources_file` reloads or direct etcd writes.
  * Updated configuration, validation guidance, and end-to-end coverage to reflect the read-only administration model.


moonming:
Review triage — every inline comment dispositioned; fixes landed in f9c10cc and e34694b.

**Fixed**
- 7 handler module docs rewritten to the read-only contract; stale rotate/CRUD section comments removed (e34694b)
- OpenAPI `Caller API Keys` tag no longer says "and key rotation"; path/`Entry` descriptions no longer claim ids are generated by the Admin API or that update/rotate bumps revisions (f9c10cc, e34694b)
- `main.rs`: unused `Some(path)` binding → `Some(_)`; canonical product name; removed a comment referencing write rejections (e34694b)
- etcd integration: refused-writes test now covers PUT and DELETE with `Allow` assertions, and the rotate-absence checks GET the rotate URIs too — POST-only 404 couldn't distinguish a deleted route from the old handler's unknown-id 404 (f9c10cc, e34694b)
- sdk-compat e2e: readiness gate switched from the SDK chat path to an independent authenticated `GET /v1/models` probe per the harness gate rules (e34694b)
- `config.example.yaml`: dead `docs/api-admin.md` link and Admin-API management claims replaced with the declarative sources (f9c10cc)

**Not adopted, with reasons**
- *Assert `Allow` with `toBe("GET")`*: the real binary answers `Allow: GET,HEAD` (axum's MethodRouter advertises HEAD alongside GET), so an exact `"GET"` match fails against actual behavior. `toContain("GET")` plus the 405 status is the contract; the 405 itself already proves no write method is routed.
- *Duplicate `key_hash` uniqueness on the direct-etcd path* (`assert_unique_key` removal): real gap, but pre-existing — the admin-side check never guarded direct etcd writes, which existed before this PR and are used by the control plane (which enforces uniqueness org-side) and the file source (which has its own per-kind identity check). Filed internally for loader-side conflict detection with `/status/config` visibility; tracked as a follow-up rather than blocking the removal.

Issue #847, opened 2026-07-30
Establish the on-CPU profiling workflow: adopt cargo-flamegraph and add a dedicated profiling build profile


membphis:
## Background

When the gateway shows high CPU but lower-than-expected throughput, we need to see exactly where the cycles go. The standard diagnostic for this is an on-CPU flame graph produced by a sampling profiler — the same flame-graph-first workflow that is battle-tested in the Nginx/OpenResty world.

We surveyed the Rust ecosystem (The Rust Performance Book recommendations, plus the profiling setups documented by TiKV, Databend, GreptimeDB, RisingWave, Quickwit, DataFusion, and the rustc team). The landscape is well converged, and we are adopting **cargo-flamegraph** as the standard tool for this scenario. This issue pins that decision, sketches the intended usage, and tracks the repo changes needed to make flame graphs actually readable for aisix builds.

## Scenario → tool mapping

| Scenario | Tool | Status |
|---|---|---|
| **On-CPU hotspots** — CPU is high, QPS is low; "what is burning CPU?" | **cargo-flamegraph** (wraps `perf` on Linux / `xctrace` on macOS; renders an interactive SVG via inferno) | **this issue** |
| Interactive drill-down — per-thread timelines, call trees, inverted stacks | samply (Firefox Profiler UI); benefits from the same build profile | optional companion; no repo change needed |
| On-demand profiling of a live production process over HTTP | pprof-rs embedded endpoint (the `/debug/pprof/profile` pattern used across the Rust infra ecosystem) | out of scope — to be filed separately |
| Async stalls — tasks parked on `.await`, scheduling problems; CPU is **not** hot | tokio-console / off-CPU profiling | out of scope |

Boundary to keep in mind: an on-CPU flame graph answers "what is burning CPU". Tasks parked on `.await` consume no CPU and are invisible to it; that class of problem needs the async tooling instead.

## Usage sketch

A dedicated internal guide with the full walkthrough will follow; this is the shape of it:

```bash
cargo install flamegraph

# Linux: perf needs privileges — either run with sudo,
# or lower kernel.perf_event_paranoid for the session:
sudo sysctl kernel.perf_event_paranoid=1

# Build, run the gateway, sample until exit, render the SVG — release is
# cargo flamegraph's default profile, and release binaries are profileable
# by design (the bin target is named `aisix`; the package is `aisix-server`):
cargo flamegraph --bin aisix -- <args>

# Or attach to any already-running aisix process — including a user's
[...]

PR #917, opened 2026-08-10
bench: onthebench-style rig harness for same-rig baselines

membphis:
## Why

Item 1 of the post-thread-per-core performance program (api7/AISIX-Cloud#1259) requires reproducing the public onthebench benchmark methodology on our own m7g.4xlarge, so that every optimization item in the program can be gated on same-rig before/after numbers instead of third-party published figures.

## What

A reproducible one-command harness under `bench/onthebench/`:

- `bench.sh` — rsyncs the working tree to the rig, provisions it, builds a native release binary, runs every load point, and collects results. The tree you run it from is the tree that gets measured; commit and dirty state are recorded in the run metadata.
- `rig-setup.sh` — idempotent provisioning (build deps, perf, rust toolchain via the repo's `rust-toolchain.toml`, inferno). The load generator and mock upstream are the prebuilt, pinned instruments from the public benchmark rig release (engine pin `f3adbb1315b26129f5e317af5279decefb1cea8f`, https://github.com/GetBusbar/benchmarking); their sha256s are frozen in the script, so a rebuilt rig either runs the byte-identical instrument or fails loudly.
- `run-baseline.sh` — the on-rig runner: three disjoint pinned core groups (gateway 0-3, load 4-9, mock 10-15), default shipped config, 25s windows, 5s warmup, 4 valid repetitions per point with `fail=0` as the validity gate (invalid windows are recorded and retried, never silently dropped), rig-floor reference windows against the mock directly, per-window gateway CPU% and peak-RSS capture, and one on-CPU flamegraph at the c=128 saturation point (perf + inferno, the #847 workflow, with a stripped-binary refusal gate).

Results land as one directory per run: `results.jsonl` (one JSON object per window), `meta.json` (commit, binary and instrument sha256s, core split, kernel, method parameters), the flamegraph SVG, generated configs, and logs.

## Default-config discipline

A config line exists only if the process cannot run the benchmark without it, following the published methodology's rule; the full claim set and per-line justification are in the README and the runner. Runtime deviations are limited to `ulimit -n 65536` (c=768 exceeds the 1024 default) and `kernel.perf_event_paranoid=1` (flamegraph sampling only, reverts on reboot).

Two hardening details worth review attention:

- The harness passes source metadata to the rig as `BENCH_SRC_*` variables and defensively unsets every `AISIX_*` variable before launching the gateway: aisix reads `AISIX_*` environment variables as config overrides, so harness metadata in the measured process's environment would either pollute the measurement or refuse boot.
- Flamegraph rendering failures downgrade to a warning and keep `perf.data`; a rendering problem must never abort or invalidate the measurement itself.

## Validation

One full pipeline run end-to-end on the rig (provision from a fresh box, native build, all four load points, both rig floors, flamegraph rendered): every measured window passed the `fail=0` validity gate on the first attempt, and the thread-per-core startup self-check (`tpc_workers`) confirmed the intended serving mode. Baseline numbers are tracked in the program's internal notes rather than in this PR.

## Not in scope

Multi-gateway comparison orchestration (program item 1's side-by-side legs) builds on this harness in a follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added an automated benchmark harness for ARM64 systems.
  * Added one-command rig provisioning, project building, benchmark execution, and result collection.
  * Added baseline load testing across concurrency and time-to-first-byte scenarios.
  * Added latency, throughput, failure, CPU, and memory measurements in JSONL output.
  * Added validation, retries, metadata capture, and flamegraph generation.

* **Documentation**
  * Added comprehensive guidance covering setup, methodology, metrics, outputs, and troubleshooting.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,6 @@ coverage/
 
 # Filled-in real-cloud objstore credentials (template: real-cloud.env.example)
 /crates/aisix-obs/tests/real-cloud.env
+
+# Local benchmark results collected by bench/onthebench/bench.sh
+/bench-results/
--- a/bench/onthebench/README.md
+++ b/bench/onthebench/README.md
@@ -1,0 +1,74 @@
+# onthebench-style rig harness
+
+Reproduces the public onthebench AI-gateway benchmark methodology on our own
+`m7g.4xlarge` (Graviton3, 16 cores, single NUMA node), so performance work can
+be measured on the same rig shape the public board uses — and so any published
+figure can be checked against a box we control.
+
+## One command
+
+```sh
+bench/onthebench/bench.sh                       # default rig, results under ./bench-results/
+bench/onthebench/bench.sh ubuntu@<rig-ip> /path/to/results
+```
+
+The tree you run it from is the tree that gets measured. The script rsyncs the
+source to the rig, provisions it (idempotently), builds a native release
+binary, runs every load point, and rsyncs the results back. The rig is
+disposable; a rebuilt box only needs passwordless ssh + sudo and this harness.
+
+## Methodology
+
+Replicates the published onthebench setup (https://onthebench.ai/gateways/performance):
+
+- **Core split** — three disjoint pinned groups via `taskset`: gateway `0-3`
+  (4 cores), load generator `4-9` (6), mock upstream `10-15` (6). Neither
+  instrument can starve the gateway or bottleneck it.
+- **Instruments** — the prebuilt, pinned `otb` load generator and `mock`
+  upstream from the public benchmark rig release (engine pin
+  `f3adbb1315b26129f5e317af5279decefb1cea8f`, tag `engine-v1`, from
+  https://github.com/GetBusbar/benchmarking). `rig-setup.sh` verifies their
+  sha256 so a re-provisioned rig either runs the byte-identical instrument or
+  fails loudly. These are the same binaries behind the public board and the
+  same `otb loadgen` used for the api7/aisix#891 / #902 tables.
+- **Default shipped config** — a setting may appear only if the process cannot
+  run the benchmark without it. The full claim set (in `run-baseline.sh`):
+  `resources_file` (boot: standalone source, else aisix demands etcd),
+  `proxy.addr` (bind: the port the harness drives), `admin.admin_keys` (boot:
+  config refuses to load without one), plus a resources file wiring the mock
+  as the only upstream and minting the single client key (boot: aisix has no
+  anonymous mode). Everything else — thread-per-core, worker count, telemetry —
+  is whatever the defaults do.
+- **Load grid** — fixed concurrency, the api7/aisix#891 grid: c=16/32/128
+  against the 0-delay mock, c=768 against the 10ms-TTFT mock
+  (`MOCK_TTFT_MS=10`). 25s windows, 5s warmup per point, ≥4 repetitions;
+  a window with any failed request (`fail`, `rigrefused`, `budgetexceeded`,
+  `spawnfailed`) is recorded but marked invalid, and the point retries until
+  4 valid windows or 8 attempts.
+- **Rig floor** — the load generator driving the mock directly (c=128 0-delay,
+  c=768 10ms), so every gateway number can be checked against the instrument's
+  own ceiling.
+- **Recorded per window** — rps, fail/ok counts, p50/p99 (µs, from otb),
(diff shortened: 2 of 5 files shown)


membphis:
Independent audit completed (correctness / reliability / security / leakage / breaking-changes / coverage angles). Findings and resolutions, all addressed in the follow-up commit:

**HIGH — committed default ssh target in a public repository.** Fixed: the rig address now comes only from the first argument or `BENCH_RIG`; usage fails loudly without one, and the README states why no default may ever be committed.

**MEDIUM (5) — all fixed:**
1. Provisioning bundled build deps with the kernel-versioned `linux-tools` package in one apt transaction; an unavailable kernel package silently dropped every build dep. Split into two transactions.
2. Metadata was written only at the end of a run, so any between-window abort left unattributable results, and the driver skipped collection on failure. Metadata is now written before the first measured window and rewritten at the end; the driver collects results unconditionally and preserves the run's exit code.
3. Two failure modes the `fail=0` gate cannot see are now asserted before measurement: the mock must demonstrate the requested TTFT, and the gateway must show thread-per-core workers.
4. The local results directory tainted the dirty-tree metadata of subsequent runs and was re-uploaded to the rig. Now gitignored and excluded from the source sync.
5. Raw stats-line interpolation could corrupt `results.jsonl` on quoted error text — precisely the rows that matter most. Lines are sanitized before JSON interpolation, and floor windows carry the same validity flag as gateway windows.

**LOW (8):** seven fixed (curl `-f` on instrument downloads, `--max-time` on readiness probes, self-terminating RSS sampler, explicit loadgen capture fallbacks, README sampling-frequency alignment, full method parameters in metadata, floor validity as above); the eighth — executing third-party bytes (rustup installer, pinned instrument binaries) as a passwordless-sudo user — is accepted and documented as the trust posture in the README: the pins freeze reproducibility, not trustworthiness, and the rig is a disposable single-purpose box holding no secrets.

membphis:
Reviewer findings resolved in d918740. Note both reviews ran against 7ffc097; four findings were already addressed by 906deb4 (the independent-audit commit) — called out below.

**Copilot:**
- Ignored files (`.env`, `.env.*`, `real-cloud.env`) synced to the rig — **fixed**: the source sync now runs through a gitignore filter (`--filter=':- .gitignore'`), so no ignored file ever reaches the rig; tracked and untracked-but-not-ignored files still sync, and the remote build cache stays protected from `--delete`. Verified with a dry-run against a planted `.env`.
- Combined apt transaction dropping build deps — already fixed in 906deb4 (split transactions).
- `tpc_workers` recorded but not enforced — 906deb4 made it fatal at ≥1; **tightened further**: now must equal exactly one worker per pinned gateway core, derived from the affinity mask.
- README 997 Hz vs 499 Hz — already fixed in 906deb4.
- (suppressed) exhausted retries only warned — **fixed**: any incomplete point flips the run to a nonzero exit after metadata is written and results are collected, so an incomplete run cannot pass as a baseline.

**CodeRabbit:**
- RUNID collision within one second — **fixed**: pid suffix added.
- `cargo build --locked` — **fixed**.
- Flamegraph SVG documented as unconditional — **fixed**: README now states the SVG is conditional and `perf.data` is kept on rendering failure.
- Product name in prose — **fixed**; remaining lowercase occurrences are issue references, binary names, and paths.
- `curl | sh` rustup install — **fixed**: version-pinned `rustup-init` (1.28.2, aarch64) downloaded and sha256-verified before execution, matching the fixed-byte-sequence rule the bench instruments already follow.
- Unpinned inferno — **fixed**: pinned to 0.12.8 with the installed version validated on every provisioning pass.
- Host identity — **fixed**: aarch64 asserted, instance type checked via IMDSv2 and refused on mismatch (recorded as `unknown` when IMDS is absent, e.g. a non-EC2 lab box); instance type added to run metadata.
- Floor windows outside the validity policy — **fixed**: floors now use the same record-mark-retry policy (two valid windows per floor point; an invalid window can never stand as the rig reference). Kept at two repetitions rather than four: floors are instrument references, not baseline points, and the four-repetition contract applies to gateway load points — now stated in the README.
- Regression tests for host-identity/floor failures — **skipped**: this is a single-purpose operator-run bash harness whose gates are exercised on every invocation; a bash test framework here is out of proportion to the risk. The gates fail loudly by construction and were fragment-tested locally.

Validation for this commit: `bash -n` on all scripts, rsync filter dry-run against planted secret files, and local fragment tests of the new gates. The provisioning path (pinned rustup/inferno) exercises end-to-end on the next rig run, which the follow-up comparison work performs anyway.

PR #919, opened 2026-08-10
bench(onthebench): shared measurement lib, generic entrant runner, method overrides

membphis:
## What

Follow-up to #917. Three changes to the same-rig harness, no default-behavior change:

1. **`lib.sh`** — the measurement core (mock lifecycle + TTFT assertion, rig floor, measured windows with CPU%/RSS sampling, validity policy, flamegraph, meta fragments) extracted from `run-baseline.sh` into one shared file. Both runners below source it, so any two runs are comparable by construction instead of by code review.
2. **`run-entrant.sh`** — a generic runner that measures any gateway-shaped process ("entrant") through the exact same code path: same instruments, core split, windows, floors and validity policy. The entrant contract (documented in the script header) is a sourced `entrant.sh` providing a start hook that leaves the measured pid in `GW_PID`, plus optional prepare/teardown hooks, identity metadata, and a request-shape override (`REQ_PATH`/`BODY`/`AUTH_HEADER`) for targets whose only ingress speaks a dialect other than OpenAI chat. Entrant definitions live outside the repository on purpose.
3. **`BENCH_*` method overrides** — grid, reps, window, warmup, max tries, floor reps and flamegraph become env-overridable (defaults unchanged, forwarded by `bench.sh`), so a two-window spot-check or an added delay tier is an invocation rather than a script edit. Overrides are recorded in `meta.json`, so a non-default run can never pass silently as the standard grid.

Floors now derive from the grid: one floor per 0-delay tier at its max concurrency (the instrument ceiling barely moves with concurrency there), one floor per concurrency on delayed tiers (the ceiling is ~conc/delay). Under the default grid this reproduces exactly the floors the harness always ran.

Two correctness fixes that fall out of generalizing:

- Liveness checks use `/proc/<pid>` existence instead of `kill -0`, which reports EPERM — reading as death — for a containerized target owned by another user (the RSS sampler would silently record 0).
- `entrant_prepare` failures propagate through the `tee` pipeline (pipefail + explicit message) instead of aborting with no context; the pipeline-subshell semantics are documented in the contract (state hand-off through files, not variables).

## Why

The performance program needs same-rig, same-instrument comparisons of multiple gateway targets, and extra load points (e.g. an added delay tier at higher concurrency), without forking the measurement methodology per target. One chokepoint (`lib.sh`) keeps the family from drifting.

## Validation

- `bash -n` on all five scripts.
- Same-rig spot-check of the refactored `run-baseline.sh` (`BENCH_GRID="0:128"`, byte-identical product binary vs the #917 baseline run) — results recorded in the program tracker, not here.
- `run-entrant.sh` exercised against native and containerized targets in the program's measurement session (entrant definitions are deliberately out-of-repo; the results land in the internal tracker).

## Summary by CodeRabbit

- **New Features**
  - Added benchmarking for gateway-shaped entrant processes, including preparation, validation, resource measurements, and optional flamegraphs.
  - Added configurable benchmark parameters through `BENCH_*` environment variables.
  - Added concurrency floors across delayed load tiers.
  - Benchmark results now identify the measured entrant and include expanded run metadata.

- **Improvements**
  - Standardized measurements, readiness checks, retries, cleanup, and validity tracking across benchmark runs.
  - Configuration overrides are forwarded to remote baseline executions.
  - Improved benchmark documentation covering load grids, validation tiers, configuration, and entrant requirements.

### DIFF
--- a/bench/onthebench/README.md
+++ b/bench/onthebench/README.md
@@ -76,6 +76,11 @@ match the #891 grid exactly): `BENCH_GRID` (`"ttft_ms:conc ..."`, e.g.
 A changed knob is recorded in `meta.json`, so a non-default run can never
 pass silently as the standard grid.
 
+Values are validated at startup and nonsense refuses to run: grid entries
+must be `ttft_ms:conc` with a positive concurrency; window, reps, max tries
+and floor reps must be positive integers (warmup may be `0`);
+`BENCH_FLAMEGRAPH` is `0` or `1`.
+
 ## Measuring another target ("entrant")
 
 `run-entrant.sh <entrant-dir> <out-dir>` measures any gateway-shaped process
--- a/bench/onthebench/README.md
+++ b/bench/onthebench/README.md
@@ -42,17 +42,19 @@ Replicates the published onthebench setup (https://onthebench.ai/gateways/perfor
   as the only upstream and minting the single client key (boot: AISIX has no
   anonymous mode). Everything else — thread-per-core, worker count, telemetry —
   is whatever the defaults do.
-- **Load grid** — fixed concurrency, the api7/aisix#891 grid: c=16/32/128
-  against the 0-delay mock, c=768 against the 10ms-TTFT mock
+- **Load grid** — fixed concurrency, the api7/aisix#891 grid by default:
+  c=16/32/128 against the 0-delay mock, c=768 against the 10ms-TTFT mock
   (`MOCK_TTFT_MS=10`). 25s windows, 5s warmup per point, ≥4 repetitions;
   a window with any failed request (`fail`, `rigrefused`, `budgetexceeded`,
   `spawnfailed`) is recorded but marked invalid, and the point retries until
   4 valid windows or 8 attempts. A run in which any point ends incomplete
   exits nonzero so it can never be mistaken for a baseline.
-- **Rig floor** — the load generator driving the mock directly (c=128 0-delay,
-  c=768 10ms), so every gateway number can be checked against the instrument's
-  own ceiling; two valid windows per floor point, under the same
-  record-mark-retry validity policy as the gateway points.
+- **Rig floor** — the load generator driving the mock directly, so every
+  gateway number can be checked against the instrument's own ceiling; two
+  valid windows per floor point, under the same record-mark-retry validity
+  policy as the gateway points. The 0-delay tier gets one floor at its
+  largest concurrency (the ceiling barely moves with c); each delayed tier
+  gets a floor per concurrency, because there the ceiling is ~conc/delay.
 - **Recorded per window** — rps, fail/ok counts, p50/p99 (µs, from otb),
   gateway CPU% (`/proc/<pid>/stat` utime+stime across the window), gateway
   peak RSS (VmRSS sampled at 5 Hz), the raw otb stats line. Idle RSS and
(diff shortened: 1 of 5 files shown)


membphis:
Independent cold-context audit ran against 3fbcf1b; all findings addressed in 5b32e4c:

- **HIGH-1** (grid_concs returns nonzero when the last grid element is another tier's; default-grid run dies silently at the first floor under set -e + pipefail) — fixed with the `if` form; a shortened default-grid live run on the rig re-validates the exact failing path (results in the internal tracker).
- **MEDIUM-1** (`prepare || die` disables errexit inside the hook; mid-prepare failures passed as success) — hook now runs in an explicit `( set -e; ... )` subshell with `PIPESTATUS[0]` checked.
- **MEDIUM-2** (per-pid CPU%/RSS silently understates multi-process targets) — child-process count is warned about at start and recorded as `target.children` in meta.json; summing over descendants would change the method mid-program, so surfacing is the chosen floor. All targets measured in the current program are single-process (thread-based); a multi-process entrant would show `children > 0` and its CPU/RSS treated as a lower bound.
- **LOW-1..5** — THREADS double-line capture, flamegraph `enabled` now gated on grid content, headers list derived via json.dumps, entrant flamegraph symbol threshold aligned to >100, contract header documents entrant_stop idempotency and hook-time variable availability.

Re-validation performed on the rig: (a) shortened default-grid baseline run (`BENCH_REPS=1 BENCH_FLOOR_REPS=1 BENCH_WINDOW=5 BENCH_WARMUP=1`) exercising the HIGH-1 path; (b) an entrant run whose grid ends in a delayed tier. Numbers live in the internal tracker per program policy.

membphis:
Disposition of the 10 inline bot findings, addressed in 9f756fe (all fixes re-validated with a shortened default-grid run on the rig):

**Fixed**
- `lib.sh` BENCH_* validation (coderabbit lib.sh:21) — grid entries, counts and the flamegraph knob are validated at startup; zero-rep runs can no longer report success while measuring nothing. Value domains documented in the README. No bash test infra exists for `bench/` in this repo; the regression check is the recorded live-rig validation run, per the harness's existing practice.
- Target death mid-window (coderabbit lib.sh:182) — CPU tick reads are fallible; `cpu_pct` records `null`, the window records `valid=false`, and the JSONL line lands instead of the runner dying on a missing `/proc/<pid>/stat`.
- Floor validity parity (coderabbit lib.sh:216) — floors now reject `rigrefused`/`budgetexceeded`/`spawnfailed` like gateway windows, and an empty loadgen line records nulls, not invalid JSON.
- Header derivation (copilot lib.sh:38) — `OTB_LOADGEN_HEADERS` is now always derived from `AUTH_HEADER` (json.dumps on the first-colon split); an ambient value can no longer diverge from what the readiness check exercised.
- `ENTRANT_NAME` JSON safety (copilot + coderabbit run-entrant.sh:55) — constrained to `[A-Za-z0-9._-]+` at startup.
- `entrant_meta_json` validation (copilot run-entrant.sh:120) — the hook is evaluated before the heredoc; hook failure or unparseable output fails the run loudly instead of shipping a null/invalid identity.
- Statement-position `[ cond ] && cmd` (coderabbit bench.sh:53) — converted to `if` form in both flagged sites as class hygiene: the claim that these sites abort the run is not correct (`&&`-list context is errexit-exempt, and live runs with all overrides unset / `BENCH_FLAMEGRAPH=0` completed exit 0 before this change), but the same pattern in *function-return* position is exactly how the earlier `grid_concs` regression happened, so the class is now gone entirely.

**No change, with reasons**
- `entrant_prepare` in a pipeline (coderabbit run-entrant.sh:60) — already fixed in 5b32e4c: the hook runs inside an explicit `( set -e; … )` subshell, which re-arms errexit regardless of pipeline context, and `PIPESTATUS[0]` is checked. Reproduction of the exact current form: a hook with a failing first step and succeeding last step yields rc=1 and does not continue past the failure. The quoted repro tests the pre-5b32e4c form.
- `sudo -n nm` for `/proc/<pid>/exe` (coderabbit run-entrant.sh:126) — rejected: with `kernel.perf_event_paranoid=1`, `perf record -p` cannot attach to another user's process anyway, so escalating the symbol probe would only let the run proceed into a guaranteed perf failure. The unreadable-exe skip is the correct gate; flamegraph-bearing entrants run as the rig user by contract.

PR #926, opened 2026-08-11
test: freeze the unit-test rate-limit clock; widen the audio-timeout e2e margin

jarvis9443:
Two CI de-flakes, both timing races a loaded runner can lose.

**Frozen rate-limit clock in proxy unit tests.** Every "the next request 429s" assertion implicitly requires both requests to land in the same fixed-window minute. On the wall clock, a test that straddles an :00 boundary puts them in different windows and the 429 never comes — reproduced on the `mcp` per-server-limit tests under a loaded machine (the two requests crossed the boundary), and any of the ~15 OK-then-429 tests in the crate can lose the same race. `ProxyState::new` now pins the limiter to `TestClock` under `#[cfg(test)]`, so every request a unit test makes lands in one window by construction. Production callers (including aisix-admin's playground, a separate crate) compile the system-clock arm; the deterministic-clock hook (`with_limiter`, previously unused) is what this wires up.

**Audio-timeout elapsed bound.** `audio-timeout-e2e` asserted `elapsed < SLOW_MS - 800` (2200ms) while the model timeout fires at 400ms — a CI run has been observed at 2202ms, failing by 2ms of runner overhead. The bound is now `SLOW_MS - 200`: still strictly below the upstream's 3000ms stall (the pre-fix behavior waits the full stall, so the regression still fails hard), with 600ms more headroom for a slow runner.

No behavior change outside test builds. Existing suite passes unchanged with the frozen clock (855/855).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved audio timeout handling in end-to-end testing to better accommodate CI timing variations while preserving the expected timeout behavior.
  * Made rate-limiter timing deterministic during automated tests, improving consistency and reliability of test results.

### DIFF
--- a/crates/aisix-proxy/src/state.rs
+++ b/crates/aisix-proxy/src/state.rs
@@ -258,10 +258,25 @@ impl ProxyState {
         let metrics = Arc::new(Metrics::new(false));
         let guardrail_index =
             LiveGuardrailIndex::new_with_sink(snapshot.clone(), None, Some(metrics.clone()));
+        // Unit tests get a frozen rate-limit clock: on the wall clock, any
+        // "the next request 429s" assertion silently races the fixed-window
+        // minute boundary — a test that straddles :00 lands its two requests
+        // in different windows and the 429 never comes (seen flaking in the
+        // mcp per-server-limit tests under a loaded runner). Freezing the
+        // clock puts every request of a test in one window by construction.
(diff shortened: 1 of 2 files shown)


jarvis9443:
Extracted the magic timestamp into `TEST_RATE_LIMIT_CLOCK_SECS` per the CodeRabbit note.

PR #931, opened 2026-08-11
feat(auth): claim_mappings — resolve verified JWT claims to an existing API key

jarvis9443:
Ref api7/AISIX-Cloud#564 (phase 1 of the claim-mapping effort; the CP schema/projection/dashboard half follows as its own PR — until it lands this is configurable only via the declarative resources file).

## Problem

JWT authentication today resolves an identity through exactly one path: the token's identity claim must equal some key's `jwt_subject` under the matching `jwt_provider`. Every external identity therefore needs a pre-provisioned key row. An enterprise IdP asserting departments/groups for hundreds of developers cannot be onboarded without minting a key per person, and nothing in the gateway can say "everyone in `department: finance` runs under the finance policy".

## Design

A new env-scoped resource kind, `claim_mappings`:

```yaml
claim_mappings:
  - name: finance-dept
    jwt_provider: corp-keycloak     # OIDC provider whose tokens this rule applies to
    priority: 100                   # lower evaluates first; ties break by name
    match:                          # ALL conditions must hold
      - claim: department           # dotted paths traverse nested objects
        op: exact                   # exact (string) | contains (array)
        values: ["finance"]         # any-of
    resolve:
      api_key: finance-policy-key   # file sugar; canonical field is api_key_id
```

Resolution order after full OIDC verification (signature/iss/aud/exp/nbf/scopes/bound claims — unchanged, and always first):

1. The direct `(jwt_provider, jwt_subject)` binding stays authoritative for its subject, including its disabled/expired lifecycle — adding a mapping can never reroute or re-enable a pinned identity.
2. Otherwise the enabled mappings for the matched provider are evaluated in `(priority, name)` order; the first fully-matching rule selects the key the request runs as. Deterministic across replicas and snapshot updates — no set-iteration ordering anywhere.
3. No match, dangling `api_key_id`, missing claim, or a claim whose JSON type doesn't fit the operator → reject (`jwt_identity_unmapped` on the wire; metrics distinguish `claim_mapping_target_missing`). Nothing ever falls through to a default.

Claims only ever **select** a key an operator already created; no claim value becomes configuration (no routes/URLs/budgets from tokens). The resolved key's `allowed_models`, MCP/A2A ACLs, rate limits, team scope, and budget apply exactly as if the caller had presented the key.

## Usage attribution

Many identities can now share one policy key, so `api_key_id` alone no longer names the caller. `UsageEvent` gains `jwt_subject`, `jwt_provider`, and `jwt_claim_mapping` (empty-omitted; cp-api stores empty as NULL, older cp-api images ignore them — standard DP-first rollout). Stamped via one shared `usage_attr::apply_jwt_identity` helper wired through **every** handler family emit path (chat / messages / responses / completions / embeddings / rerank / audio / images / videos / realtime / passthrough / jobs incl. deferred batch attribution / MCP / A2A / shared error emitter), so the family can't drift. OTLP spans carry them as `aisix.jwt_*` attributes; the Datadog/SLS sinks serialise the event with serde and inherit them automatically. No new Prometheus labels (subject cardinality is unbounded by design).

## Declarative file + export

- `claim_mappings` joins the resources-file KINDS with `resolve.api_key` name sugar (desugared to the derived id, exactly like `models[].provider_key`).
- Load-time cross-checks: `jwt_provider` must name a provider defined in the file; a canonical `resolve.api_key_id` must land on a file-defined key; empty `match` is rejected by schema (a rule is always an explicit selection, never match-all).
- The snapshot export emits the kind and resugars `api_key_id` back to the key's name so the round-trip loads.

## Ecosystem comparison

Surveyed three mainstream gateways before building (design notes internal): one resolves claim *fields* to internal team/user/org/virtual-key entities (values must equal entity ids, with optional alias lookup and auto-create), one maps claims to consumer groups that gate tool ACLs, one validates claims with exact/contains/regex matchers but stops short of entity resolution. This design follows the shared principle — claims resolve to pre-existing controlled objects, default-deny — and takes the rule form (operators + priority) because enterprise claim values (AD groups, department names) never equal our resource ids; the per-issue discussion in AISIX-Cloud#564 specifies exactly this shape. Deliberate deltas: no auto-create/upsert of targets (default-deny stance), and deterministic multi-rule ordering instead of unordered set iteration. JWT claim semantics per RFC 7519 §4.1.

## Tests

- Unit: matcher semantics (strict typing, nested paths, conjunctive conditions), priority/tie-break/disabled/provider-scoping of rule selection, model serde/schema round-trips, file-source load + desugar + every cross-check error shape.
- E2E (`claim-mapping-e2e.test.ts`, real binary + etcd + mock IdP + mock upstream + OTLP receiver, 9 cases): matched rule runs as its key (ACL enforced), priority wins with nested `contains`, direct binding beats rules both ways, no-match 401, dangling target 401 + distinct metric reason, disabled target key 401 `api_key_disabled`, and span-attribute attribution for rule-resolved / directly-bound / plain-key requests.
- Existing `jwt-auth-e2e` (25) + `auth-baseline-e2e` (4) suites pass unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added claim mappings that resolve verified JWT claims to API keys.
  * Supports provider-scoped matching, priorities, nested claims, exact matches, and array-contains matches.
  * Added validation, configuration schema support, snapshot handling, and export/import support.
  * JWT identity details are now included in usage telemetry and observability data.
* **Bug Fixes**
  * Direct provider-and-subject API-key bindings continue to take precedence over claim mappings.
  * Invalid, disabled, unmapped, or conflicting configurations now produce clear errors.

### DIFF
--- a/crates/aisix-core/src/bin/dump-schema.rs
+++ b/crates/aisix-core/src/bin/dump-schema.rs
@@ -62,6 +62,7 @@ fn main() {
         "mcp_policy",
         "a2a_agent",
         "oidc_provider",
+        "claim_mapping",
     ] {
         dump_value(
             &out_dir,
--- a/crates/aisix-core/src/filesource/desugar.rs
+++ b/crates/aisix-core/src/filesource/desugar.rs
@@ -143,6 +143,40 @@ pub(crate) fn desugar_model(doc: &mut Value, maps: &IdentityMaps) -> Result<(), 
     Ok(())
 }
 
+/// `claim_mappings[].resolve.api_key` (name) → `resolve.api_key_id`
+/// (derived id).
+pub(crate) fn desugar_claim_mapping(doc: &mut Value, maps: &IdentityMaps) -> Result<(), String> {
+    let Some(resolve) = doc.get_mut("resolve").and_then(Value::as_object_mut) else {
+        return Ok(()); // missing / mistyped resolve: canonical validation reports it
+    };
+    let Some(name_value) = resolve.get("api_key") else {
+        return Ok(());
+    };
+    let Some(name) = name_value.as_str() else {
+        return Err("`resolve.api_key` must be a string (an api key display_name)".into());
+    };
+    if resolve.contains_key("api_key_id") {
+        return Err(
+            "`resolve.api_key` (a name reference) and `resolve.api_key_id` are mutually \
+             exclusive — set exactly one"
+                .into(),
+        );
+    }
+    let resolved = maps
+        .get("api_keys")
+        .and_then(|m| m.get(name))
+        .cloned()
+        .ok_or_else(|| {
+            format!(
+                "`resolve.api_key` references unknown api key {name:?} ({})",
+                known_names(maps, "api_keys")
+            )
+        })?;
+    resolve.remove("api_key");
+    resolve.insert("api_key_id".into(), Value::String(resolved));
+    Ok(())
+}
+
 /// `api_keys[]`: strip the identity-only `display_name`, resolve
 /// `key_env` XOR `key_hash`. The plaintext read from the environment is
 /// hashed and dropped — it must never surface in errors, logs, or the
--- a/crates/aisix-core/src/filesource/mod.rs
+++ b/crates/aisix-core/src/filesource/mod.rs
@@ -45,11 +45,11 @@ use std::path::Path;
 use yaml_rust2::{Yaml, YamlLoader};
 
 use crate::models::{
-    validate_a2a_agent, validate_apikey, validate_cache_policy, validate_guardrail,
-    validate_mcp_server, validate_model, validate_observability_exporter, validate_oidc_provider,
-    validate_provider_key, validate_rate_limit_policy, A2aAgent, ApiKey, CachePolicy, Guardrail,
-    McpServer, Model, ObservabilityExporter, OidcProvider, ProviderKey, RateLimitPolicy,
-    SchemaError,
+    validate_a2a_agent, validate_apikey, validate_cache_policy, validate_claim_mapping,
+    validate_guardrail, validate_mcp_server, validate_model, validate_observability_exporter,
+    validate_oidc_provider, validate_provider_key, validate_rate_limit_policy, A2aAgent, ApiKey,
+    CachePolicy, ClaimMapping, Guardrail, McpServer, Model, ObservabilityExporter, OidcProvider,
+    ProviderKey, RateLimitPolicy, SchemaError,
 };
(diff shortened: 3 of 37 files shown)


jarvis9443:
Independent audit completed (correctness / reliability / security / leakage / breaking-changes / coverage). No HIGH findings. Resolution of the rest, now in 93bf621e:

- **MEDIUM — ambiguous binding fell through to the rules**: two keys claiming one `(jwt_provider, jwt_subject)` previously failed closed; the first commit's rule fallback would instead have evaluated mappings for such a subject. Fixed — ambiguity is now reported by `key_for_subject` and rejected before rule evaluation (`jwt_binding_ambiguous` metrics reason, wire code unchanged), with unit + e2e coverage (the e2e discriminates: pre-fix the duplicate-bound subject resolved through a rule to 200).
- **MEDIUM — "disabled binding beats matching rule" untested**: behavior was correct but unpinned; a plausible wrong refactor would have kept every suite green. New e2e pins 401 `api_key_disabled`.
- **MEDIUM — export resugar untested**: the file-loader round-trip test now carries an api_key + trust provider + claim mapping (resugar → re-resolve to the derived id), plus a dangling `resolve.api_key_id` → blocking-diagnostic case.
- **LOW — propagation probe pinned only the first-written rule** (slow-CI flake window): probe now discriminates on the last-written rule's error code.
- **LOW — supervisor per-kind watch guards not extended**: `claim_mappings` added to the apply_put/apply_delete propagation guard tests (the #826-class trap).
- **LOW — duplicate `(priority, name)` rows could pick different rules per replica**: sort key now ends with the entry id — a total order.
- **LOW — realtime post-auth error events carry no caller attribution**: pre-existing gap (empty `api_key_id` on that path predates this PR); the new jwt fields inherit the same shape. Filed as #932 rather than expanding this PR.
- **LOW — per-request sort cost**: evaluated and accepted — O(N log N) over the environment's rules only on JWT requests with no direct binding, negligible next to the RSA verification on the same path; an indexed snapshot is not warranted at realistic rule counts.

One deliberate note for the CP/docs half: with JWT auth enabled, usage events / OTLP spans / external sinks now carry `jwt_subject` (the identity claim, which operators may configure to an email). That is the feature's purpose (per-identity attribution), but it is a new PII flow toward configured exporters and the docs will call it out.

jarvis9443:
Superseded by #936 — identical content (audit fixes included). GitHub stopped creating Actions check suites for this PR's branch after the first push (synchronize/reopen events never materialized; third-party app suites stuck queued as well), so CI could not run here. The fresh branch/PR restored delivery.

Issue #941, opened 2026-08-12
Collapse repeated per-request snapshot loads and provider-key resolutions in the chat dispatch path


membphis:
## Context

A profiling audit of the zero-config hot path found the request lifecycle loads the config snapshot handle up to 6 times per chat request (auth extractor, dispatch, quota gate, request-metrics emit x2, usage-event emit; +1 on the streaming path) and resolves the same ProviderKey 4-5 times (dispatch-time resolutions plus `provider_key_metric_name` twice in `request_metrics::record`/`record_usage` and the telemetry-tags lookup in `emit_usage_event`).

The dominant costs found by the same audit (full exporter-table scans per request, empty-table shard walks) have been fixed at the `ResourceTable` chokepoint. What remains is the structural part:

- Thread one `Arc<AisixSnapshot>` (loaded once at dispatch entry) through the emit helpers (`emit_usage_event`, `request_metrics::record`/`record_usage`) instead of re-loading per helper. Note this freezes a request's config view to its start, which is arguably more correct but is a small observable change for exporters added mid-request.
- Resolve the attempt's ProviderKey entry once per completion and pass the resolved entry (or its display name + telemetry tags) into the metric and usage emitters instead of re-resolving by id in each.

## Why deferred

Each load is one atomic RMW and each resolution one DashMap read + small clone (~0.5us/request combined, frame-inclusive). Capturing it means changing the signatures of emit helpers shared by the whole endpoint family (10+ handler files, streaming and non-streaming branches) — a wide, cross-family diff that was out of proportion for the remaining size in the zero-config package PR. Doing it later as a focused refactor keeps the family in lockstep in one review.

## Size

~0.4-0.6us/request estimated recoverable; touches aisix-proxy handler family + request_metrics + usage_attr.

membphis:
Landed in api7/aisix#953 (squash-merged as `a04a66c`).

**What shipped.** Both halves of this issue, across the whole client-facing
handler family in one PR:

- One `Arc<AisixSnapshot>` per request, threaded from handler entry through
  dispatch, the quota gate and every terminal emitter. A non-streaming chat
  request loads twice (the auth extractor, then the handler) instead of six
  times; the streaming path drops its request-start load as well.
- One ProviderKey resolution per emitted attempt, via a new
  `usage_attr::ResolvedPk` that serves both the `provider_key_name` metric
  label and the usage event's `telemetry_tags` from a single row read —
  replacing three reads per request.

[...]

Issue #940, opened 2026-08-12
Budget gauge writes are not gated when budgets are disabled


membphis:
## Context

On deployments where budget enforcement is disabled (`BudgetClient::disabled()`, i.e. standalone mode with no cp-api), every request still pays a constant gauge write: `chat.rs` calls `record_budget_gauges(..., None)` -> `clear_budget_gauges` -> `aisix_budget_details_present{api_key_id,team_id,user_id} = 0`. The write costs ~0.2us/request (label-key construction dominates; the cached-handle path still builds the 3-label key every time).

## Why it was not included in the zero-config fast-path package

Gating the write on "budgets disabled" removes the `aisix_budget_details_present` series for keys that have traffic, which violates the /metrics series-compatibility invariant the package is shipped under. The alternatives all have worse trade-offs:

- Value-memoisation only saves the atomic store, not the label-key construction (the dominant cost).
- A periodic resync over `snapshot.apikeys` would inflate the series set from "keys with traffic" to "all configured keys" — unacceptable on deployments with very large key counts.

## What this issue asks for

Decide and implement one of:

1. Accept the series change: when budgets are disabled, do not emit `aisix_budget_details_present` at all (documented behavior change, changelog entry). The gauge is arguably meaningless when no budget subsystem is active.
2. Emit the 0-value series once per key (first request) instead of every request, preserving series presence with a bounded write cost.
3. Keep the status quo and close as won't-fix with rationale.

Option 2 preserves the observable contract exactly (a gauge is level-triggered; re-writing 0 every request is redundant) and is likely the cheapest safe form.

## Size

~0.2us/request on the zero-config hot path (~4% of the package's total recoverable time). Low priority, pure telemetry-cost hygiene.

membphis:
Closing as not planned. The design question does have a fourth candidate, and it costs more structure than the 0.2µs is worth.

## Option 2 cannot skip the work it exists to skip

The per-request cost of `clear_budget_gauges` is: build the lookup key into the worker's reusable buffer (instance prefix + site + the three label values), FNV-hash it, probe the thread-local handle map, then `gauge.set(0.0)`.

To answer "has this label set already been published?" you have to *identify the label set* — which is the key build and the hash, i.e. everything except the final atomic store. Memoising the value saves only that store, which the issue already identifies as the small half.

## Making the memo cheaper means keying it on something other than the label values

Each candidate breaks something:

[...]

PR #944, opened 2026-08-12
perf: zero-config fast paths for the per-request hot path

membphis:
# perf: zero-config fast paths for the per-request hot path

## What

A profiling audit of the default single-model deployment (no rate-limit
policies, no guardrails, no exporters, no metric-based routing) found the
request hot path paying for a series of mechanisms that have **no
configured consumer**: empty-table shard walks, bookkeeping writes nobody
reads, eagerly computed values only error paths use. This PR removes that
spend without changing any observable behavior.

Per cluster:

1. **O(1) table emptiness + zero-policy quota gate** — `ResourceTable`
   keeps a cached `AtomicUsize` entry count (`insert`/`remove` maintain
   it; `Clone` re-seeds it), so `len()`/`is_empty()` stop walking every
   DashMap shard. `reserve_policy_layers` returns early on an empty
   policy table — covering both the request gate and the per-target gate
   — and the reservation list allocates on first use instead of always.
2. **Guardrail empty-index fast path** — an empty index resolves to the
   same empty chain for every request, so
   `LiveGuardrailIndex::resolve` returns early: no resolve walk, no
   applied-set copy, no sink attach. One chokepoint; all 14 endpoint
   call sites inherit it.
3. **On-demand routing bookkeeping** — the in-flight counter's only
   reader is `least_busy` ordering, the latency EWMA's only reader is
   `least_latency` ordering, and the read side already gates on
   `is_metric_based()`; the write side did not. A shared
   `BookkeepingFlags` derives three predicates from the live snapshot
   (least_busy configured / least_latency configured / background health
   checks enabled), recomputed at most once per snapshot version. While
   **any** predicate holds, all four tracker methods run their
   historical write paths byte-for-byte. With none: `begin_in_flight`
   hands out a no-op guard, `record_latency` returns, `mark_healthy`
   takes a read-first path (steady state = one read guard; anything to
   do — cooldown recovery, first-success Healthy publish — still takes
   the full write path), and `record_success` resets the failure counter
   under the read guard without creating entries.
4. **Lazy auth-path side computations** — the source IP is resolved only
   where it is consumed (denial logs on the failure path, one
   `ClientContext` resolve on the success path — previously computed
   twice per request); key-expiry only reads the wall clock for keys
   carrying a deadline; the JWT gate checks the (now O(1))
   enabled-provider predicate before the structural token probe;
   `looks_like_jwt` no longer allocates; `wildcard_matches` decides
   without materialising the capture (a bare `"*"` allowlist entry no
   longer copies the model name per request).
5. **O(1) empty fast paths for table scans** — `entries()` / `any()` /
   `find_unique_by()` return early on an empty table. Every per-request
   full-table scan funnels through these (~25 call sites: the OTLP
   exporter fan-out and content-capture cap run two scans per request,
   plus cache-policy scans and OIDC probes), with zero call-site churn.
   `sanitize_tag` returns clean in-cap input as-is.
6. **One telemetry middleware layer instead of two** — the in-flight
   gauge and the client-cancel recorder sat at the same stack position
   with nothing between them; a single layer arms both guards (drop
   order preserved), removing a boxed service hop, one route-normalize
   call, and two label allocations per request.

## What deliberately did NOT change

- `/metrics` series set: verified name+label identical (209 vs 209
  series) between main and this branch under an identical mixed load
  (non-streaming chat, streaming chat, embeddings, 404, 401 — the
  embeddings leg exercises the single-attempt handler family, which
  calls `mark_healthy` without `begin_in_flight`). `begin_in_flight`
  keeps its entry-creation side effect on the gated path, and
  `mark_healthy` stays a no-op for a missing entry, so the
  `aisix_deployment_state` series appears for exactly the targets and
  at exactly the lifecycle point it does on main — for every endpoint
  family.
- Deployments with `least_busy`, `least_latency`, or background health
  checks configured: any active predicate routes all four tracker
  methods through their historical write paths unchanged.
- Rejection semantics, expiry enforcement, `"*"` matching: byte-level
  equivalent (wildcard fast path is the same decision function without
  the capture; expiry check is the same predicate without the clock read
  for `None`).
- Cooldown early-recovery on request success is preserved for every
  configuration, including zero-config direct models (where a cooled
  model fails fast and in-flight successes are the recovery signal).
- No new user-facing configuration surface: the bookkeeping predicates
  derive from existing routing/health config, so no control-plane
  changes are required.

## Numbers (same-rig A/B, m7g.4xlarge Graviton3, 4 pinned gateway cores)

c=128 saturation, 25s windows x4 per leg, every window fail=0, anchor
drift front->back -0.30%:

| leg | rps | us/req CPU |
|---|---|---|
| main `b84d77b` (anchor pool) | 35,440 | 112.69 |
| **this branch (default state)** | **39,486** | **101.13** |

**+11.4% rps / -11.6 us/req** — 2.5x the +3.5-4.5% frame-share
estimate, consistent with the systematic underestimation frame
accounting has shown on every prior switch-measured mechanism.

Per-leg attribution (single binary, runtime leg switches, methodology as
in #925): structural clusters (lazy auth path, source-IP dedup,
sanitize fast path, middleware merge) -5.0 us; on-demand bookkeeping
-2.6 us; empty-table scan fast paths -6.0 us; the quota gate and
guardrail fast paths land below single-leg resolution (+-1.5 us) and are
covered by the package total.

Combined with the URL-cache PR (both default states), the full grid:
c=16 34,432 / c=32 37,164 / c=128 38,983 / c=768@10ms 33,620 rps — all
fail=0, no shape anomaly.

Post-package LSE census (same-session flamegraphs, leaf-attributed):
total atomic-contention share 7.31 -> 5.39 us/req; the quota
(0.59->0.06), snapshot/DashMap (0.70->0.25), dispatch-misc
(1.82->0.80), guardrails (0.11->0.02) and bookkeeping (0.21->0.10)
buckets collapsed as the audit predicted; the remainder is framework
floor (axum/tower 1.48, bytes/tokio/hyper 1.55, metrics/usage 0.86).

## Deferred (tracked, not silent)

- #940 budget gauge write when budgets are disabled (needs a series-
  contract decision first; ~0.2µs/req).
- #941 collapsing repeated snapshot loads + ProviderKey resolutions in
  the emit helpers (wide cross-handler signature change; ~0.5µs/req).

## Tests

- Unit: workspace green (aisix-proxy incl. 7 new invariant tests for
  the bookkeeping predicates and fast paths — among them one pinning
  that `mark_healthy` on a never-seen id stays a no-op; core
  cached-count mutation coverage plus a differential oracle locking
  `wildcard_matches` to `wildcard_capture` across every pattern shape;
  guardrails empty-fast-path shape test).
- E2E (vitest, real gateway + etcd + mocks): full suite green in BOTH
  runtime modes — thread-per-core and work-stealing (191 files, 546
  tests).
- Metrics series diff harness: 185/185, zero missing, zero extra.

### DIFF
--- a/crates/aisix-core/src/snapshot.rs
+++ b/crates/aisix-core/src/snapshot.rs
@@ -18,17 +18,38 @@
 use crate::resource::{Resource, ResourceEntry};
 use arc_swap::ArcSwap;
 use dashmap::DashMap;
-use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::atomic::{AtomicU64, AtomicUsize, Ordering};
 use std::sync::Arc;
 
 /// Per-kind table with primary id-index and secondary name-index.
 ///
 /// Both indices point at the same `Arc<ResourceEntry<T>>` so there is no
 /// duplicate storage — the name map just holds ids.
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub struct ResourceTable<T: Resource> {
     by_id: DashMap<String, Arc<ResourceEntry<T>>>,
     by_name: DashMap<String, String>,
+    /// Cached entry count, maintained by [`ResourceTable::insert`] /
+    /// [`ResourceTable::remove`]. DashMap's own `len()` / `is_empty()`
+    /// visit every shard (a CAS pair per shard), so per-request
+    /// emptiness checks on the hot path go through this counter
+    /// instead — one relaxed load, O(1) regardless of shard count.
+    count: AtomicUsize,
+}
+
+/// Manual impl: `AtomicUsize` is not `Clone`. The count is re-seeded
+/// from the cloned map's length, which the etcd watch supervisor's
+/// clone-then-mutate update cycle relies on being exact.
+impl<T: Resource> Clone for ResourceTable<T> {
+    fn clone(&self) -> Self {
+        let by_id = self.by_id.clone();
+        let count = AtomicUsize::new(by_id.len());
+        Self {
+            by_id,
+            by_name: self.by_name.clone(),
+            count,
+        }
+    }
 }
 
 impl<T: Resource> Default for ResourceTable<T> {
@@ -36,6 +57,7 @@ impl<T: Resource> Default for ResourceTable<T> {
         Self {
             by_id: DashMap::new(),
             by_name: DashMap::new(),
+            count: AtomicUsize::new(0),
         }
     }
 }
@@ -46,11 +68,11 @@ impl<T: Resource> ResourceTable<T> {
     }
 
     pub fn len(&self) -> usize {
-        self.by_id.len()
+        self.count.load(Ordering::Relaxed)
     }
 
     pub fn is_empty(&self) -> bool {
-        self.by_id.is_empty()
+        self.len() == 0
     }
 
     /// Insert or replace an entry, updating both indices.
@@ -70,12 +92,15 @@ impl<T: Resource> ResourceTable<T> {
         }
 
         self.by_name.insert(name, id.clone());
-        self.by_id.insert(id, Arc::new(entry));
+        if self.by_id.insert(id, Arc::new(entry)).is_none() {
+            self.count.fetch_add(1, Ordering::Relaxed);
+        }
     }
 
     /// Remove by id; also removes the matching name index entry.
     pub fn remove(&self, id: &str) -> Option<Arc<ResourceEntry<T>>> {
         let (_, entry) = self.by_id.remove(id)?;
+        self.count.fetch_sub(1, Ordering::Relaxed);
         let name = entry.value.name().to_string();
         self.by_name.remove_if(&name, |_, v| v == id);
         Some(entry)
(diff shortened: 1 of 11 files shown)


membphis:
## Independent audit + bot review disposition

A cold-start independent audit (correctness / reliability / security / info-leakage / breaking changes / test coverage) ran against this PR, alongside the Copilot and CodeRabbit reviews. Findings and dispositions:

**MEDIUM-1 (audit + Copilot inline, health.rs): deployment-state series growth on single-attempt endpoints — FIXED in d3df98e.** The inactive `mark_healthy` created an entry (and published `aisix_deployment_state`) for a missing id, but embeddings/images/completions/audio/rerank/count_tokens call `mark_healthy` without `begin_in_flight`, so on main their targets never had that series. Fix: `begin_in_flight` keeps its entry-creation side effect on the gated path (read-guard existence check steady-state), `mark_healthy` stays a no-op for a missing entry — byte-for-byte main semantics for every family. New unit test pins the no-`begin_in_flight` case; the series harness now drives /v1/embeddings and holds zero diff (209/209).

**MEDIUM-2 (audit, wildcard.rs): duplicated authz decision procedures without a differential test — FIXED in d3df98e.** Added the oracle test locking `wildcard_matches` to `wildcard_capture(..).is_some()` across all pattern shapes (multi-star, interior star, overlap guard, empty capture, literals).

**Major (CodeRabbit, snapshot.rs) / LOW-3 (audit): count published after the map insert — FIXED in d3df98e.** The count is now incremented before the insert and corrected on replace, so during any mutation window it can only read high: the empty fast paths may take one redundant scan but can never skip an entry already visible in the map. (Production mutates tables clone-then-store before publication, so the window was theoretical — the high-side ordering removes even that.)

**LOW-4 (audit, BookkeepingFlags version lag): no change needed** — verified against `SnapshotHandle::store` ordering; a poisoned write self-heals on the next call, as the code comment states.

**LOW-5 (audit): series harness not in-repo** — the harness lives with the perf-program materials (deliberately untracked); the load gap it had is closed (embeddings leg added) and the invariant is now ALSO pinned in-repo by the new unit test. Test-count wording in the description corrected.

**CodeRabbit nitpicks (maintainability, health.rs)**: the flag-bit constants and packed encoding stay as-is — the packing is load-bearing (single-atomic no-tear read) and documented inline.

Issue #942, opened 2026-08-12
Extend upstream URL caching to the A2A bridge and token-minting vendor bridges


membphis:
Follow-up to the per-ProviderKey endpoint URL cache (aisix_gateway::url_cache).

Two same-shape call sites were deliberately left out of the first PR:

1. **A2A bridge** (`crates/aisix-a2a/src/bridge.rs`): posts `&self.upstream.url` (a String), so reqwest re-parses per call. The upstream URL belongs to an A2A agent resource, not a ProviderKey, and the bridge is constructed per call — caching needs either an agent-id-keyed variant of the cache or a parsed `Url` carried on the agent's resolved upstream struct.

2. **Token-minting vendor bridges** (Vertex-style): the request URL embeds the upstream model and per-request material; nine call sites re-format and re-parse per request. Caching needs the key to include the model dimension.

Both are off the primary hot path (the chat family is covered), so this is small-margin hygiene: est. <0.2us/request on affected endpoints.

membphis:
Scope update from the PR #945 audit round:

1. **Azure bridge added to this issue's scope** (withdrawn from #945): its URL varies per deployment (per Model) while the cache keys per ProviderKey x endpoint — a resource key serving several deployments would fingerprint-miss every request and pay a rebuild + inner-map write per request. Needs a deployment-keyed inner level (e.g. endpoint -> deployment -> Url with `String: Borrow<str>` lookup) before wiring.
2. **Additional unwired same-shape sites** found by the audit, all cold-path: `realtime.rs:233`, `jobs.rs:282`, `videos.rs:645/659/960`. The hot siblings (native /v1/messages passthrough, /v1/responses, /v1/rerank) were wired in #945 directly.

#945 now covers: OpenAI-compatible bridge (5 endpoints), Anthropic bridge (2), and the direct-dispatch proxy paths messages / responses / rerank / count_tokens / audio x3.

PR #945, opened 2026-08-12
perf(gateway): cache parsed upstream endpoint URLs per ProviderKey

membphis:
# perf(gateway): cache parsed upstream endpoint URLs per ProviderKey

## What

Every bridge dispatch rebuilt its upstream endpoint URL on every
request: base resolution (trim + endpoint-suffix scan + allocations), a
`format!`, and — because the request builder was handed a `&String` — a
full URL parse inside reqwest's `IntoUrl` (`into_url.rs`: a string is
`Url::parse`d on every call; a `Url` value is accepted as-is). The URL
is a pure function of the ProviderKey's configuration and the endpoint,
so this PR parses it once and hands the builder a cloned `Url` value.

New `aisix_gateway::url_cache`:

- Two-level map (ProviderKey id → endpoint literal → parsed `Url`), so
  the hit path performs no allocation for the lookup. Misses are
  single-flight per (key, endpoint): the row's shard guard is held
  across the build, so racing cold requests serialize and the build
  closure runs once per transition.
- **Fingerprint revalidation instead of invalidation hooks**: each row
  stores the raw inputs it was built from (`api_base` + the per-bridge
  second component — vendor for the OpenAI-compatible family, deployment
  for Azure). A hit whose fingerprint differs rebuilds, so an edited
  ProviderKey takes effect on the next request.
- A malformed URL is returned raw and never cached — the request builder
  surfaces exactly the error it always produced for a bad `api_base`.
- Bounded: the outer map caps at 8192 ProviderKey ids and resets on
  overflow. (Like the per-key TLS client cache, rows are never
  individually evicted; unlike it, this cache adds the hard cap.)
- Contexts without a ProviderKey id (tests, standalone constructions)
  bypass the cache entirely.

Wired call sites:

- OpenAI-compatible family bridge: chat, streaming chat, completions,
  embeddings, image generation.
- Anthropic bridge: messages, streaming messages.
- Direct-dispatch proxy paths: native `/v1/messages` passthrough,
  `/v1/responses`, `/v1/rerank` (fingerprint carries the provider
  label — its URL falls back to a per-provider default host),
  `messages/count_tokens`, audio transcription/translation/speech.

Behavior note: on the wired bridge paths, base resolution now runs
inside the build closure (after credential extraction), so a
doubly-misconfigured key (bad base AND bad credential) reports the
credential error first; single-fault configurations are unchanged.

## Out of scope (tracked in #942)

- The A2A bridge: its upstream URL belongs to an agent resource, not a
  ProviderKey, and its bridge is constructed per call.
- Token-minting vendor bridges whose URL embeds per-request material
  beyond the deployment name.
- The Azure bridge (withdrawn after audit): its URL varies per
  deployment while this cache keys per ProviderKey×endpoint, so a key
  serving several deployments would miss every request and pay a
  rebuild+write per request. Needs a deployment-keyed inner level.
- Cold-path direct-dispatch sites (`realtime`, `jobs`, `videos`).

## Numbers (same-rig A/B)

Measured as the final leg increment on the same session as the
zero-config package PR: the step lands at +1.3 us/req, **below the
session's single-leg resolution** (+-1.5 us leg-to-leg noise; the
frame-share estimate for this mechanism was ~0.6 us). The mechanism
itself is verified by unit test — single-flight: the build/parse
closure runs once per (key, endpoint) transition, including under a
barrier-based 8-thread race — rather than by the throughput number. The combined default state (this PR + the
zero-config package) measured 38,983 rps / 102.45 us/req at c=128
(anchor 35,440 / 112.69; fail=0; anchor drift -0.30%), with the full
grid c=16 34,432 / c=32 37,164 / c=768@10ms 33,620.

## Tests

- Unit: four url_cache tests (per-key+endpoint caching with
  fingerprint-change rebuild; empty-id bypass + malformed-URL
  no-caching; build-error passthrough with nothing cached; concurrent
  single-flight under a barrier), plus a bridge-layer test pinning that
  an api_base edit on the SAME resource id reaches the new host on the
  next request. Full workspace suites green.
- E2E (vitest, real gateway + etcd + mocks): new
  `provider-key-api-base-repoint-e2e` — seeds a provider_key, warms the
  URL path against host A, re-points api_base to host B in etcd, and
  asserts every post-propagation request lands on B (the exact
  stale-URL regression this cache must never introduce). Full suite
  green on the combined branch in work-stealing mode; TPC mode green on
  the package branch.

### DIFF
--- a/crates/aisix-gateway/src/lib.rs
+++ b/crates/aisix-gateway/src/lib.rs
@@ -31,6 +31,7 @@ pub mod sse;
 pub mod upstream_headers;
 pub mod upstream_http;
 pub mod upstream_tls;
+pub mod url_cache;
 
 pub use bridge::{
     capture_in_band_error, capture_upstream_error_http, content_type_is_json, parse_retry_after,
--- a/crates/aisix-gateway/src/url_cache.rs
+++ b/crates/aisix-gateway/src/url_cache.rs
@@ -87,24 +87,21 @@ pub fn cached_endpoint_url<E>(
     // Clone the inner Arc out of the outer guard so no outer shard lock
     // is held while reading, building, or parsing.
     let known = cache.get(provider_key_id).map(|g| Arc::clone(&g));
-    if let Some(per_key) = known {
-        if let Some(hit) = per_key.get(endpoint) {
-            if hit.fingerprint.0 == fingerprint.0 && hit.fingerprint.1 == fingerprint.1 {
-                return Ok(EndpointUrl::Parsed(hit.url.clone()));
+    let per_key = match known {
+        Some(per_key) => per_key,
+        None => {
+            // First sighting of this key: bound the outer map before
+            // inserting. `len()` walks shards, but only key-creation
+            // (admin-rate) pays it.
+            if cache.len() >= MAX_PROVIDER_KEYS {
+                cache.clear();
             }
+            cache
+                .entry(provider_key_id.to_string())
+                .or_insert_with(|| Arc::new(DashMap::new()))
+                .clone()
         }
-        return build_into(&per_key, endpoint, fingerprint, build);
-    }
-
-    // First sighting of this key: bound the outer map before inserting.
-    // `len()` walks shards, but only key-creation (admin-rate) pays it.
-    if cache.len() >= MAX_PROVIDER_KEYS {
-        cache.clear();
-    }
-    let per_key = cache
-        .entry(provider_key_id.to_string())
-        .or_insert_with(|| Arc::new(DashMap::new()))
-        .clone();
+    };
     build_into(&per_key, endpoint, fingerprint, build)
 }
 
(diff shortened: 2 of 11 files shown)


membphis:
## Independent audit + bot review disposition

A cold-start independent audit ran against this PR, alongside the Copilot review. Findings and dispositions:

**MEDIUM-1 (audit + Copilot inline): high-traffic siblings unwired (native /v1/messages passthrough, /v1/responses, /v1/rerank) — FIXED in f806115.** All three are now wired through the same helper (rerank's fingerprint carries the provider label since its URL falls back to a per-provider default host). Remaining cold-path sites (realtime/jobs/videos) are tracked in #942 with the description corrected accordingly.

**MEDIUM-2 (audit): Azure wiring degrades under multi-deployment keys — FIXED in f806115 by withdrawing it.** The Azure URL varies per deployment while the cache keys per ProviderKey x endpoint; a resource key serving several deployments would fingerprint-miss every request and pay a rebuild + shard write per request. Withdrawn (also removes the cfg(test) bypass the audit flagged as LOW-4) and added to #942 with the deployment-keyed design note.

**MEDIUM-3 (audit) + Copilot single-flight finding — FIXED in f806115.** `build_into` now holds the row's shard guard across build+parse, so racing cold requests serialize and the closure runs once per transition (barrier-based 8-thread test added). Wiring-level coverage added at two layers: a bridge test pinning that an api_base edit on the SAME resource id reaches the new host on the next request, and a new e2e (`provider-key-api-base-repoint-e2e`) seeding a provider_key, warming the URL path on host A, re-pointing api_base to host B in etcd, and asserting every post-propagation request lands on B.

**LOW-5 (audit): error-priority reorder on doubly-misconfigured keys** — documented in the PR description (credential error now reported first when BOTH base and credential are invalid; single-fault behavior unchanged).

**LOW-6 (audit): cap-model comparison in the description** — corrected.

The audit explicitly verified the security question — five config-change sequences (api_base edited/removed, vendor edited, TLS edited, PK deleted+recreated) — and found no path by which a cached URL routes a request or credential to a host the current config does not point at.

PR #948, opened 2026-08-12
perf(server): set TCP_NODELAY on accepted downstream connections

membphis:
## What

Wires `axum_server`'s `NoDelayAcceptor` into every listener the gateway serves on, so `TCP_NODELAY` is set on each accepted client connection.

Before this, Nagle's algorithm was on for every downstream connection: `axum_server` accepts with the option off and nothing turned it on — `grep -rn nodelay crates/` had no hits anywhere in the tree. reqwest already sets it on the upstream side, so only the client-facing half was affected.

## Why

Nagle holds a small segment back until the previous one has been acknowledged. A short buffered answer leaves in a single segment and never notices. A streamed response writes a frame at a time, and a frame produced while an earlier one is still unacknowledged is held in the kernel until the client's delayed ACK arrives — up to 40ms on Linux. That is a latency tax on exactly the responses a user watches arrive token by token (SSE chat streams, `/v1/messages` streaming, audio streaming).

HTTP/2 control frames (`WINDOW_UPDATE`, `SETTINGS` ACK) and the WebSocket surfaces run on the same accepted socket and inherit the option, so a large h2 upload no longer risks stalling on a held flow-control frame either.

**This is not a throughput claim.** The saturation benchmark drives non-streaming requests whose response fits a single segment, which Nagle never delays; the change is invisible there. No benchmark leg is attached for that reason.

## Scope

All four listener constructions, which is every socket the process accepts on:

- shared-runtime HTTP and HTTPS (`serve_http`) — proxy, admin, and metrics listeners all route through it
- thread-per-core HTTP and HTTPS (`run_tpc_worker`), one per worker

Under TLS the no-delay acceptor is slotted *beneath* the rustls one (`Server::map` + `RustlsAcceptor::acceptor`), so the socket option is set before the handshake rather than after. Wiring it with `Server::acceptor` instead would *replace* the rustls acceptor and serve plaintext on the TLS port — the guard test rejects that form explicitly.

## Tests

- `the_downstream_acceptor_sets_tcp_nodelay` — accepts a real connection, asserts the option is **off** beforehand (so the test cannot pass vacuously) and set afterwards.
- `every_listener_sets_tcp_nodelay` — source guard over the production half of `main.rs`: every listener construction must wire the acceptor, TLS sites must wire it under rustls with `.map`, no `use axum_server::` may hide a site from the probe, and the probe must still match at least four sites. A listener that misses this serves correctly, benchmarks identically, and only pays the tax on streamed responses, so nothing else would catch the drift. Both rules are mutation-verified.
- `cargo test -p aisix-server`: 92 passed.
- e2e vitest, both serving modes: TPC 193 files / 550 passed / 6 skipped, work-stealing 193 files / 550 passed / 6 skipped, no failures.

## Risk

`TCP_NODELAY` disables write coalescing in the kernel: slightly more small packets on streamed responses, in exchange for not waiting on a delayed ACK. It is the standard setting for a proxy (nginx ships `tcp_nodelay on`; reqwest already defaults it on for the upstream half of this same process), and it is applied unconditionally rather than behind a knob for that reason.

One behavior delta worth naming: the acceptor fails a connection if `set_nodelay` errors. On Linux the option has no socket-state precondition and cannot fail; on BSD-family kernels it can report a peer that has already gone away, and such a connection was not answerable regardless.

Source: perf program AISIX-Cloud#1259, todo 3 audit (main table row 12, proposal B).

## Summary by CodeRabbit

* **Performance**
  * Improved network responsiveness for HTTP and HTTPS connections by enabling TCP_NODELAY on accepted sockets.
  * Applied the optimization consistently across all supported server runtime modes.

### DIFF
--- a/crates/aisix-server/src/main.rs
+++ b/crates/aisix-server/src/main.rs
@@ -1477,24 +1477,30 @@ fn background_check_interval(snapshot: &aisix_core::AisixSnapshot) -> std::time:
     std::time::Duration::from_secs(min_interval.max(1))
 }
 
-/// Completes when the process receives SIGINT or SIGTERM (best-effort on
-/// Windows — Ctrl+C only) OR when another part of the system has already
-/// flipped the cancel channel.
 /// Set on every connection the gateway accepts from a client.
 ///
-/// Nagle's algorithm holds a small write back until the previous one has
-/// been acknowledged. A buffered answer leaves as one write and never
-/// notices — which is why this is invisible in a throughput benchmark —
-/// but a streamed one writes a frame at a time, and a frame that lands
-/// while an earlier one is still unacknowledged waits out the client's
-/// delayed ACK (up to 40ms on Linux) before it leaves the box. That is a
-/// latency tax on exactly the responses users watch arrive.
+/// Nagle's algorithm holds a small segment back until the previous one
+/// has been acknowledged. A short buffered answer leaves in a single
+/// segment and never notices — which is why this is invisible in the
+/// throughput benchmark — but a streamed one writes a frame at a time,
+/// and a frame that lands while an earlier one is still unacknowledged
+/// waits out the client's delayed ACK (up to 40ms on Linux) before it
+/// leaves the box. HTTP/2 control frames and the WebSocket surfaces run
+/// on the same accepted socket and inherit the option.
 ///
 /// `axum_server` accepts with the option off, and the gateway had never
 /// turned it on. reqwest already sets it on the upstream side.
+///
+/// A `set_nodelay` failure fails that connection: on Linux the option
+/// carries no socket-state precondition and cannot fail, and elsewhere
+/// it reports a peer that has already gone away — a connection that was
+/// not answerable regardless.
 const DOWNSTREAM_NODELAY: axum_server::accept::NoDelayAcceptor =
     axum_server::accept::NoDelayAcceptor;
 
+/// Completes when the process receives SIGINT or SIGTERM (best-effort on
+/// Windows — Ctrl+C only) OR when another part of the system has already
+/// flipped the cancel channel.
 /// Serve `router` on `addr`, choosing HTTPS when `tls` is configured and
 /// plain HTTP otherwise. Both variants honour the shared `cancel` watch for
 /// graceful shutdown so the proxy/admin surfaces stop in lockstep with the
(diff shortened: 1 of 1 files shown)


membphis:
Independent audit disposition (0 HIGH / 3 MEDIUM / 5 LOW), fixed in 6ab5df0.

**M1 — the constant inherited a stale doc paragraph.** Real: the constant landed directly under three lines ("Completes when the process receives SIGINT or SIGTERM…") that belong to `shutdown_signal` and had already drifted onto `serve_http` on `main`, so rustdoc rendered them as the constant's description. Fixed by moving the constant *above* that paragraph, which leaves the pre-existing drift exactly where it was rather than cleaning up code this PR does not otherwise touch.

**M2 — the guard rewarded a form that serves plaintext on the TLS port.** The best finding here. `Server::acceptor` *replaces* the acceptor, so `from_tcp_rustls(..).acceptor(DOWNSTREAM_NODELAY)` compiles, drops the rustls acceptor, and satisfies a check that only greps for the constant's name. Fixed: any `_rustls` construction site must use the `.map(|tls| tls.acceptor(..))` form, and the failure message says why. Mutation-verified — rewriting one TLS site to `.acceptor(..)` now fails with that line number.

**M3 — probe scope.** Fixed within the file: also matches `axum_server::bind`, and the production half must not contain `use axum_server::` (an import would hide a site from a probe that matches the qualified path). Not extended to a workspace-wide scan: the audit itself confirmed `serve_http` is the only production accept path and that every other `axum::serve`/`TcpListener::bind` in `crates/` is under `#[cfg(test)]`, and `aisix-server` has a single `[[bin]]`.

**L1 — new connection-drop path if `set_nodelay` errors.** Not changing the implementation. `TCP_NODELAY` has no socket-state precondition in Linux's `do_tcp_setsockopt`, so it cannot fail on the shipped targets; where it can (BSD-family `tcp_ctloutput` returning `ECONNRESET`), it reports a peer that is already gone, and that connection was not answerable either way. Carrying a hand-written fail-open acceptor to cover it costs more than it buys. Recorded here as the one behavior delta, and in the constant's doc comment.

**L2 — "single writev" was the wrong reason.** Correct: a response longer than one segment ends in a partial segment Nagle can hold. The benchmark's responses fit one segment; wording in the doc comment now says that, and the PR body is updated.

**L3 — underclaim.** Correct and now named: HTTP/2 control frames (`WINDOW_UPDATE`, `SETTINGS` ACK) and the WebSocket surfaces run on the same accepted socket and inherit the option, so an h2 upload can benefit too.

**L4/L5** — fixed: the split marker now matches the repo's `production_half` convention, and the `sites >= 4` message tells whoever consolidates the listeners to lower the count deliberately.

No production code changed in 6ab5df0 (the only non-test edit is the doc-comment move), so the two e2e mode runs reported above still stand.

membphis:
Bot review disposition.

**Copilot — the constant splits `serve_http`'s doc comment.** Same finding as the independent audit's M1; fixed in 6ab5df0 by moving the constant above that paragraph, which leaves the pre-existing drift (those three lines belong to `shutdown_signal`) exactly where it was rather than cleaning up code this PR does not otherwise touch.

**CodeRabbit — make the socket precondition deterministic by calling `set_nodelay(false)` before the initial assertion.** Not taking this one, deliberately. The assertion is the test's canary, not an incidental precondition: if a platform ever accepts with `TCP_NODELAY` already on, this PR is unnecessary there and the test should say so loudly — which is what the assertion message spells out ("if the platform default changed, this test no longer proves anything"). Forcing the option off first would make the assertion trivially true and hide exactly the condition worth knowing about. The shipped targets are linux-gnu, where the kernel default is Nagle-on unconditionally, so the test is deterministic in practice on every platform CI runs.

PR #951, opened 2026-08-12
[dont merge me!] fix(gateway): stop following upstream redirects on dispatched requests

membphis:
## What

Every client that carries a caller's request to an AI provider is now built from `dispatch_client_builder()` — `client_builder()` plus `redirect(Policy::none())`. An upstream that answers a dispatched request with a 3xx now surfaces as an upstream status instead of being obeyed.

## Why

reqwest follows up to 10 redirects by default and the gateway had never opted out. The error path was written as if it had: `BridgeError::http_status` collapses an upstream 3xx to 502, with the comment *"redirects we don't follow are 502-worthy"* (`crates/aisix-gateway/src/bridge.rs`). The gateway did follow them.

What a followed redirect actually did, per the follow-redirect middleware reqwest 0.12 installs unconditionally (`tower-http-0.6.8/src/follow_redirect/mod.rs:272-290`, implementing RFC 9110 §15.4):

| upstream answers | what the gateway then did |
|---|---|
| `301` / `302` on a POST | rewrote the method to `GET`, dropped the body and the payload headers, and returned the redirect host's answer to that bodyless GET as the completion |
| `303` | same, for any method |
| `307` / `308` | replayed the request body verbatim to the redirect host |

Credential handling makes it worse: `remove_sensitive_headers` (`reqwest-0.12.28/src/redirect.rs:239-250`) strips only `authorization`, `cookie`, `cookie2`, `proxy-authorization`, and `www-authenticate` when the hop crosses hosts. The vendor schemes that do not use `authorization` — Azure's `api-key`, Anthropic's `x-api-key` — travelled to whatever host the `Location` named. And none of it is visible: the access log records the configured endpoint, not the one that answered.

The regression test demonstrates the old behavior exactly. Reverting the one-line builder change makes it fail with `UpstreamStatus { status: 404 }` — the redirect target's answer to a GET it has no route for, returned to the caller as if the configured provider had said it.

## Behavior change (for release notes)

An upstream 3xx on a dispatched request now fails the call with **502** instead of silently following the redirect. AI provider APIs do not redirect POSTed inference calls, so no working deployment should notice; a deployment that had been *relying* on an endpoint redirect (for example an `api_base` pointing at a host that 301s to the real one) will now fail, and the fix is to configure the final URL in the Provider Key's `api_base`.

## Scope

Applied to the clients that dispatch a caller's request:

- the four provider bridges (`openai`, `anthropic`, `azure-openai`, `vertex`)
- the proxy's passthrough client (`/v1/messages`, `/v1/responses`, `count_tokens`, rerank, audio, videos, the jobs surface, the raw tunnel)
- the gateway's per-worker pool and per-ProviderKey TLS-override pool

Deliberately **not** a global default on `client_builder()`. Some outbound clients have callers for which a redirect is ordinary — an OpenAPI document fetched for the MCP surface, a telemetry endpoint behind a rewrite. The outbound clients that must not follow one (JWKS/OIDC discovery, MCP OAuth token, A2A) already refuse at their own construction site, for the same reason.

Bedrock dispatches on the AWS SDK's own hyper client, not reqwest, and is unaffected by this setting.

## Not a performance change

reqwest parses the next URL *before* it consults the policy (`redirect.rs:310` parses `attempt.location()`, then `policy.check(...)`), so refusing redirects saves no per-request CPU. The audit that found this (AISIX-Cloud#1259 todo 3, main table row 9) measured the redirect layer at ~2.25µs/request and concluded it is not recoverable from the application side. Nothing in this PR is expected to move a benchmark.

## Tests

- `upstream_redirects_are_not_followed` (openai bridge) — a `301` whose `Location` points at a second, healthy OpenAI-shaped upstream. Asserts the call fails 502 **and** that the second host is never dialed. Mutation-verified: with the old builder the test fails, and fails by returning the other host's response.
- `every_dispatch_client_refuses_redirects` (source guard) — every dispatch client must be built from `dispatch_client_builder()`. Provider bridges are matched by *shape* (`aisix-provider-*/src/bridge.rs`), so a new vendor crate is covered the day it lands; the gateway pools and the proxy passthrough client are named. Asserts the probe still finds at least six sites so it cannot pass vacuously.
- e2e `upstream-redirect-not-followed-e2e` — drives a real `/v1/chat/completions` through the gateway against a mock upstream that 301s to a healthy second upstream; asserts 502, no `choices`, and zero requests on the redirect target.
- `cargo test --workspace --no-fail-fast`: all green except `aisix-mcp connect_timeout_bounds_an_unreachable_upstream`, which fails identically on unmodified `main` in this sandbox (dial behavior under the local network policy; not reachable from this change, which does not touch the MCP client).
- e2e both serving modes: TPC 194 files / 551 passed; work-stealing 194 files, one file failed on an `etcd put` 503 timeout and passed 12/12 on rerun (known harness flake, not a product failure).

Source: perf program AISIX-Cloud#1259, todo 3 audit (main table row 9, proposal C).

## Summary by CodeRabbit

* **Bug Fixes**
  * Upstream HTTP requests now refuse redirects instead of following them.
  * Redirect behavior is consistently enforced across providers, MCP connections, guardrails, and proxy requests.
  * Requests that encounter redirects now return an appropriate upstream error without contacting the redirect destination.

* **Tests**
  * Added coverage for redirect handling, fallback behavior, and end-to-end gateway scenarios.
  * Added safeguards to detect outbound clients that do not use the standard redirect policy.

### DIFF
--- a/crates/aisix-gateway/src/lib.rs
+++ b/crates/aisix-gateway/src/lib.rs
@@ -51,7 +51,7 @@ pub use upstream_headers::{
     RESERVED_UPSTREAM_HEADERS,
 };
 pub use upstream_http::{
-    client_builder, dispatch_client_builder, error_with_causes, transport_error_message,
-    UpstreamHttpConfig,
+    client_builder, dispatch_client_builder, dispatch_client_fallback, error_with_causes,
+    transport_error_message, UpstreamHttpConfig,
 };
 pub use upstream_tls::TlsSettings;
--- a/crates/aisix-gateway/src/lib.rs
+++ b/crates/aisix-gateway/src/lib.rs
@@ -51,6 +51,7 @@ pub use upstream_headers::{
     RESERVED_UPSTREAM_HEADERS,
 };
 pub use upstream_http::{
-    client_builder, error_with_causes, transport_error_message, UpstreamHttpConfig,
+    client_builder, dispatch_client_builder, error_with_causes, transport_error_message,
+    UpstreamHttpConfig,
 };
 pub use upstream_tls::TlsSettings;
--- a/crates/aisix-gateway/src/upstream_http.rs
+++ b/crates/aisix-gateway/src/upstream_http.rs
@@ -173,13 +173,31 @@ pub fn client_builder() -> reqwest::ClientBuilder {
 /// the redirect turns the upstream's 3xx into the 502 the error path
 /// already describes.
 ///
-/// Scoped to inference dispatch on purpose. The other outbound clients
-/// have callers for which a redirect is ordinary — an OpenAPI document
-/// fetched for the MCP surface, a telemetry endpoint behind a rewrite —
-/// and the ones for which it is not (JWKS, OAuth token, A2A) already
-/// refuse it at their own construction site.
+/// The same reasoning covers the guardrail vendors: an inspection call
+/// POSTs the caller's prompt to an operator-configured endpoint under a
+/// vendor credential header (`Ocp-Apim-Subscription-Key`, and the rest),
+/// none of which reqwest strips on a cross-host hop either.
+///
+/// Not applied to every outbound client. The remaining ones either
+/// already refuse redirects at their own construction site (JWKS/OIDC
+/// discovery, MCP OAuth token, MCP OpenAPI tool calls, A2A) or talk to
+/// an operator's own collector, where an endpoint behind a rewrite is
+/// ordinary (telemetry, heartbeat, OTLP export).
 pub fn dispatch_client_builder() -> reqwest::ClientBuilder {
     client_builder().redirect(reqwest::redirect::Policy::none())
+}
+
+/// The client to fall back to when [`dispatch_client_builder`] fails to
+/// build — a malformed deployment CA, say.
+///
+/// The connection settings are lost either way; what must not be lost is
+/// the redirect refusal, which `reqwest::Client::new()` would silently
+/// restore. Builds from a policy alone, which cannot fail.
+pub fn dispatch_client_fallback() -> reqwest::Client {
+    reqwest::Client::builder()
+        .redirect(reqwest::redirect::Policy::none())
+        .build()
+        .expect("a client with only a redirect policy always builds")
 }
 
 /// Layer the outbound trust decision onto a builder. Split out so the
@@ -300,6 +318,50 @@ mod tests {
         assert!(client.is_ok(), "{:?}", client.err());
     }
 
+    /// Both dispatch clients hand a 3xx back to the caller instead of
+    /// following it — including the fallback, which is reached when the
+    /// deployment's TLS material fails to apply and which
+    /// `reqwest::Client::new()` would have quietly restored to
+    /// following-by-default.
+    #[tokio::test]
+    async fn dispatch_clients_hand_back_a_redirect_instead_of_following_it() {
+        use std::io::{Read, Write};
+
+        let listener = std::net::TcpListener::bind(("127.0.0.1", 0)).expect("bind");
+        let addr = listener.local_addr().expect("addr");
+        let server = std::thread::spawn(move || {
+            for _ in 0..2 {
+                let (mut socket, _) = listener.accept().expect("accept");
+                let mut buf = [0u8; 1024];
+                let _ = socket.read(&mut buf);
+                let _ = socket.write_all(
+                    b"HTTP/1.1 301 Moved Permanently\r\n\
+                      Location: http://127.0.0.1:1/elsewhere\r\n\
+                      Content-Length: 0\r\nConnection: close\r\n\r\n",
+                );
+            }
+        });
+
+        for (label, client) in [
+            (
+                "dispatch_client_builder",
+                dispatch_client_builder().build().expect("builds"),
+            ),
+            ("dispatch_client_fallback", dispatch_client_fallback()),
+        ] {
+            let res = client
+                .post(format!("http://{addr}/v1/chat/completions"))
+                .body("{}")
+                .send()
+                .await
+                .unwrap_or_else(|e| panic!("{label}: {e}"));
+            // Following it would have dialed port 1, where nothing
+            // listens, and surfaced as a transport error instead.
+            assert_eq!(res.status(), 301, "{label} followed the redirect");
+        }
+        server.join().expect("server thread");
+    }
+
     /// Every outbound HTTP client in the workspace must be built from
     /// [`client_builder`], or it silently keeps reqwest's defaults — no
     /// connect timeout, TCP keepalive off, and a 90s pooled-connection
@@ -331,9 +393,16 @@ mod tests {
                 if sanctioned_rmcp_site && line.contains("rmcp_reqwest::Client::") {
                     continue;
                 }
-                if line.contains("reqwest::Client::builder()")
-                    || line.contains("reqwest::Client::new()")
-                {
+                // Prose about a constructor is not a call to one.
+                if line.trim_start().starts_with("//") {
+                    continue;
+                }
+                // `Client::new()` unqualified, too: every one of these
+                // files imports the type, and the fallback arm of a
+                // failed build is where a bare client hides
+                // (`.unwrap_or_else(|_| Client::new())` gives back
+                // reqwest's defaults, redirect following included).
+                if line.contains("reqwest::Client::builder()") || line.contains("Client::new()") {
                     offenders.push(format!("{}:{}", file.display(), n + 1));
                 }
             }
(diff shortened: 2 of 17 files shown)


membphis:
Independent audit disposition (0 HIGH / 5 MEDIUM / 5 LOW). All five MEDIUM fixed in 1be3ac7.

**M1 — the guardrail clients still followed redirects.** The best finding: seven inspection clients POST the caller's prompt to an operator-configured endpoint under a vendor credential header, and `Ocp-Apim-Subscription-Key` is no more stripped on a cross-host hop than `api-key` is. That is this PR's own threat model with a different vendor, and "some outbound callers legitimately redirect" was never true of content moderation. All seven now build from `dispatch_client_builder`.

**M2 — `.unwrap_or_else(|_| Client::new())` restored redirect following.** Real, and invisible to both guards. Five dispatch sites (plus the rmcp line) fell back to reqwest's defaults exactly when the deployment's TLS material failed to apply. Added `dispatch_client_fallback()`, which keeps the redirect refusal when it cannot keep anything else, and a test that drives a real `301` through both the builder and the fallback — mutation-verified against `Client::new()`.

**M3 — the guard was shaped the wrong way round.** Agreed and inverted, as suggested: every `client_builder()` site in the workspace must either build a dispatch client or appear in an explicit `NON_DISPATCH` list with a reason. Added a staleness check on that list too, so an entry that no longer matches a real site has to be dropped rather than left as cover. Also extended the companion `no_production_code_builds_a_bare_reqwest_client` to the unqualified `Client::new()` spelling — every one of the five M2 fallbacks was written that way and none was flagged.

**M4 — the doc cited a client that does not exist.** Verified: MCP specs are read from the resource snapshot and never re-fetched (`aisix-mcp/src/openapi.rs`), and the OpenAPI tool-call client sets `Policy::none()` itself. Replaced with the real non-dispatch cases (the operator's own telemetry/OTLP/heartbeat collectors) and added MCP OpenAPI to the already-refusing list. Point taken that a wrong example in the source is how the next person decides not to fix M1.

**M5 — rmcp's own reqwest line.** Fixed rather than deferred: a `tools/call` carries caller-supplied arguments under the MCP server's `x-api-key` or bearer, which is the same shape. Its build-failure fallback got the same treatment as M2.

---

**L1 (3xx is now retryable) — accepted as is.** A redirect is deterministic, so the same-target retry is indeed wasted work. But `is_retryable` gates failover as well, and a Model Group whose one target is misconfigured to redirect is exactly the case where failing over to a healthy sibling is the better outcome. Trading that away to save one backoff on a misconfiguration is the wrong side of the trade; leaving it noted here instead.

**L2 (a bodyless 301 renders an empty `error.message`) — real, filed as #952** rather than widened into this PR: the fix belongs in the upstream-error renderer and would change the message for every empty-bodied upstream error, not just 3xx.

**L3 (a 301 HTML body can name the redirect target)** — pre-existing behavior for all error statuses, no credential is in the body, and the `Location` header is never read. Noted, not changed.

**L4 — the "no CPU saved" reasoning was imprecise.** Correct: `redirect.rs:310` is only reached on a 3xx. The accurate mechanism is that `FollowRedirect` is installed unconditionally and `call()` runs `policy.on_request(&mut req)` on **every** request (`tower-http-0.6.8/src/follow_redirect/mod.rs:216`), where reqwest does a full `Uri → String → Url` round trip (`reqwest-0.12.28/src/redirect.rs:337-339`), plus a per-request header clone and body-clone attempt. `Policy::none()` removes none of it — which is also what the audit that raised this measured (~2.25µs/request, AISIX-Cloud#1259 todo 3 row 9). PR body corrected.

**L5 (guard headroom)** — moot under the inverted rule; the floor now tracks the dispatch-site count (13) and any unclassified site is a hard failure regardless.

membphis:
Bot review disposition — Copilot's six comments all match findings from the independent audit and are fixed in 1be3ac7.

**The five `.unwrap_or_else(|_| Client::new())` fallbacks** (proxy passthrough client + the four provider bridges) — audit M2. Correct and well spotted: the fallback gave back reqwest's defaults, redirect policy included, exactly when the deployment's TLS material failed to apply. Added `dispatch_client_fallback()`, which keeps the redirect refusal when it cannot keep anything else, and pointed all five at it. The rmcp line had the same fallback and got the same treatment. A test now drives a real `301` through both the builder and the fallback, mutation-verified against `Client::new()`.

**The doc's OpenAPI example contradicts `aisix-mcp/src/openapi.rs`** — audit M4, and correct on the facts: that client sets `Policy::none()` itself, and MCP specs come from the resource snapshot rather than being fetched at runtime. Replaced with the real non-dispatch cases (the operator's own telemetry/OTLP/heartbeat collectors) and added MCP OpenAPI to the list of clients that already refuse.

membphis:
## Reference-implementation survey: what other AI gateways actually do about upstream 3xx

Moved to draft while this is settled. Below is what three mainstream AI gateways do, read from source rather than from docs prose, plus what the MCP reference implementations do — because the MCP and guardrail halves of this PR turn out to be a different question from the provider-dispatch half.

**Headline: the three do not agree, and none of them does what this PR does.**

### 1. Follows redirects, deliberately, on LLM dispatch — and refuses on one specific surface

Its shared HTTP handler — the one LLM dispatch, most integrations, and the guardrail calls go through — sets `follow_redirects=True` on both the async and sync clients:

- `litellm/llms/custom_httpx/http_handler.py:616` (async) and `:1163` (sync)
- `litellm/llms/openai/common_utils.py` sets it on the OpenAI client too
- `litellm/experimental_mcp_client/client.py` sets it on the MCP client

This is an opt-in, not an inherited default: httpx does **not** follow redirects by default ("By default, HTTPX will **not** follow redirects for all HTTP methods, although this can be explicitly enabled" — https://www.python-httpx.org/quickstart/ ). Someone chose it.

The same codebase refuses redirects where the target URL is user-supplied:

```python
# litellm/proxy/plugin_routes.py:314-316
# Do not follow redirects — a redirect to an internal URL would allow
# the plugin to SSRF the proxy into fetching arbitrary internal services.
resp: Final = await handler.client.send(req, follow_redirects=False)
```

and its SSRF-hardening helper tells callers the same thing: *"Callers should also disable follow_redirects to prevent redirect-based SSRF bypasses"* (`litellm/litellm_core_utils/url_utils.py:247-248`).

So the split there is not "LLM traffic vs everything else" — it is **operator-configured endpoint (follow) vs user-supplied URL (refuse)**.

### 2. Never follows, and never converts — the 3xx goes back to the caller

The AI proxying plugin is a filter chain over the gateway's ordinary proxy path (`kong/plugins/ai-proxy/handler.lua` registers filters; `kong/llm/plugin/shared-filters/normalize-request.lua` rewrites the body/headers and lets the normal proxy path dispatch). A reverse proxy does not follow upstream redirects — the 3xx, `Location` header included, is returned to the client verbatim.

Where it *does* make a direct call rather than proxying (`kong/llm/drivers/shared.lua:898 http_request()`, used by the request/response-transformer plugins), it uses `lua-resty-http`, which has no redirect following at all — the library exposes no `allow_redirects`/`max_redirects` option and never mentions redirect following ( https://github.com/ledgetech/lua-resty-http ).

### 3. Does not follow upstream 3xx by default, but uses redirects internally as a mechanism

Envoy passes a 3xx from upstream through to the downstream client unless the route opts in via `internal_redirect_policy`; when enabled, at most one redirect is followed by default ( https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto ).

What makes this one interesting is the opposite direction: its ai-proxy plugin **relies on** Envoy internal redirects for its own fallback path, and has had to build credential handling around them. From `plugins/wasm-go/extensions/ai-proxy/main.go`:

> Distinguish "first hop into this gateway" from "internal_redirect re-entry". […] Absence => first hop. Distrust any incoming `X-HI-ORIGINAL-AUTH` — it may have been set by a client or by an upstream cascaded gateway running its own ai-proxy. […] **SAFETY DEPENDENCY:** this signal is reliable only when external callers cannot supply `x-higress-fallback-from`.

A gateway that lets redirects into its request path ends up owning a credential-re-anchoring problem across hops. That is a data point *for* refusing them on dispatch — and a warning that "just follow them" is not free either.

### Where this PR sits

| | upstream answers a dispatched POST with 301 |
|---|---|
| SDK-shaped gateway (1) | follows it; caller gets the redirect target's answer |
| proxy-shaped gateways (2, 3) | returns the 3xx to the caller unchanged |
| **this PR** | **fails the call as 502** |

The third row is nobody's behavior. Refusing to follow matches (2) and (3); turning the refusal into a 502 does not — a proxy hands the 3xx back and lets the client decide. Worth deciding explicitly which of those two we want, because they are different products for the caller: 502 says "your provider is broken", a passed-through 301 says "your provider wants you somewhere else".

### The MCP half is a different question, and the reference implementations are against us

This is the part I think should not ship as written.

- The MCP Python SDK's shared client factory is documented as *"Always enables `follow_redirects`"* and sets `{"follow_redirects": True}` unconditionally (`src/mcp/shared/_httpx_utils.py:30,79`).
- The Rust SDK we actually build on keeps **two** clients and picks per operation:
  ```rust
  // rmcp: crates/rmcp/src/transport/auth.rs:102-118
  struct ReqwestOAuthHttpClient {
      follow_redirects: ReqwestClient,
      stop_redirects: ReqwestClient,   // built with Policy::none()
  }
  ```
  with the trait contract stated as *"They must honor the request's redirect policy"* (`:95`), and error variants for `OAuth discovery redirect to non-same-origin URL rejected` and `OAuth discovery exceeded N redirects` (`:82-84`).

That last one is a **third** policy shape neither this PR nor the survey above considered: follow, but bounded — same-origin only, with a hop limit. It is also a live hazard for the change in this PR: rmcp's `ReqwestOAuthHttpClient::new` takes the *caller-supplied* client as its follow-redirects half, so handing rmcp a `Policy::none()` client would silently disable the redirect following its OAuth discovery expects. In our tree that path is not currently reached (we mint MCP OAuth tokens ourselves in `aisix-mcp/src/oauth.rs`, which already sets `Policy::none()` deliberately), but the rmcp transport client change in this PR is one refactor away from that being wrong.

The MCP spec itself is silent on 3xx — it specifies `Accept` headers, session ids and 4xx handling, but states no client requirement about redirects ( https://modelcontextprotocol.io/specification/draft/basic/transports/streamable-http ). So "the spec says so" is not available to either side; the reference implementations are the only signal, and they follow.

### Guardrails

No spec, no reference-implementation consensus I could find. The argument in this PR (an inspection call POSTs the caller's prompt under a vendor credential header that reqwest does not strip cross-host) still stands on its own, but note that the one gateway with comparable guardrail integrations routes them through the client that *does* follow.

### Suggested way forward

1. **Split the PR.** Provider dispatch is where the evidence is clearest and the blast radius smallest; MCP and guardrails each need their own decision.
2. **Decide refuse-vs-passthrough for dispatch**, since the two proxy-shaped references pass the 3xx through and we would be the only one converting it to 502.
3. **Do not blanket-apply to the rmcp client** — both MCP reference SDKs follow, and rmcp's own OAuth machinery treats the policy as per-operation, not per-client.
4. Consider whether the honest general rule is the one the SDK-shaped gateway landed on: **follow for operator-configured endpoints, refuse for anything whose URL came from a request**. That rule would keep the A2A/JWKS/OAuth refusals we already have, and would *not* have produced this PR's provider-dispatch change.

Sources are file paths at current `main` of each project plus the linked docs; every quote above is copied from the source, not paraphrased.

PR #953, opened 2026-08-12
perf(proxy): load the config snapshot once per request and resolve the provider key once per emit

### LINKED ISSUE #941: Collapse repeated per-request snapshot loads and provider-key resolutions in the chat dispatch path

membphis:
Fixes #941.

## What this changes

A chat request used to load the `ArcSwap` config snapshot up to six times —
the auth extractor, dispatch, the quota gate, both `request_metrics` emits
and the usage-event emit each took their own — and to resolve the same
ProviderKey row three times, once for every emitter that wanted something
off it (`provider_key_name` for the two metric emits, `telemetry_tags` for
the usage event).

Both collapse:

- **One snapshot per request.** Each handler loads `state.snapshot` once at
  entry and threads `&AisixSnapshot` through dispatch, the quota gate and
  every terminal emitter. A non-streaming chat request now loads twice —
  the auth extractor, then the handler — instead of six times.
- **One ProviderKey resolution per emitted attempt.** `usage_attr::ResolvedPk`
  looks the row up once and hands out both the `provider_key_name` metric
  label and the `telemetry_tags` the usage event stamps.

## Why it is one PR across the whole family

`request_metrics::Upstream` now takes a `PkLabels { id, name }` pair instead
of a bare `provider_key_id`. That is the point of the shape: the emit helpers
are shared by every client-facing endpoint, so a bare-id field would let a
new call site silently reintroduce a per-emit lookup, and a defaulted `name`
field would let it silently emit `provider_key_name="unknown"` next to a real
id. Taking the pair makes both a compile error.

Every endpoint is converted here for the same reason — chat, messages
(+`count_tokens`), responses, completions, embeddings, rerank, images, audio,
videos, realtime, MCP, A2A, passthrough and the files / batches / fine-tuning
jobs surface, streaming and non-streaming branches alike. Nothing is left as
a follow-up.

## Observable behaviour change

**A request's view of configuration is frozen at handler entry.** An exporter
added while a request is in flight no longer reaches that request's usage
event; today each emit re-loads and would see it. This is the consistency
semantic the issue describes as "arguably more correct" — one request, one
config generation — but it is a real, if narrow, change and is called out
here rather than buried.

Long-lived work is deliberately excluded from the freeze: the end-of-stream
emits (chat, messages on both wire protocols, responses) and the realtime
session-end emit read a **fresh** snapshot instead of pinning the one the
request started on, because either can outlive several config generations,
and pinning would also hold an old snapshot alive for the life of the stream.

One label-resolution timing shift falls out of that: on the chat streaming
path the `provider_key_name` on `aisix_llm_time_to_first_token_seconds` used
to be resolved at request start while the same label on the adjacent
`record_usage` emit was resolved at stream end. Both now come from the single
stream-end resolution, so the two agree; the values differ from before only
if the ProviderKey was renamed mid-stream.

## Measurements

Same-rig anchored A/B, main → branch → main, each leg its own fat-LTO build,
full grid, 25s x 4 windows per point, every window `fail=0`, loadgen floor
641-654K.

| leg | c=16 | c=32 | c=128 | 768@10ms | c=128 CPU/req |
|---|---|---|---|---|---|
| main (front) | 34,063 | 37,398 | 39,741 | 34,183 | 100.5 us |
| **branch** | 33,752 | 37,781 | **39,743** | 34,332 | **100.5 us** |
| main (back) | 34,000 | 36,663 | 38,626 | 33,408 | 103.4 us |

**No regression, and the gain is not resolvable in this session.** The
front-to-back anchor drift at c=128 is −2.81% — larger than the effect being
measured. Against the front anchor the branch is +0.005%; against the
linearly interpolated anchor pool (the branch leg ran at the session's
temporal midpoint) it is +1.43% / −1.45 us. The honest reading is that the
true value lies somewhere in [0%, +1.4%]; the issue's own estimate of
0.4-0.6 us sits inside that band and this session cannot narrow it further.

**The census is the load-bearing evidence**, and it is unambiguous. Counting
the distinct call sites `SnapshotHandle::load` is sampled under, reproduced
independently on both main legs:

| call site | main (front) | main (back) | branch |
|---|---|---|---|
| `auth::authenticate_token` | 0.0388% | 0.0570% | 0.0748% |
| `chat::dispatch` → `chat::chat_completions` | 0.0323% | 0.0618% | 0.0326% |
| `request_metrics::record` | 0.0257% | 0.0572% | **gone** |
| `chat::emit_usage_event` | 0.0194% | 0.0333% | **gone** |
| `quota::reserve_policy_layers` | 0.0130% | 0.0238% | **gone** |
| `request_metrics::record_usage` | 0.0064% | 0.0143% | **gone** |
| **distinct call sites** | **6** | **6** | **2** |

Those six are exactly the six this issue named. The ProviderKey resolution
collapses the same way: two sampled call sites on main
(`record`, `record_usage`) to one on the branch.

A methodology note, because it cost a pass: comparing frame *shares* between
the two legs is not sound here. The two binaries are LTO'd independently and
disagree about what is inlined, the folded captures carry different total
sample counts (72.7G vs 99.9G), and at 0.01-0.07% per site the sampling
error dominates — the *unchanged* `auth` load site alone swings +93% between
the two main legs. The call-site SET survives all of that, because an emit
that no longer loads cannot appear as a parent of the load frame in any
sample.

## Series contract

Names, label sets and label values are unchanged, with one race-window
exception noted below.

A local main-vs-branch comparison boots both binaries against an identical
config and mock upstream, drives the whole handler family, and diffs the
`/metrics` series sets (metric name + sorted label `k=v` pairs, sample values
ignored, so a changed label *value* fails the diff):

- 24 driven cases: chat and chat-stream, `/v1/messages` on both the
  cross-provider and the native Anthropic-protocol path (each streaming and
  non-streaming), `count_tokens`, responses, completions, embeddings, rerank,
  images, speech, nine model-not-found failures and a 401.
- **859 series on each leg, zero missing, zero extra, no status drift.**

The one value that *can* differ from `main`: if a ProviderKey is deleted
while a request is in flight, `main` resolved `provider_key_name` from the
live snapshot at emit time and reported `"unknown"` next to the real id,
whereas the branch resolves from the request's snapshot and reports the real
name. The streaming paths mirror it in the other direction, since they read
fresh at stream end. Both are race-window-only and arguably the better
value, but "label values unchanged" is not literally true there.

Driving the single-attempt endpoints is deliberate: the last time this family
was touched, the real bug the review round caught was hiding in exactly the
handlers that never call `begin_in_flight`, and a chat-only comparison
reported clean.

## Tests

- Workspace unit tests green (`aisix-proxy` 894). The one failure on the
  author's box, `aisix-mcp connect_timeout_bounds_an_unreachable_upstream`,
  reproduces identically on unmodified `main` there — a sandbox-network dial
  difference, not a regression.
- New `usage_attr` tests pin the label contract that used to live implicitly
  inside the two deleted lookup helpers: both halves fall back to `"unknown"`,
  an unresolvable id still reaches the label verbatim, one resolution serves
  both the label and the tags, and `display_name` keeps its control-char strip
  and 256-char cap.
- A new e2e, `provider-key-name-pairing-e2e`, pins the contract *per series*:
  the readable name must ride the same sample as the id it was read off,
  asserted across the request, token and streaming-TTFT families and across
  chat, `/v1/messages` and `/v1/embeddings`. The pre-existing assertion was a
  whole-scrape substring check driven by chat alone — satisfied by one
  correct emit while every other endpoint reports `unknown`. Verified by
  mutation: breaking the name resolution turns the new spec red.
- e2e vitest green in both runtime modes: 194 files / 551 tests each.

## Scope notes

- Pure data-plane mechanics: no new or changed user-facing configuration
  surface, so there is no control-plane counterpart to land with it.
- No wire-format, request-transform or response-normalization change — the
  request-scoped config handle is an internal lifetime change only.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved quota enforcement and usage tracking consistency across chat, audio, images, embeddings, messages, videos, and other request types.
  - Enhanced provider-key attribution in metrics and telemetry, reducing mismatched or unknown labels.
  - Improved streaming completion and abandonment reporting with up-to-date usage information.
  - Preserved accurate billing and usage data when responses are blocked by guardrails.

- **Tests**
  - Added end-to-end coverage validating provider-key ID and name pairing across request metrics.


membphis:
## Review disposition

An independent cold-start audit (no shared context, briefed only with the PR
URL and the contract it claims) returned **0 HIGH / 4 MEDIUM / 5 LOW**, and
Copilot filed 4 inline comments. All HIGH and MEDIUM are resolved in
`1d703de`; the LOWs are resolved or answered below.

### MEDIUM — fixed

**M1 — a deleted exporter kept receiving in-flight requests' events,
including captured content.** The freeze applied to the exporter table too,
which the description only covered in the benign direction ("an exporter
*added* mid-request"). The consequential direction is deletion: an operator
who removes a `content_mode: full` exporter — say after finding it points at
the wrong tenant — would keep shipping prompt and response text to it for as
long as the longest in-flight request runs. Exporter membership is a
delivery-authorization decision, not a label, so it is no longer frozen:
`usage_attr::live_exporters` re-reads the live snapshot, guarded by an
emptiness check on the request's own snapshot, so a deployment with no
exporters still pays one relaxed atomic load and never reaches the reload.
Routed through one helper rather than 13 copies.

**M2 — the multipart handlers resolved config before the upload, not
after.** On `main`, `audio::multipart_dispatch` and `jobs::create_file`
loaded the snapshot *after* draining the request body; this PR had moved
that to handler entry. On a large upload over a slow link that made the
model resolution, the client-IP allowlist, the upstream credential and the
rate-limit policy table all resolve against config captured before the body
started arriving — a freshness regression, not the telemetry-only freeze the
description claimed. Both now load after the drain and hand the snapshot
back so the terminal emits still share one. Every other jobs handler takes
its body through an extractor, where entry and post-body are the same
instant. `/v1/realtime` got the related fix: one snapshot for the whole
pre-upgrade phase, so a refused upgrade cannot report against a different
generation than `prepare` resolved with.

**M3 — the series contract was pinned by nothing in the repo.** The only
`provider_key_name` value assertion in the tree was a whole-scrape
`toContain` driven by chat alone, which stays green while `/v1/messages`,
`/v1/responses`, `/v1/embeddings` or the streaming TTFT series report
`unknown`. `provider-key-name-pairing-e2e` now asserts the pair *within a
series*, across the request / token / TTFT families and across chat,
`/v1/messages` and `/v1/embeddings`. Mutation-verified: breaking the name
resolution turns it red.

**M4 — no perf measurement.** Added to the description: same-rig anchored
A/B plus the call-site census. Short version — no regression, the effect is
smaller than this session's −2.81% anchor drift and is not resolvable, and
the census shows `SnapshotHandle::load` collapsing from 6 distinct call
sites to 2 (reproduced independently on both main legs).

### LOW

- **L1** (`PkLabels` fields public, so the invariant was a convention) —
  fixed: private fields with accessors, so the pair can only come from one
  resolution.
- **L2** (the "one resolution per attempt" claim held only for chat) — fixed
  rather than reworded, and it is the same gap Copilot filed on
  `responses.rs`, `messages.rs` and `audio.rs`: all three now resolve the
  winning attempt once for both the metric emit and the usage event. The
  redundant `provider_key_id` parameter is gone from both emitters, so the
  id cannot drift from the name.
- **L3** (label values do diverge when a key is deleted mid-request) —
  accurate, and the description now says so instead of claiming values are
  unconditionally unchanged.
- **L4** (`OtlpHttpFanOut::gc` has zero callers, so deleted exporters'
  pipelines are never stopped) — pre-existing and out of scope for a
  surgical change; filed separately as #954.
- **L5** (the batch-attribution loop rebuilt the exporter list per model
  slice) — fixed, hoisted out of the loop.

Copilot's fourth comment (the realtime error path observing a second config
generation) is the `/v1/realtime` fix under M2.

Re-verified after all of the above: series comparison 859 identical with no
drift, workspace unit tests green, e2e green in both runtime modes at
194 files / 551 tests.

membphis:
Fixed in `82425b0` — the review is right and I had the harness contract
backwards.

`stop()` only terminates the binary; the etcd prefix, the tmp config dir and
the snapshot cache all survive it. That is deliberate for the restart
scenarios (`config-last-known-good`, `status-config`), where a successor app
reuses the same prefix and *its* `exit()` performs the cleanup. This spec
spawns no successor, so every run was leaving its provider keys, models and
caller key behind in the shared etcd instance. 202 of the 205 teardowns in
`src/cases/` use `exit()`; mine was one of the three that should not have.

Re-ran the spec after the change: still green.

Also for the record, since this PR was rebased onto `main` after the
review round: the four earlier Copilot comments (the second ProviderKey read
on `/v1/responses`, `/v1/messages` and audio, and the realtime error path
observing a second config generation) were all fixed in `97d86e9` — see the
disposition comment above. The rebase resolved a conflict in `realtime.rs`
with #947, which moved auth out of `prepare` and made the rejection arm
carry the resolved key: that structure is kept, with the single pre-upgrade
snapshot layered on top of it.

PR #955, opened 2026-08-12
perf(gateway): cache endpoint URLs on the A2A, Vertex and Azure bridges

### LINKED ISSUE #942: Extend upstream URL caching to the A2A bridge and token-minting vendor bridges

membphis:
## What

Extends the per-ProviderKey endpoint URL cache (`aisix_gateway::url_cache`, #945) to the dispatch paths the first pass left out, so every bridge that rebuilt a URL per request now parses it once.

Fixes #942.

## Measurement

No same-rig A/B is attached, and one is structurally impossible for this change: the benchmark harness drives `/v1/chat/completions` against an OpenAI-shaped mock upstream, so the Vertex, Azure and A2A code paths this PR touches never execute in it. The same mechanism on the half that *is* benchmarked (#945, chat family) came out at +1.32µs — the unfavourable direction, inside a ±1.5µs single-leg resolution — and shipped with the mechanism pinned by tests instead.

What can be measured is the work removed per affected request. Microbenchmark, 200k iterations per leg, release build, idle x86 dev box (**not** the Graviton rig), on the Vertex Gemini URL shape (~120 chars, the longest of the three):

| | ns/op |
|---|---|
| build + `Url::parse` (today) | 817.3 |
| cache hit (two-level lookup + `Url::clone`) | 100.3 |
| **saved** | **717.0** |

Caliber, so this is not over-read: it is a microbenchmark, not a request-level A/B; the cached leg allocates a row-key `String` per iteration where production reuses a thread-local one, so it is pessimistic on the cache side; the Azure URL is shorter and saves less; and the absolute numbers are x86, not the aarch64 rig.

~0.72µs is roughly 0.7% of the gateway's ~102µs CPU per request, on endpoints that also pay a token lookup, a JSON serialization, and a provider round trip. Worth having, invisible in any throughput number. (Issue #942 estimated <0.2µs — the measured figure is ~3.5× that, the same direction of under-estimate the program has seen before when reading costs off frame shares.)

## Why these were left out, and what each needed

**A2A** (3 sites) posts to the agent's configured endpoint — a string with nothing derived from the request. The string is therefore its own identity, so this needs no fingerprint: `cached_url` parses it once process-wide. An `HttpBridge` is constructed per request, so this turns a `Url` parse per request into a map lookup. Agent-card discovery reuses the same parsed endpoint, so discovery and dispatch cannot disagree about the agent's URL.

**Vertex** (9 sites) splits two ways. Seven embed the upstream model in the path (`publishers/<vendor>/models/<model>:<method>`); two — the OpenAI-compat shim rail — do not, because the shim keys off the body's `model` field. The seven take a row per (endpoint, model); the two shim sites take a plain per-endpoint row. The fingerprint is `api_base` + project + region, which is every input `resolve_api_base` and the path template read.

**Azure** (2 sites) is the case #945's audit pulled back on. The URL embeds the deployment name, so keying by ProviderKey alone made each deployment invalidate the previous one's row: a key fronting several deployments would miss and rebuild on *every* request — measurably slower than not caching. It now takes a row per deployment.

## The cache change that makes the model dimension possible

Folding the model into the fingerprint is exactly the Azure failure above, so rows are keyed by `endpoint \x1f upstream_model` instead. Three supporting changes, no behavior change for the existing callers:

- Rows are keyed by `Box<str>` rather than a `&'static str` literal, and the hit path reads under a **shared** guard (`get`) before reaching for the row's write guard — so a hit now takes a read lock where it previously took the shard's write lock.
- The fingerprint is a slice rather than a pair. Vertex's URL is derived from three inputs; the old two-slot tuple could not express that, and concatenating inputs to fit is how a cache starts missing edits.
- The model-keyed row key is assembled in a reused thread-local buffer (the same trick `aisix-obs`'s handle cache uses for label keys), so **the row key** allocates nothing on a hit. (The returned `Url` is cloned, which does allocate — that is the copy reqwest takes by value instead of re-parsing.) A model id containing the separator byte skips the cache rather than risk aliasing another row.

Rows per resource are bounded: the model dimension makes the row count scale with configured models rather than with the fixed endpoint set. A row past the cap is served **uncached** rather than evicting to make room — clearing would leave an above-cap configuration at a permanent 100% miss rate plus a `len()` shard walk per request, which is worse than not caching at all. Checked only on the miss path.

## Not wired, deliberately

The videos, jobs and realtime paths. They are cold — one URL per long-running operation or per WebSocket session, not per inference request — and the jobs URLs concatenate the forwarded client query string, which would have to become part of the cache key. Stated here rather than left as an unexplained gap.

## Tests

- `url_cache_holds_a_row_per_upstream_model` (Vertex) and `url_cache_holds_a_row_per_deployment` (Azure) — one Provider Key alternating between two models / two deployments, three rounds, each request must land on its own path. Both mutation-verified: dropping the model or deployment from the row key fails them on the mock's request-count expectations.
- `url_cache_follows_an_api_base_edit_for_the_same_key_id`, on both bridges — a re-pointed `api_base` must reach the model/deployment-keyed rows, not keep serving the host it was moved away from. Mutation-verified by removing `api_base` from the fingerprint.
- The Azure tests deliberately avoid `with_url_override`: that seam bypasses the cache, so a test using it proves nothing about this. A wiremock URI reaches the production path through `AzureUpstreamRef::resolve`'s verbatim-override branch.
- `no_dispatch_site_posts_an_unparsed_url_variable` — guard test: no production dispatch site in the provider or proxy crates may post a URL string. That is the shape that re-parses per request, and it is how three bridges stayed unwired through #945.
- `url_cache` unit tests: model rows do not evict each other, a model-keyed row and a plain row on the same endpoint name are distinct, a separator-carrying model id bypasses the cache, a fingerprint that gains a component counts as a change, and rows past the cap are served uncached without evicting warm ones.
- `cargo test --workspace --no-fail-fast`: all green except `aisix-mcp connect_timeout_bounds_an_unreachable_upstream`, which fails identically on unmodified `main` in this sandbox and is not reachable from this change.

Source: perf program AISIX-Cloud#1259 (todo 8 deferred list).

### DIFF
--- a/crates/aisix-a2a/src/bridge.rs
+++ b/crates/aisix-a2a/src/bridge.rs
@@ -278,6 +278,10 @@ pub trait A2aBridge: Send + Sync {
 #[derive(Debug)]
 pub struct HttpBridge {
     upstream: A2aUpstream,
+    /// The agent's JSON-RPC endpoint, parsed once process-wide instead of
+    /// by reqwest on every call. A bridge is built per request, so this
+    /// is one cache lookup per request in place of one `Url` parse.
+    endpoint: aisix_gateway::url_cache::EndpointUrl,
     client: reqwest::Client,
 }
 
@@ -287,6 +291,7 @@ impl HttpBridge {
     /// per-request, so a shared client does not lose the per-agent deadline.
     pub fn new(upstream: A2aUpstream) -> Self {
         Self {
+            endpoint: aisix_gateway::url_cache::cached_url(&upstream.url),
             upstream,
             client: shared_client(),
         }
@@ -402,8 +407,9 @@ impl A2aBridge for HttpBridge {
     async fn send(&self, request: &serde_json::Value) -> Result<serde_json::Value, A2aError> {
         let resp = self
             .prepare(
-                self.client
-                    .post(&self.upstream.url)
+                self.endpoint
+                    .clone()
+                    .post_on(&self.client)
                     .timeout(self.upstream.timeout)
                     .json(request),
             )
@@ -438,8 +444,9 @@ impl A2aBridge for HttpBridge {
         let resp = tokio::time::timeout(
             self.upstream.timeout,
             self.prepare(
-                self.client
-                    .post(&self.upstream.url)
+                self.endpoint
+                    .clone()
+                    .post_on(&self.client)
                     .header(reqwest::header::ACCEPT, SSE_CONTENT_TYPE)
                     .json(request),
             )
--- a/crates/aisix-a2a/src/bridge.rs
+++ b/crates/aisix-a2a/src/bridge.rs
@@ -324,8 +324,19 @@ impl HttpBridge {
     /// A registered URL that is already at the origin yields the origin
     /// candidates alone — the two bases coincide and probing twice is waste.
     fn agent_card_urls(&self) -> Result<Vec<reqwest::Url>, A2aError> {
-        let base = reqwest::Url::parse(&self.upstream.url)
-            .map_err(|e| A2aError::Connect(format!("invalid upstream url: {e}")))?;
+        // The same parse the POST path uses, so card discovery and the
+        // JSON-RPC call can never disagree about the agent's endpoint.
+        let base = match &self.endpoint {
+            aisix_gateway::url_cache::EndpointUrl::Parsed(url) => url.clone(),
+            aisix_gateway::url_cache::EndpointUrl::Unparsed(raw) => {
+                // `Unparsed` means the parse already failed; re-run it
+                // only to render the same message this returned before.
+                let cause = reqwest::Url::parse(raw)
+                    .err()
+                    .map_or_else(|| "not a valid URL".to_string(), |e| e.to_string());
+                return Err(A2aError::Connect(format!("invalid upstream url: {cause}")));
+            }
+        };
         let prefix = base.path().trim_end_matches('/').to_string();
         let mut urls = Vec::with_capacity(AGENT_CARD_PATHS.len() * 2);
         for path in AGENT_CARD_PATHS {
--- a/crates/aisix-gateway/src/url_cache.rs
+++ b/crates/aisix-gateway/src/url_cache.rs
@@ -151,6 +151,34 @@ thread_local! {
         const { std::cell::RefCell::new(String::new()) };
 }
 
+/// Parse `raw` once, for a caller whose URL *is* a configured string with
+/// nothing derived from the request — an A2A agent's JSON-RPC endpoint.
+///
+/// No resource id and no fingerprint: the string is its own identity, so
+/// an edited agent simply lands on a different row and a renamed or
+/// deleted one leaves a dead row behind, bounded by [`MAX_PARSED_URLS`].
+/// Callers whose URL is *built* from configuration want
+/// [`cached_endpoint_url`] instead, which revalidates against the inputs.
+pub fn cached_url(raw: &str) -> EndpointUrl {
+    static PARSED: OnceLock<DashMap<Box<str>, Url>> = OnceLock::new();
+    let cache = PARSED.get_or_init(DashMap::new);
+    if let Some(hit) = cache.get(raw) {
+        return EndpointUrl::Parsed(hit.clone());
+    }
+    let Ok(url) = Url::parse(raw) else {
+        return EndpointUrl::Unparsed(raw.to_string());
+    };
+    if cache.len() >= MAX_PARSED_URLS {
+        cache.clear();
+    }
+    cache.insert(Box::from(raw), url.clone());
+    EndpointUrl::Parsed(url)
+}
+
+/// Upper bound on [`cached_url`] rows, on the same reasoning as
+/// [`MAX_RESOURCES`].
+const MAX_PARSED_URLS: usize = 8192;
+
 /// This resource's rows, creating them on first sight. `None` means the
 /// caller passed no id and wants the cache bypassed.
 fn rows_for(resource_id: &str) -> Option<Arc<EndpointRows>> {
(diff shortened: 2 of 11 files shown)


membphis:
Independent audit disposition (1 HIGH / 2 MEDIUM / 6 LOW). HIGH and both MEDIUM fixed in 32cde09.

**HIGH-1 — the Azure wiring had no test reaching it.** Correct, and the worst place for it: every Azure dispatch test goes through `with_url_override` (which bypasses the cache) and none sets a resource id (which bypasses it again), so deleting the deployment from the row key or `api_base` from the fingerprint failed nothing. Two tests now drive the production path — a wiremock URI reaches it through `AzureUpstreamRef::resolve`'s verbatim-override branch, no override seam involved. One Provider Key alternating between two deployments must land on each deployment's own path every round; an edited `api_base` must re-point the deployment-keyed rows. Both mutation-verified: dropping the deployment from the key fails the first, dropping `api_base` from the fingerprint fails the second.

**MEDIUM-1 — the cap evicted instead of refusing.** The best finding here, and it inverts the PR's own thesis: a configuration above the cap clears every warm row each time a new one is asked for, rebuilds them all next round, and clears again — a permanent 100% miss rate plus a `len()` shard walk per request, which is slower than not caching. Rows past the cap are now served uncached, so warm rows keep serving and the set stays bounded and monotonic. The suggested `>=` also settles the off-by-one (L1): the cap holds at 1024, not 1025.

**MEDIUM-2 — Vertex's `cfg(test)` `api_base_override` was not a fingerprint input.** Correct, and correctly diagnosed as latent-until-this-PR: no existing Vertex test sets both an override and a resource id, but this PR introduces the second half of that pair. Taken the symmetric route rather than the fingerprint one — `url_cache_key` bypasses the cache while the seam is set, matching the Azure URL override.

**L2 — the row-key buffer was borrowed across the build closure.** Fixed as suggested (take/replace rather than `borrow_mut`), so a future build closure that reaches back into the cache allocates its own buffer instead of panicking in the request path.

**L6 — A2A's third site.** Wired: `agent_card_urls` now reuses the same parsed endpoint the JSON-RPC POST uses, so discovery and dispatch cannot disagree about the agent's URL. PR body corrected to 3 sites.

**L3 / L4 / L5 — PR body inaccuracies.** All three stand and are corrected in the body: "the hit path allocates nothing" was about the *row key* (`Url::clone` allocates); there was no `cfg(test)` cache bypass on `main` for this to remove; and the reason the videos/jobs/realtime paths are left out is that they are cold and the jobs URLs concatenate the forwarded client query string (which would have to be in the key) — not that all of them embed a per-operation id.

**Guard test added**, per the audit's closing suggestion and the handler-family lockstep rule: no production dispatch site in the provider or proxy crates may post a URL string. That is the shape that re-parses per request and it is how three bridges stayed unwired through #945 — the request succeeds, the tests pass, and the site is simply absent from the cache. Mutation-verified.

The A2A `api_base` userinfo observation (no `@` rejection, unlike Vertex/Azure) is a real pre-existing gap and is not touched here; filing it separately rather than folding it into this PR.

membphis:
The A2A userinfo gap from the audit is filed as #956 (pre-existing; the Vertex and Azure bridges reject an `api_base` with embedded userinfo, A2A does not).

membphis:
Bot review disposition — both Copilot comments match independent-audit findings and are fixed in 32cde09.

**The per-resource cap is off by one (1025 rows, not 1024).** Correct. Fixed as part of the larger cap change (audit MEDIUM-1): the condition is now `>=`, and a row past the cap is served **uncached** rather than evicting to make room. Evicting was the worse half of this: an above-cap configuration would clear every warm row each time a new one was asked for, rebuild them all next round, and clear again — a permanent 100% miss rate plus a `len()` shard walk per request, i.e. slower than not caching at all.

**The Azure cache path is not exercised by the existing dispatch tests.** Correct, and the sharpest finding on this PR — the tests bypass it twice over (no `provider_key_id`, and `with_url_override` short-circuits ahead of the cache), so deleting the deployment from the row key failed nothing. Two tests now drive the production path via `AzureUpstreamRef::resolve`'s verbatim-override branch, with no override seam involved; both mutation-verified.

membphis:
Erratum on the measurement caliber, for anyone citing this later.

The figures in the description and the merge commit — 817.3 → 100.3 ns/op, 717.0 ns saved — were taken on a local x86 development box. The same benchmark on the project's reference machine (aarch64 Graviton3, same source, same 200k iterations, same Vertex Gemini `:generateContent` URL shape) gives:

| | ns/op |
|---|---|
| build + `Url::parse` | 1008.4 |
| cache hit | 121.2 |
| **saved** | **887.2** |

**The rig figures are the ones to quote.** The rig is the shared, reproducible reference every other number in this program is anchored to; a local development box is not something a reader can measure against. The x86 pair understates the effect on the shipped architecture, so it is superseded rather than contradicted.

887.2 ns is ≈0.87% of the ~102.4µs/request anchor. The caliber caveats are unchanged and still apply to both: this is a microbenchmark rather than a request-level A/B, and the endpoints this PR touches (Vertex, Azure, A2A) do not execute at all under the throughput benchmark, which is why no rig A/B leg is attached.

Issue #860, opened 2026-07-31
Upstream TLS trust: no config surface for a private CA, and no verification override


janiussyafiq:
## Description

Outbound TLS trust has no configuration surface. There is no `ssl_verify`-style switch and no way to point the gateway at a private CA bundle from config:

- no `tls` / `ca_file` / `verify` field in the `upstream:` config section, which otherwise carries the connection-layer settings (timeouts, keepalive, pool, retries);
- no TLS field on any resource in `schemas/resources/` — in particular `provider_key`, which is where an upstream endpoint (`api_base`) is declared;
- `AISIX_UPSTREAM_SSL_VERIFY` is rejected at boot:

  ```
  unknown field `upstream_ssl_verify`, expected one of `etcd`, `resources_file`,
  `proxy`, `admin`, `observability`, `cache`, `ratelimit`, `upstream`,
  `downstream`, `managed`, `bedrock_endpoint_url`
  ```

A private CA *can* be trusted today, but only as an OS-level side effect and only process-wide: `SSL_CERT_FILE` / `SSL_CERT_DIR` are honoured, and are additive to the built-in roots (setting `SSL_CERT_FILE` to a private CA alone still leaves public providers trusted). Neither is documented.

Requested, in order:

1. **Document the mechanism that already works** — `SSL_CERT_FILE` / `SSL_CERT_DIR`, and the fact that it is additive. Zero code.
2. **A first-class config surface for a private CA**, so trust is declared where the endpoint is declared rather than in the process environment — a deployment-level `upstream.tls.ca_file`, and/or a per-`provider_key` CA for deployments facing more than one private CA.
3. **An explicit, scoped verification override** for test environments, as the peers below provide. Not global, and not the default.

## Why

Self-hosted model endpoints and internal OIDC providers behind an enterprise CA are routine in on-prem deployments, and every outbound path is affected — the provider bridges, guardrails, MCP OAuth, A2A, OTLP export, and JWKS/discovery all share `aisix_gateway::client_builder()` (`crates/aisix-gateway/src/upstream_http.rs:121`).

The current answer is an environment variable that does not appear in the docs, the example config, or the resource schemas, so it is not discoverable from the failure. What operators see is a generic 502:

```
transport error: error sending request for url (https://...): client error (Connect): invalid peer certificate: UnknownIssuer
```

[...]

PR #961, opened 2026-08-12
build(deps): drop an unused dependency and narrow unused features

membphis:
## What

Dependency-hygiene pass over the workspace manifests: removes the one unused direct dependency and narrows four feature lists to what the code actually uses. No behavior change; every cut is backed by a zero-usage search across `src/` and `tests/` of all crates.

## Changes and evidence

| Change | Evidence | Tree effect |
|---|---|---|
| aisix-a2a: drop direct `http` dependency | zero `http::` references in the crate; its tests use `axum::http` (axum's re-export, not the direct dep). Only `cargo machete` finding in the workspace, manually verified (no cfg-gated / feature-gated / test-only use) | none (http stays in the tree for its real consumers); manifest-only cleanup |
| tower-http: `["trace", "cors", "limit", "compression-gzip", "set-header"]` → `["set-header"]` | `TraceLayer`/`CorsLayer`/`RequestBodyLimitLayer`/`CompressionLayer` have zero hits; body limiting is `axum::extract::DefaultBodyLimit` plus the crate's own `enforce_request_body_limit` (aisix-proxy/src/lib.rs); the only tower-http item in use is `SetResponseHeaderLayer` (server header) | tower-http's trace/cors/limit/compression modules stop compiling. Note: `async-compression` itself stays — reqwest's `gzip` feature (deliberately kept: Accept-Encoding negotiation) enables tower-http's decompression side |
| axum: drop `macros` | no `#[debug_handler]`, no `derive(FromRef)`; `FromRef` appears only as a trait bound, which the feature does not gate | drops the `axum-macros` proc-macro crate |
| uuid: drop `serde` | no `Uuid`-typed field crosses a serde boundary; every call site converts immediately (`to_string()`/`simple()`/`as_u128()`) | declaration-only |
| tracing-subscriber: drop `json` | no `.json()` formatter is constructed and there is no log-format config knob, so the JSON formatter is unreachable at runtime | drops `tracing-serde` |

## Deliberately kept (audited, not cut)

- **axum `tracing`** — has no API surface; it gates axum's internal logging of extractor rejections. It is also an axum default feature (default features stay on), so the explicit entry is documentation of intent rather than what enables it; the manifest comment records both facts so a future sweep does not re-flag it.
- **config `toml` / `json`** — statically zero-hit, but `config::File::from` infers the format from the file extension, so existing `config.toml` / `config.json` deployments depend on them. Removing them would be a breaking operator-facing change.
- **tokio `full`** — only `io-std` and `process` are unused, and both are thin internal tokio modules whose removal compiles out no third-party crate; narrowing buys nothing measurable.
- **reqwest features** — `gzip` (Accept-Encoding negotiation), `multipart` (audio), `stream`/`json` all in use; `rustls-tls-native-roots` is load-bearing (#860); `http2` is enabled by object_store's reqwest dependency either way.

## Verification

- `cargo check --workspace --all-targets --all-features` — clean
- `cargo test --workspace --all-features` — 0 failed
- `cargo clippy --workspace --all-targets --all-features` — no warnings; `cargo fmt --check` clean
- `cargo run -p aisix-core --bin dump-schema` — zero schema diff
- e2e (both serving modes) gated by CI

Stacked on #960: both PRs touch the workspace manifest and lockfile; this one retargets to `main` automatically when #960 merges.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -41,7 +41,6 @@ dependencies = [
  "async-trait",
  "axum 0.7.9",
  "futures",
- "http 1.4.0",
  "reqwest 0.12.28",
  "serde",
  "serde_json",
@@ -1078,7 +1077,6 @@ checksum = "edca88bc138befd0323b20752846e6587272d3b03b0343c8ea28a6f819e6e71f"
 dependencies = [
  "async-trait",
  "axum-core 0.4.5",
- "axum-macros",
  "base64 0.22.1",
  "bytes",
  "futures-util",
@@ -1171,17 +1169,6 @@ dependencies = [
  "sync_wrapper",
  "tower-layer",
  "tower-service",
-]
-
-[[package]]
-name = "axum-macros"
-version = "0.4.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "57d123550fa8d071b7255cb0cc04dc302baa6c8c4a79f55701552684d8399bce"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn",
 ]
 
 [[package]]
@@ -5205,7 +5192,6 @@ dependencies = [
  "tower 0.5.3",
  "tower-layer",
  "tower-service",
- "tracing",
 ]
 
 [[package]]
@@ -5265,16 +5251,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "tracing-serde"
-version = "0.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "704b1aeb7be0d0a84fc9828cae51dab5970fee5088f83d1dd7ee6f6246fc6ff1"
-dependencies = [
- "serde",
- "tracing-core",
-]
-
-[[package]]
 name = "tracing-subscriber"
 version = "0.3.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5284,15 +5260,12 @@ dependencies = [
  "nu-ansi-term",
  "once_cell",
  "regex-automata",
- "serde",
- "serde_json",
  "sharded-slab",
  "smallvec",
  "thread_local",
  "tracing",
  "tracing-core",
  "tracing-log",
- "tracing-serde",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


membphis:
Note on the stack: CI does not trigger while the base is `build/deps-security-convergence` (this repo's workflows only run against `main`-based PRs). Once #960 squash-merges and its branch is deleted, GitHub retargets this PR to `main`; it then needs a rebase onto main to drop the duplicated #960 commit, after which the full CI matrix runs. Local verification (unit suite, clippy, fmt, schema dump) already passed on this head.

membphis:
Closing as **not needed**, per maintainer decision. For the record, the full reasoning — this is a value judgment, not a correctness problem:

**Why it is safe (what closing does NOT mean).** The independent audit on this PR came back 0 HIGH / 0 MEDIUM: all five zero-usage claims were verified across src, tests, doc-tests, cfg-gated code and the lockfile, and the one LOW (the axum `tracing` comment wording) was already fixed on this branch. Nothing here was found wrong.

**Why it is not worth carrying anyway.**

- **No measurable benefit.** The only things that leave the build graph are `axum-macros` (a proc-macro) and `tracing-serde` — both tiny. The runtime binary is unaffected: under fat LTO the unused tower-http modules were already dead-code-eliminated, and `async-compression` stays in the tree regardless because reqwest's `gzip` feature (deliberately kept for Accept-Encoding negotiation) enables tower-http's decompression side. The compile-time and size wins measured in the hygiene pass (−3.3% cold build, −4.1% binary) came almost entirely from #960's removal of the retired hyper 0.14 + rustls 0.21 connector, not from these trims.
- **The cost is not zero.** It is one more lockfile-touching change to review, land and rebase around, for feature entries that are additive, compile-time-only, and harmless when unused.

**What survives this closure.** The durable output was the audit itself, and that is recorded (here and in the internal dependency-hygiene report): the unused-feature inventory with evidence, the deliberately-kept list (axum `tracing` is a default feature and gates rejection logging; config `toml`/`json` are operator-facing format support; reqwest features are load-bearing), and the note that aisix-a2a's direct `http` dependency is declared but unused. If a future manifest change wants to pick any of these trims up, the diff and the evidence tables in this PR are ready to reuse — nothing needs to be re-investigated.

PR #965, opened 2026-08-12
bench(onthebench): pin the dwarf stack dump and flag flamegraphs that did not unwind

membphis:
## What

Two changes to the rig harness, both about the same failure: a flamegraph
that renders fine and is entirely wrong.

1. **Pin the per-sample stack dump** to 32KB (`BENCH_PERF_STACK`, validated
   like the other `BENCH_*` knobs, forwarded to the rig by `bench.sh`).
2. **Check the collapsed output** after rendering and warn when the capture
   is unusable.

## Why

While taking the closing census for the performance program, a c=128
capture on a fat-LTO build came back with **95.5% of its stacks collapsed
to `[unknown]`**. perf's default dump is 8KB, which is smaller than this
gateway's async stacks under fat LTO, so libunwind had nothing to walk.

The failure is silent in every layer: `perf record` exits 0, `perf script`
produces output, inferno renders an SVG, and the picture reads as "the
kernel is 94% of the time". A re-capture of the same commit reproduced it
at 91.7%, so it is not a fluke. Attribution taken from such a graph is
wrong in a way that looks authoritative.

Same build, same load, after the fix:

| | before | after |
|---|---|---|
| collapsed stacks | 529 | 1876 |
| resolved user-space share | ~5% | 48.4% |

## The threshold is measured, not picked

The check warns above **80% unresolved**. That number comes from this
repo's own captures, not from judgement:

| capture | unresolved | verdict |
|---|---|---|
| old rig, thin LTO + glibc | 0.0% | usable |
| this rig, thin LTO + glibc | 37.0% | usable |
| this rig, fat LTO, 32KB dump | 51.1% | usable |
| this rig, fat LTO (earlier scene) | 56.6% | usable |
| this rig, fat LTO, 8KB dump | 91.7% | **failed** |
| this rig, fat LTO, 8KB dump | 95.5% | **failed** |

The two populations separate cleanly. A tighter threshold flags healthy
captures (the 57% one is a real, usable profile); a looser one lets the
failure through.

Verified by running the check against all six captures plus an empty file:
warns on exactly the two failed ones and the degenerate input, silent on
the four usable ones.

## Why warn instead of abort

A bad flamegraph does not invalidate a run's throughput numbers - they come
from the measured windows, not from perf. Aborting a 15-minute measurement
over an auxiliary artifact would be the wrong trade. The share is printed
on every capture either way, so a run's log always states how much of its
own profile is trustworthy.

## Note for anyone reading older profiles

Captures taken before this change were subject to the same effect. The two
at 92-96% are unusable outright; the ones in the 51-57% band are fine, but
their "kernel" share is an upper bound - deep stacks that failed to unwind
land in the same bucket.

## Summary by CodeRabbit

* **New Features**
  * Added configurable perf stack sizing for benchmark flamegraph captures.
  * Added validation for stack size values, including range and alignment requirements.
  * Added warnings when flamegraph stack symbols cannot be resolved reliably.
  * Documented the new benchmark override and its default constraints.

### DIFF
--- a/bench/onthebench/README.md
+++ b/bench/onthebench/README.md
@@ -72,14 +72,17 @@ Every method knob is a `BENCH_*` environment variable (defaults in `lib.sh`
 match the #891 grid exactly): `BENCH_GRID` (`"ttft_ms:conc ..."`, e.g.
 `"20:1536 20:2048"` for an added delay tier, `"0:128"` for a spot-check),
 `BENCH_REPS`, `BENCH_WINDOW`, `BENCH_WARMUP`, `BENCH_MAX_TRIES`,
-`BENCH_FLOOR_REPS`, `BENCH_FLAMEGRAPH`. `bench.sh` forwards them to the rig.
+`BENCH_FLOOR_REPS`, `BENCH_FLAMEGRAPH`, `BENCH_PERF_STACK`. `bench.sh` forwards
+them to the rig.
 A changed knob is recorded in `meta.json`, so a non-default run can never
 pass silently as the standard grid.
 
 Values are validated at startup and nonsense refuses to run: grid entries
 must be `ttft_ms:conc` with a positive concurrency; window, reps, max tries
 and floor reps must be positive integers (warmup may be `0`);
-`BENCH_FLAMEGRAPH` is `0` or `1`.
+`BENCH_FLAMEGRAPH` is `0` or `1`. `BENCH_PERF_STACK` is the per-sample stack
+perf copies for dwarf unwinding (default 32768; a multiple of 8 up to 65528) -
+raise it if a capture reports a high unresolved-stack share.
 
 ## Measuring another target ("entrant")
 
--- a/bench/onthebench/README.md
+++ b/bench/onthebench/README.md
@@ -61,10 +61,13 @@ Replicates the published onthebench setup (https://onthebench.ai/gateways/perfor
   VmHWM are in the metadata, alongside commit, binary sha256, instrument
   sha256s, core split, kernel, and the full method parameters.
 - **Flamegraph** — one on-CPU flamegraph at the c=128 saturation point per run
-  (`perf record -F 499 --call-graph dwarf` → inferno), the api7/aisix#847
+  (`perf record -F 499 --call-graph dwarf,32768` → inferno), the api7/aisix#847
   workflow. `rig-setup.sh` sets `kernel.perf_event_paranoid=1` (session-scoped,
   reverts on reboot) so an unprivileged run can sample its own process, and
-  `run-baseline.sh` refuses a stripped binary before wasting a run.
+  `run-baseline.sh` refuses a stripped binary before wasting a run. Every
+  capture reports the share of its stacks that failed to unwind, into both the
+  terminal and the collected `perf.log`; a capture that unwound poorly is
+  called out rather than left to look like a profile.
 
 ## Method overrides
 
--- a/bench/onthebench/bench.sh
+++ b/bench/onthebench/bench.sh
@@ -48,7 +48,7 @@ echo "== run baseline ($RUNID) =="
 # printf %q keeps values with spaces intact through the remote shell.
 ENVPASS="BENCH_SRC_COMMIT=$COMMIT BENCH_SRC_DIRTY=$DIRTY"
 for v in BENCH_GRID BENCH_REPS BENCH_WINDOW BENCH_WARMUP BENCH_MAX_TRIES \
-         BENCH_FLOOR_REPS BENCH_FLAMEGRAPH; do
+         BENCH_FLOOR_REPS BENCH_FLAMEGRAPH BENCH_PERF_STACK; do
     # if, not `[ ] &&`: harmless here, but the && form as a function's last
     # statement is exactly how the grid_concs regression happened once.
     if [ -n "${!v:-}" ]; then ENVPASS="$ENVPASS $v=$(printf %q "${!v}")"; fi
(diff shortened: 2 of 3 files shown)


membphis:
## Independent audit — findings and disposition

Cold-start review by an agent with no context on this work. It reproduced
every number independently and found the mechanism sound and the framing
wrong in three places. All findings fixed in `a0e4bec`.

### HIGH

**H1 — the new knob was not recorded in `meta.json`, while this PR extended a
README sentence promising that it is.** A run with `BENCH_PERF_STACK=8192`
would have been archivally indistinguishable from a default run — the exact
failure this PR exists to prevent. Fixed: `stack_bytes` and the real
`callgraph` string now land in `meta.json`.

**H2 — the evidence table was six captures out of forty-seven.** The full
corpus has **four** failures, not two:

| capture | unresolved | in the original table? |
|---|---|---|
| todo10 finale, aisix-final | 95.5% | yes |
| todo10 finale, re-capture | 91.7% | yes |
| todo7 A/B base-back | **90.0%** | **no** |
| todo7 A/B base-front | **88.1%** | **no** |
| issue941 A/B base-front | **71.0%** (highest passing) | **no** |

The six quoted numbers reproduce exactly, so the table was not wrong — it was
incomplete in a way that flattered the margin. The real separation is
**71 → 88 (17 points)**, not 57 → 92 (35). The threshold still sits in the
gap, but "a tighter threshold would flag healthy captures" holds down to ~72,
not ~58. The closing note about older profiles was also under-reporting the
damage and is corrected.

### MEDIUM

**M1 — the stated cause was backwards.** Fat LTO does not make stacks too deep
for the 8KB default. The first fat-LTO capture in the corpus (`cea29be`) is
**healthy at 43.2%**, and the two worst failures are its **thin-LTO** base
(`2cbf39c`). The same commit has produced both a 42% and a 71% capture, so
there is a large capture-to-capture component. Verified independently:
`git show 2cbf39c:Cargo.toml` → `lto = "thin"`, `cea29be` → `lto = "fat"`.
**32768 is a mitigation, not a diagnosis** — the measurement is what carries
the guarantee. Comments and this description corrected.

**M2 — the measured share reached no collected artifact.** It went to stderr
of the on-rig runner, i.e. to whoever was watching the terminal. `perf.log`
is rsynced back; the share is written there too now.

**M3 — "a symbol at any depth" is a weak signal.** A passing capture in the
corpus measures 71.0% by that metric but 93.1% per frame. The leaf frame is
what a flamegraph attributes self-time to, and separates the corpus more
widely (79.3% worst passing vs 90.9% best failing). Both shares are now
measured and either can trigger the warning.

**M4 — a 32KB dump writes ~4x the perf.data** (~1.5GB per capture, measured
from archived `perf.log`s), on a shared rig that has run out of disk before.
Free-space warning added before the capture.

### LOW

L1 comment had perf's rounding behaviour wrong in both directions (it rounds
up, and hard-errors above the bound rather than capping) — corrected against
`callchain.c: get_stack_size()`. L2 the awk assignment could abort the run
if the file were unreadable — guarded. L3 `length()` on a split array is a
gawk/mawk extension — now uses the return of `split`. L4 README still showed
the old perf invocation. L5 an empty capture now says so instead of
reporting "100% unresolved" like a collapsed one.

### Verified correct, left alone

The `awk && echo` idiom under `set -euo pipefail`, the validation chain's
precedence (tested 32768/8/65528 accepted; 65529/65536/0/-8/abc/`$(id)`
rejected), `--call-graph dwarf,N` syntax and the 65528 bound, `$NF` against
frame names containing spaces, and the `i = 2` start being load-bearing
rather than cosmetic.

Re-validated after the fixes: all four failures warn, the 71.0% capture
passes, and the share appears in `perf.log`.

membphis:
## Bot review disposition

**Copilot — `meta_method_json` records only `callgraph: "dwarf"`, contradicting the README's promise that a changed knob is recorded.** Agreed and fixed in `a0e4bec`; `meta.json` now carries `"callgraph": "dwarf,<n>"` and `"stack_bytes": <n>`. The independent audit raised this as its only HIGH, from the same reasoning — a run with a non-default stack size was archivally indistinguishable from a default one, which is the failure this PR exists to prevent.

**CodeRabbit — run the symbolization check after collapse succeeds, not after rendering.** Correct, and fixed in `5838749`. The check sat inside a branch requiring both `inferno-collapse-perf` and `inferno-flamegraph` to succeed, so a capture that collapsed but failed to render reported nothing — losing the diagnostic precisely when something had already gone wrong. Collapse and render are now separate steps; a failed collapse also stops reporting itself as a rendering failure.

Verified with a stubbed toolchain (collapse succeeds, render exits 3):

```
  flamegraph: 90.0% unresolved stacks, 90.0% unresolved leaves
WARNING: ... unwound poorly - do not attribute from this flamegraph
WARNING: flamegraph rendering failed; perf.data kept
```

The share reaches both the terminal and the collected `perf.log`, and `perf.data` is retained for an off-rig retry.

**CodeRabbit — README still shows the old `perf record` invocation.** Fixed in `a0e4bec` (CodeRabbit has since marked it addressed).

All three bot findings are fixed in code; none were dismissed.

membphis:
## Verified on the rig

Ran on the real Graviton box (c=128 saturation, `fail=0`), not a stubbed toolchain:

| | before (same commit, 8KB default) | after (`dwarf,32768`) |
|---|---|---|
| collapsed stacks | 529 | **2,093** |
| unresolved share | 95.5% / 91.7% (two captures) | **57.8%** (leaves 69.3%) |
| guard verdict | would warn | **silent** — 57.8% sits inside the usable band |
| `meta.json` | `"callgraph": "dwarf"` only | `"callgraph": "dwarf,32768", "stack_bytes": 32768` |

Three things confirmed at once on live traffic: the larger dump genuinely
changes unwinding (4x the stacks, unresolved 96% → 58%), the 80% threshold
does not false-positive on a healthy capture from this rig, and the knob
reaches the archived metadata — which was the audit's HIGH finding.

PR #966, opened 2026-08-12
bench(onthebench): verify that the process being measured is the one this run started

membphis:
## What

After the readiness probe, assert two things for the mock, the gateway and
every entrant:

- **liveness** — our own process is still running
- **ownership** — the listener on the port is our pid (or a child of it, for
  a launcher that forks)

## Why

`wait_http_200` accepts any HTTP 200 on the port. It does not check who
answered. When a process loses the bind race — a stale mock from an
interrupted run, a concurrent session on a shared box — the loser exits on
`Address already in use`, and every window afterwards measures a stranger
with `fail=0` throughout and nothing in the log to suggest otherwise.

This has bitten the performance program three times:

1. A leftover mock on the rig answered a whole scene; the floor read 270K
   instead of ~640K and the scene was scrapped.
2. A sibling harness spent a full batch driving load at another session's
   mock. Its own mock had logged `Address already in use` 81 times and never
   bound once.
3. In that same batch, the poisoned floor then **mis-fired the headroom
   criterion**: two legs were judged instrument-bound at 1.67–1.80× headroom
   and re-measured on a different core count, when the true headroom was
   11–12×. That round of work chased a problem that did not exist.

Point 3 is the reason this is worth a guard rather than a note. The
headroom criterion is correct; its *input* was poisoned. A correct
criterion fed a poisoned input produces confident wrong answers, which is
worse than having no criterion. Identity has to be established before
anything derived from the floor is trusted.

## Two checks, because neither covers the other

| | catches |
|---|---|
| liveness | the lost bind race, whatever the other listener is |
| ownership | a process alive but bound elsewhere, or answering from an earlier run of itself |

Ownership degrades to a warning when `ss` cannot attribute the socket (a
listener owned by another user needs privilege to attribute). Liveness never
degrades.

## Why not fingerprint the response

The stand-in mock in incident 2 answers **byte-identically** to ours — same
message body, same completion id — because it was written from the same
semantics. A response fingerprint cannot separate them. The listening pid
can. (This was my first suggestion to the affected session; their
measurement disproved it.)

## Why unconditional

The harness already asserts that a delayed mock actually delays, which would
have caught an impostor — but only on a grid that has a delay tier, and the
batch that went wrong had trimmed its grid to a single `0:128` point. **A
trimmed grid should cost coverage, not an assertion.** So the identity check
does not hang off a delay tier.

(The finale scene for the closing report had the same blind spot in its
flamegraph re-capture leg, which also ran `0:128` — it produced no
throughput numbers, so nothing was affected, but the gap was the same one.)

## Verification

Built the three failure modes and checked the guard's verdict on each:

| scenario | expected | result |
|---|---|---|
| listener is our own process | pass | pass |
| foreign listener holds the port, our process alive elsewhere | fatal | fatal |
| our process exited after losing the bind race | fatal | fatal |

The third is the exact shape of incident 2.

### DIFF
--- a/bench/onthebench/lib.sh
+++ b/bench/onthebench/lib.sh
@@ -144,15 +144,57 @@ wait_http_200() { # wait_http_200 <url> <label> [max_s]
     return 1
 }
 
+# An HTTP 200 says something is listening; it does not say the thing that is
+# listening is the thing this run started. A process that loses the bind race
+# exits, a stale listener from an earlier run keeps answering, and every
+# window afterwards measures a stranger with fail=0 throughout. Two checks,
+# because neither covers the other:
+#
+#   liveness    our own process is still up. This is the one that catches a
+#               lost bind race whatever the other listener is, since the
+#               loser exits on "Address already in use".
+#   ownership   the listener on the port is our pid. This catches a process
+#               that is alive but bound elsewhere, or answering from an
+#               earlier run of itself.
+#
+# Ownership degrades to a warning when ss cannot report the pid (a listener
+# owned by another user needs privilege to attribute); liveness never does.
+#
+# Response-shape fingerprinting deliberately not used: a stand-in mock built
+# from the same semantics answers byte-identically, so a fingerprint cannot
+# tell the two apart. The listening pid can.
+assert_listener() { # assert_listener <port> <pid> <label>
+    local port="$1" want="$2" label="$3" owners p
+    [ -d "/proc/$want" ] ||
+        { echo "FATAL: $label (pid $want) is not running after readiness - it likely lost the bind race on :$port; something else answered the readiness probe" >&2
+          exit 1; }
+    owners=$(ss -ltnpH "sport = :$port" 2>/dev/null | grep -oE 'pid=[0-9]+' | cut -d= -f2 | sort -u)
+    if [ -z "$owners" ]; then
+        echo "WARNING: cannot attribute the listener on :$port; $label identity unverified (liveness passed)" >&2
+        return 0
+    fi
+    grep -qx "$want" <<<"$owners" && return 0
+    # A launcher that forks the real listener is legitimate; a stranger is not.
+    for p in $owners; do
+        [ "$(ps -o ppid= -p "$p" 2>/dev/null | tr -d ' ')" = "$want" ] && return 0
+    done
+    echo "FATAL: :$port is held by pid $(echo "$owners" | paste -sd, -), not the $label this run started (pid $want) - measuring it would measure a foreign process" >&2
+    exit 1
+}
+
 start_mock() { # start_mock <ttft_ms>
     [ -n "$MOCK_PID" ] && { kill "$MOCK_PID" 2>/dev/null || true; wait "$MOCK_PID" 2>/dev/null || true; }
     MOCK_TTFT_MS="$1" taskset -c "$MOCK_CORES" "$TOOLS/mock" -port "$MOCK_PORT" \
         >> "$OUT/mock.log" 2>&1 &
     MOCK_PID=$!
     wait_http_200 "http://127.0.0.1:$MOCK_PORT$REQ_PATH" "mock (ttft=${1}ms)"
+    assert_listener "$MOCK_PORT" "$MOCK_PID" "mock"
     # A mock that ignores MOCK_TTFT_MS would make the delayed leg silently
     # measure a 0-delay upstream with fail=0 throughout; assert the delay is
-    # actually in effect before any window runs against it.
+    # actually in effect before any window runs against it. Note this only
+    # runs on a delayed tier: a grid trimmed to 0-delay points has no such
+    # assertion, which is exactly why the identity check above is
+    # unconditional rather than a side effect of running a delay tier.
     if [ "$1" -gt 0 ]; then
         local ttfb
         ttfb=$(curl -s --max-time 2 -o /dev/null -w '%{time_starttransfer}' -X POST \
--- a/bench/onthebench/lib.sh
+++ b/bench/onthebench/lib.sh
@@ -147,39 +147,65 @@ wait_http_200() { # wait_http_200 <url> <label> [max_s]
 # An HTTP 200 says something is listening; it does not say the thing that is
 # listening is the thing this run started. A process that loses the bind race
 # exits, a stale listener from an earlier run keeps answering, and every
-# window afterwards measures a stranger with fail=0 throughout. Two checks,
-# because neither covers the other:
+# window afterwards measures a stranger with fail=0 throughout.
 #
-#   liveness    our own process is still up. This is the one that catches a
-#               lost bind race whatever the other listener is, since the
-#               loser exits on "Address already in use".
-#   ownership   the listener on the port is our pid. This catches a process
-#               that is alive but bound elsewhere, or answering from an
-#               earlier run of itself.
-#
-# Ownership degrades to a warning when ss cannot report the pid (a listener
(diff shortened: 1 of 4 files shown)


membphis:
## Independent audit — findings and disposition

Cold-start review by an agent with no context on this work, verified
empirically on a live box including a real Docker container. It found the
guard's own failure handling broken. **Three HIGH, both MEDIUM, all fixed in
`2b07b8a`.**

### HIGH-1 — the advertised WARNING path was unreachable; the guard exited silently

Under `set -euo pipefail`, an `ss` that prints nothing or a `grep` that
matches nothing fails the pipeline, and the plain assignment aborts the
runner with `exit 1` and **no output at all**. The `if [ -z "$owners" ]`
branch was never reached.

Reproduced here before fixing:

```
$ bash -c 'set -euo pipefail; source al.sh; assert_listener 41977 $$ mock; echo REACHED'
about to call assert_listener on an unbound port
exit code: 1        # no FATAL, no WARNING, no output
```

The worst case is the one `run-entrant.sh` explicitly documents support for:
a containerized entrant whose pid belongs to another user. **This guard would
have killed those runs silently** — worse than the bug it exists to catch.

My own verification missed it because both failure cases I constructed had a
non-empty owner set. The empty case is the one that mattered.

### HIGH-2 — the degradation was backwards

`ss` attributes any socket owned by our own uid. So for the mock and the
gateway — both started by the harness under its own uid — an empty owner set
is not "unknown": readiness proved something answered, and this proves that
something is **not ours**. That is precisely the incident, and it was being
waved through. Now fatal, with the warning reserved for a target whose own
`/proc/<pid>/fd` we cannot read (the container case).

### HIGH-3 — membership, not exclusivity: a SO_REUSEPORT co-binder passed

Two `SO_REUSEPORT` processes both appear in the owner set; ours was among
them, so the guard returned 0 while the kernel split the load between us and
a stale instance — a silently halved-throughput run, since CPU is sampled on
`GW_PID` only. Third-party entrants commonly set `SO_REUSEPORT`. Every owner
must now be this run's process or a descendant.

(The audit checked and credited the product side: `aisix-server` does a
non-`SO_REUSEPORT` probe bind first, deliberately, so the baseline gateway
loses such a race loudly. The exposure was on the entrant path.)

### MEDIUM-4 — liveness was presented as the primary check but loses the race

Measured: a warm foreign listener answers readiness in **0.5–0.9 ms**, while
a losing process takes **~10 ms** to reach its `bind()` and exit. `wait_http_200`
returns on the first 200, so the loser is typically still alive when the
check runs — ownership carries the entire load in the incident shape.
Comment corrected; HIGH-2's fix is what closes the hole.

### MEDIUM-5 — the descendant fallback covered exactly one level

A wrapper that `exec`s collapsed to one level and passed; one that
backgrounds-and-waits produced a **false FATAL**. Which of two near-identical
idioms an entrant author picked decided whether the run aborted. Now walks
the ancestry chain, bounded at 10 hops, and warns when the serving process is
not the one CPU%/RSS are sampled on — matching the warning `run-entrant.sh`
already emits for the same hazard.

### LOW

`grep -oE 'pid=[0-9]+'` also matched a `pid=` substring inside a process
name, injecting a phantom owner (harmless before, a false FATAL once
exclusivity landed) — extraction now anchors on the `,fd=` that follows.
`ss` and `ps` became hard dependencies without being provisioned or verified;
added to `rig-setup.sh` next to the existing `perf --version` check.

### Verification after the fixes

| scenario | expected | result |
|---|---|---|
| own listener | pass silently | pass |
| foreign holder, our process alive elsewhere | fatal | fatal |
| SO_REUSEPORT co-binder present | fatal | fatal |
| nothing listening, target is ours | fatal | fatal |
| target owned by another user | warn, continue | warn |
| our process exited after losing the bind race | fatal | fatal |
| grandchild serves the port | warn, continue | warn |

Rows 3, 4, 5 and 7 are the ones the previous revision got wrong; 4 and 5 were
the silent exits.

### Verified correct, left alone

`sport = :PORT` matches `0.0.0.0`, `*` (IPv6 dual-stack) and `127.0.0.1`
bindings alike, so the bind-address/probe-address asymmetry is a non-issue.
`taskset` execs, so the recorded pid is the server's own. The `&&` chains are
exempt from `set -e` as non-final members of an AND-list. All three
`wait_http_200` call sites received the assertion — no deferred sibling.

membphis:
## Bot review disposition

Both Copilot findings were already fixed in `2b07b8a`, and both independently
match findings the cold-start audit raised — two reviewers arriving at the
same two defects from different directions is the strongest signal either of
them produced.

**Copilot (`lib.sh:171`, also 176) — under `set -euo pipefail`, a `grep` that
matches nothing terminates the harness before the intended warning branch
runs.** Correct, and it was the blocking defect: the advertised WARNING path
was unreachable, so every unattributable listener — including a containerized
entrant, a configuration `run-entrant.sh` documents support for — killed the
run with `exit 1` and no output at all. Reproduced before fixing:

```
$ bash -c 'set -euo pipefail; source al.sh; assert_listener 41977 $$ mock; echo REACHED'
exit code: 1        # no FATAL, no WARNING, nothing
```

Fixed with `|| true` on both stages. The empty-owner case then also got
stricter rather than more lenient: for a target this harness starts under its
own uid, `ss` would have attributed it, so an empty owner set is proof the
responder is foreign, not grounds to continue. It is fatal now, with the
warning reserved for a target whose own `/proc/<pid>/fd` is unreadable.

**Copilot (suppressed comment, `lib.sh:180`) — the check succeeds as soon as
any reported owner matches, so under `SO_REUSEPORT` a co-binding stranger
would still be permitted.** Correct, and worth un-suppressing: two processes
sharing a port means the kernel splits the load between them while CPU is
sampled on `GW_PID` only — a silently halved-throughput run that no validity
rule in this harness would catch. Fixed by requiring every owner to be this
run's process or a descendant of it, with the descendant walk extended past
one level (a wrapper that backgrounds-and-waits used to produce a false
fatal).

Verified across seven scenarios: own listener passes silently; a foreign
holder, a `SO_REUSEPORT` co-binder, an unbound port with a live target, and an
exited target are all fatal; a target owned by another user and a grandchild
listener warn and continue.

membphis:
## Verified on the rig against a real impostor

Both scenarios ran on the real Graviton box, not a constructed local case.

**Honest run — the guard stays out of the way.** Gateway measured at 38,409
rps / 103.8µs (in line with the 39.0K anchor), no warning, no fatal, run
completed normally.

**Impostor holding `:8000` — the guard stops the run.** A stand-in Python
mock was parked on the mock port before the harness started, answering the
readiness probe with a byte-identical body (same message, same completion
id — it was written from the same semantics, which is why fingerprinting
cannot separate them). Result:

```
FATAL: :8000 is held by pid 293041, not (only) the mock this run started
       (pid 294683) - measuring it would measure a foreign process
### exit code: 1 ###
```

The run aborted before a single measured window. Without this change it
would have driven every window at the stranger and reported `fail=0`
throughout — the exact shape of the incident that motivated the PR.

## A footgun this verification walked into itself

The verification script died silently after the impostor leg because its
cleanup line was `ssh RIG 'pkill -f impostor.py; …'` — **the remote shell's
own command line contains `impostor.py`, so pkill killed the ssh session**
(rc 255), and `set -e` took the script down with it. Fixed by making the
pattern not match itself: `pkill -f 'impostor[.]py'`.

Worth stating because it is this PR's own thesis from the other side:
**do not use pattern matching to refer to a process — use the pid.** The
harness now does that for the thing it measures; the surrounding scripts
should follow the same rule.

PR #976, opened 2026-08-13
build: add thin-LTO release-dev profile for iteration builds

membphis:
## What

Adds a `release-dev` cargo profile: inherits `release`, flips `lto = "fat"` → `"thin"` and `codegen-units = 1` → `16`. One file, 11 lines, purely additive — the shipped `release` profile is untouched. A follow-up commit hardens the bench run scripts' sanity gate (see below).

**Status: on hold pending a profile-matrix decision** — the LTO×PGO 2×2 was measured on the x86 probe rig (2026-08-14); results below.

## Why

Fat LTO + CU=1 serializes the whole LTO stage onto a single core. That is the right trade for shipped artifacts (measured per-request win, #939), but every link-heavy local rebuild pays minutes of single-threaded work, and the PGO pipeline (#975) pays it **twice per training cycle**: the changed `RUSTFLAGS` invalidates every fingerprint, so the instrumented and optimized builds are both full rebuilds.

## Build-time measurements (same 12-core host, cold builds, `--bin aisix`)

| profile | wall | binary size |
|---|---|---|
| `release` (fat, CU=1) | 6m37s | 52.9 MB |
| `release-dev` (thin, CU=16) | **2m27s** | 74.5 MB |

## Runtime: the LTO×PGO 2×2 (x86 probe rig, 0:128 grid, in-batch anchors, drift +0.24% / −0.07%)

| | no PGO | PGO (matched-config profile) |
|---|---|---|
| thin CU=16 | 25,866 rps (−9.9%) | 34,579 rps (+20.5%) |
| fat CU=1 | 28,700 rps (anchor) | 36,502 rps (**+29.0%**) |

Each cell normalized to its own batch anchor. Two findings:

1. **Fat LTO keeps a real edge under PGO: PGO+thin trails PGO+fat by ~5–7%** (anchor-normalized −6.6%, raw −5.3%) — outside the rig's ±2% decision band. The hypothesis that PGO would close thin's gap is refuted for this codebase.
2. **PGO profiles do not transfer across LTO configs.** A profile collected on a thin-LTO instrumented build and applied to a fat-LTO optimized build changed codegen (binary 52.9→57.4 MB) but delivered **+0.6%** instead of +29%. An earlier revision of this PR suggested building the instrumented leg with `release-dev` to halve pipeline time — that suggestion is withdrawn: **instrumented and optimized legs must use the same profile**, exactly as the Dockerfile already does.

## Discipline pinned in the profile comment

`release-dev` output is functionally identical but less optimized (−9.9% without PGO). Any number that is reported, compared against a baseline, or fed into the saturation grid must come from `--release`, never from this profile.

## Open decision

Whether `release-dev` ships at all is under discussion (single-profile preference vs the measured 5–7% cost of flipping `release` to thin). The bench sanity-gate commit stands on its own either way.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -207,10 +207,11 @@ strip = "debuginfo"
 
 # Iteration-speed sibling of `release`: thin LTO parallelizes the LTO
 # stage across cores instead of fat LTO's single-threaded whole-graph
-# pass, cutting link-heavy rebuilds from minutes to tens of seconds.
-# Functionally identical output, slightly less optimized — so any number
-# that gets reported, compared against a baseline, or fed into the
-# saturation grid MUST come from `--release`, never from this profile.
+# pass — measured cold builds of the aisix binary on a 12-core host:
+# 6m37s (release) vs 2m27s (release-dev). Functionally identical output,
+# slightly less optimized — so any number that gets reported, compared
+# against a baseline, or fed into the saturation grid MUST come from
+# `--release`, never from this profile.
 [profile.release-dev]
 inherits = "release"
 lto = "thin"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -205,6 +205,17 @@ lto = "fat"
 codegen-units = 1
 strip = "debuginfo"
 
+# Iteration-speed sibling of `release`: thin LTO parallelizes the LTO
+# stage across cores instead of fat LTO's single-threaded whole-graph
+# pass, cutting link-heavy rebuilds from minutes to tens of seconds.
+# Functionally identical output, slightly less optimized — so any number
+# that gets reported, compared against a baseline, or fed into the
+# saturation grid MUST come from `--release`, never from this profile.
+[profile.release-dev]
+inherits = "release"
+lto = "thin"
+codegen-units = 16
+
 [profile.coverage]
 inherits = "dev"
 opt-level = 0
--- a/bench/onthebench/run-baseline.sh
+++ b/bench/onthebench/run-baseline.sh
@@ -27,7 +27,12 @@ BIN="$SRC/target/release/aisix"
 
 # ---- sanity -----------------------------------------------------------------
 
-[ -x "$BIN" ] || { echo "FATAL: $BIN missing - build first"; exit 1; }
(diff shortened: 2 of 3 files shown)


membphis:
**Independent pre-merge audit (per repo rule 8) — verdict: approve; 1 MEDIUM, 2 LOW. All resolved or justified below.**

- **MEDIUM — stale-binary risk in bench run scripts.** `run-baseline.sh` / `run-decay.sh` only checked the executable bit on `target/release/aisix`; with iteration builds moving to `release-dev`, a stale release binary becomes a silent way to publish numbers attributed to the wrong commit. **Fixed in 6a42d78**: the sanity gate now runs `cargo build --locked --release --bin aisix` (a no-op when fresh, loud failure otherwise). Provenance was already covered — `meta.json` records `binary_sha256` and `commit`.
- **LOW — unmeasured claim in the profile comment.** "tens of seconds" had no measurement behind it. **Fixed in 6a42d78**: the comment now states only the measured pair (cold builds, 6m37s fat vs 2m27s thin).
- **LOW — staleness warning in `lib.sh`.** Skipped: the audit itself marked this as an alternative to the MEDIUM rebuild fix, which is the stronger of the two and is now in place.

No findings on correctness, reliability (no leak path of `release-dev` into CI/Dockerfile/PGO legs — verified by grep and by the audit), security, sensitive-info, or breaking changes.

membphis:
Closing per the measured verdict: the 2×2 in the description shows PGO+thin trailing PGO+fat by ~5–7% on the probe rig — outside the ±2% decision band — so `release` stays fat+CU=1 and the repo keeps a single release profile instead of gaining a second one. Two results worth keeping from the exercise: (1) fat LTO's edge is now a measured number at current HEAD (+29.0% with PGO vs +20.5% for thin, −9.9% for thin without PGO); (2) PGO profiles do not transfer across LTO configs (+0.6% when a thin-collected profile feeds a fat build) — the pipeline's matched instrumented/optimized legs are load-bearing and must stay matched. The bench sanity-gate commit (6a42d78) is closed along with this PR; the stale-binary risk it addressed largely disappears once iteration and measurement builds share `target/release` again, and it can be revived independently if that risk resurfaces.

PR #983, opened 2026-08-17
fix(passthrough): OTLP route/identity attribution, route on rejected events, header_key 401 names its header

jarvis9443:
Observability and error-surface follow-ups to #982, found by running the forward-proxy acceptance harness (a real TLS-terminating-device simulation relaying live traffic through a standalone gateway with an OTLP exporter attached):

- **OTLP spans now carry passthrough attribution.** The usage event has carried `passthrough_route_name` / `client_identity` since #982, but the OTLP sink builds span attributes by hand and had no passthrough branch — a relayed request exported as a bare `chat.completions` span with the route and end-user identity recorded nowhere, and `gen_ai.operation.name: chat` misclassified relays as model inference. Passthrough events now export as `passthrough <route>` with `aisix.passthrough.route_name`; `aisix.client_identity` exports unconditionally when set, so any future handler that populates it inherits the attribute. The JSON-serializing sinks (Datadog, SLS, object store) already carried both fields and are unchanged.
- **Rejected requests keep their route.** The passthrough error path emitted usage events through the shared `emit_error_usage_event`, which cannot set handler-specific fields — a 401/403/422 event lost `passthrough_route_name` even though the route had matched. The helper is split into `build_error_usage_event` + `emit_prepared_usage_event` (the 11 other call sites are untouched), and the passthrough arm attributes the route before emitting.
- **`header_key` 401s name the right header.** A missing gateway key on a `header_key` route answered "missing or malformed Authorization header" — but on those routes `Authorization` deliberately carries the caller's own upstream credential, so the message pointed at exactly the wrong header. New `MissingRouteAuthHeader` variant names the route's configured header.

- **The strict schema rejects cross-mode auth companions.** `auth_header_name` outside `header_key` and `anonymous_key_id` outside `anonymous` were accepted-and-ignored; two new coupling branches reject them (covering the absent-`auth_mode` default), with the CP enforcing the identical rule on its side. `source_cidrs` stays mode-agnostic — the gateway honors it as an extra IP allowlist on every mode.

Tests: two OTLP sink unit tests, a coupling test for the cross-mode rejections (passthrough span shape + protocol-ungated identity), and the header_key e2e scenario now asserts the 401 names the configured header and not `Authorization` (fails on the pre-fix message).

Verified live on the acceptance harness: the relayed span exports `passthrough copilot-auxiliary` + both attributes, and the 401 names `x-aisix-api-key`.

Refs api7/AISIX-Cloud#1312.

## Summary by CodeRabbit

* **New Features**
  * Added support for custom gateway-key authentication headers on passthrough routes.
  * Added route names and client identity details to exported telemetry.
  * Improved passthrough operation and span naming for clearer observability.

* **Bug Fixes**
  * Missing route authentication now returns a clear 401 error identifying the required header.
  * Authentication settings are now validated according to the selected mode.
  * Error telemetry now includes the matched passthrough route.

### DIFF
--- a/crates/aisix-core/src/models/passthrough_route.rs
+++ b/crates/aisix-core/src/models/passthrough_route.rs
@@ -379,6 +379,30 @@ pub fn passthrough_route_coupling() -> Value {
                 }
             }
         },
+        // Cross-mode leftovers are configuration errors, not ignored
+        // fields: a companion outside its mode is never consulted, so a
+        // row carrying one is rejected rather than half-honored. The CP
+        // enforces the identical rule on create and patch.
+        {
+            "title": "auth_header_name only in header_key mode",
+            "if": {
+                "anyOf": [
+                    { "title": "auth_mode omitted (defaults to gateway_key)", "not": { "required": ["auth_mode"] } },
+                    { "title": "auth_mode: gateway_key or anonymous", "properties": { "auth_mode": { "enum": ["gateway_key", "anonymous"] } }, "required": ["auth_mode"] }
+                ]
+            },
+            "then": { "not": { "required": ["auth_header_name"] } }
+        },
+        {
+            "title": "anonymous_key_id only in anonymous mode",
+            "if": {
+                "anyOf": [
+                    { "title": "auth_mode omitted (defaults to gateway_key)", "not": { "required": ["auth_mode"] } },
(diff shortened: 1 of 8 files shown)


jarvis9443:
@coderabbitai review

PR #999, opened 2026-08-19
feat(guardrails): local CPU embedding-model guardrail MVP (AISIX-Cloud#1331)

membphis:
MVP vertical slice of the second-tier guardrail (api7/AISIX-Cloud#1331): prove that a local CPU embedding model can sit inside the guardrail chain, produce a span-level judgement, and drive a real in-place rewrite on `/v1/chat/completions`. One hardcoded category (EDA software version numbers), one prototype vector encoded at load from a Chinese category description, hardcoded `***` replacement. Nothing else from the full design is included on purpose.

## What it proves (the four hard requirements)

1. **In-process inference**: `ort` 2.0.0-rc.13 + `tokenizers` 0.23 statically link ONNX Runtime into the data plane behind a new default-off `local-model` feature. Model deliverables are `model.onnx` + `tokenizer.json` only — the official int8 export of `ibm-granite/granite-embedding-97m-multilingual-r2` (`onnx/model_quint8_avx2.onnx`, standard `ai.onnx` opset; the q4/q4f16/bnb4 variants carry private ops and are excluded). Offline builds are supported through `ORT_OFFLINE=1` + `ORT_LIB_PATH` (verified in `ort-sys` `build/vars.rs`).
2. **Bounded, spin-free execution**: inference runs in `spawn_blocking` behind a `Semaphore(1)` sized to the session count, with one intra-op thread and intra/inter-op spinning explicitly disabled (the ONNX Runtime default spin burns ~9.4% of a core while idle).
3. **The model decision drives a real rewrite**: candidate spans come from a regex with exact byte offsets; a ±50-char context window is embedded (CLS pooling + L2 norm) and compared against the prototype by cosine; above-threshold spans are rewritten in place through the existing segment-moderation mask write-back channel (`moderate_input_segments` / `moderate_output_segments`) — no proxy pipeline changes. Verdict is always `Allow` (rewrite, never block); per-pass model-call cap of 8 degrades to fewer masks, never to blocking.
4. **Real HTTP end to end**: the opt-in e2e sends a request whose user text carries the sensitive sentence; the reply reaches the caller with the version number rewritten to `***` and the mock upstream's received body shows the masked prompt (the value never left the gateway). Verified against both debug and release binaries.

## Measured on a 12-core avx2+vnni host (release)

| Metric | Value |
|---|---|
| Single inference (~35-token window) | p50 19.0 ms (min 18.5 / max 20.7, n=20); ~7 ms at 7 tokens |
| Resident memory increment | +216 MiB (30.8 → 246.7 MiB after load + prototype inference; load 906 ms) |
| Release binary increment | +28.5 MiB (52.1 → 80.6 MiB) with the feature ON; default builds unchanged |

## Threshold calibration (and the headline negative result)

Default cosine gate 0.80: the acceptance-style positive scores ~0.90, every probed negative (compile-log timings, memory sizes, IPs, plain numbers) ≤0.76. Harder positives (`升级到 2022.4`, `Virtuoso IC6.1.8`) score 0.75–0.79 and are NOT masked — a deliberate precision-leaning default, since a mask false-positive corrupts user content. The negative result worth recording: across 5 prototype phrasings × 7 probe windows, single-prototype zero-shot cosine had **negative margin** over the hard negatives in every case. Closing that gap is the design issue's rule-scoring layer + real-sample prototypes, not threshold tuning.

## Configuration surface (deliberately minimal)

Activation is env-only on a feature build: `GUARDRAIL_LOCAL_MODEL_DIR` (switch + path) and optional `GUARDRAIL_LOCAL_MODEL_THRESHOLD`. The prefix is intentionally outside the `AISIX_*` namespace — the config loader maps every `AISIX_*` env var onto a config field and strictly rejects unknown ones. Feature on + env set + broken model dir is boot-fatal (a masking guardrail the operator asked for that silently isn't there would leak the very content it exists to rewrite). Feature off + env set logs a warning.

## Design notes per the repo research rule

- The window-around-candidate approach follows the keyword-proximity window pattern mainstream DLP engines use (typical proximity windows of 50–300 chars); the window size sits at the low end of that range.
- The full chain "anchor regex → window cut → model judgement → in-window second-pass locate and replace" has **no public precedent we could find** (the design issue reached the same conclusion); it is a custom contract surface and is flagged as such here explicitly.
- Upstream contracts pinned in the module docs: model repo pooling/normalization config (`1_Pooling/config.json`, `modules.json`), ONNX graph inputs/outputs, and the `ort-sys` offline-build variables.

## Explicitly deferred (tracked on api7/AISIX-Cloud#1331 — this PR is the DP half of a cross-plane feature)

- **Sibling endpoint families are NOT wired**: `/v1/messages`, `/v1/responses`, legacy completions, and MCP (the MCP mask write-back channel itself is blocked on AISIX-Cloud#1330). Chat-only is the MVP scope pin, stated here per the handler-family rule.
- **No control-plane surface**: no cp-admin.yaml resource, no dashboard, no attachment scoping — the prototype library as a first-class hot-reloadable resource (etcd stores category text; DP encodes vectors with its own model and atomically swaps on a watch tick) is the next-dimension design, recorded in the module docs.
- No rule-scoring layer (hotword co-occurrence), no standard risk categories, no multi-category, no capture-group replacements, no evaluation set. Streaming output rides the default `BufferFull` hold-back plus the same segment pass but has no dedicated e2e.
- Throughput is one lane (~50 inferences/s): scaling is N sessions behind `Semaphore(N)`; `ort`'s `run(&mut self)` currently forbids the shared-weights concurrent-run form the ONNX Runtime C API allows, so each lane costs its own ~190 MiB weight copy (details in the module docs).

## Test plan

- `cargo test -p aisix-guardrails --features local-model` (253 passed; model-backed probes run with the model dir env + `--include-ignored`).
- `cargo test -p aisix-proxy` (957 passed) — chain composition touched `chat.rs`.
- Existing mask-channel e2e regression: `guardrail-pii-redaction` + `bedrock-anonymize-mask` (11 tests) green against a feature-built binary with the env unset (proves inertness).
- New opt-in e2e `guardrail-local-model-e2e.test.ts` green against debug and release binaries.
- clippy clean on the touched crates (with and without the feature); default builds unaffected.

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in local CPU model guardrail for chat requests and responses.
  * Supports environment-based configuration for the model location, detection threshold, and processing lanes.
  * The guardrail masks detected sensitive content without blocking requests and is enabled through the server’s local-model feature.
* **Bug Fixes**
  * Improved handling of invalid thresholds, oversized candidates, tokenizer limits, and streaming output.
  * Invalid configuration now fails safely or reports startup errors when the feature is enabled.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -179,12 +179,14 @@ dependencies = [
  "base64 0.22.1",
  "chrono",
  "hmac 0.12.1",
+ "ort",
  "regex",
  "reqwest 0.12.28",
  "serde",
  "serde_json",
  "sha1",
  "thiserror 1.0.69",
+ "tokenizers",
  "tokio",
  "tracing",
  "tracing-subscriber",
@@ -1218,6 +1220,12 @@ dependencies = [
 
 [[package]]
 name = "base64"
+version = "0.13.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9e1b586273c5702936fe7b7d6896644d8be71e6314cfe09d3167c95f712589e8"
+
+[[package]]
+name = "base64"
 version = "0.21.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9d297deb1925b89f2ccc13d7635fa0714f12c87adce1c75356b39ca9b7178567"
@@ -1418,6 +1426,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "castaway"
+version = "0.2.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dec551ab6e7578819132c713a93c022a05d60159dc86e7a7050223577484c55a"
+dependencies = [
+ "rustversion",
+]
+
+[[package]]
 name = "cc"
 version = "1.2.60"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1552,6 +1569,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "compact_str"
+version = "0.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9dfdd1c2274d9aa354115b09dc9a901d6c5576818cdf70d14cae2bdb47df00ab"
+dependencies = [
+ "castaway",
+ "cfg-if",
+ "itoa",
+ "rustversion",
+ "ryu",
+ "serde",
+ "static_assertions",
+]
+
+[[package]]
 name = "compression-codecs"
 version = "0.4.37"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1666,6 +1698,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "crossbeam-deque"
+version = "0.8.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5181e0de7b61eb03a81e347d6dd8797bae9da5146707b51077e2d71a54ec0ceb"
+dependencies = [
+ "crossbeam-epoch",
+ "crossbeam-utils",
+]
+
+[[package]]
 name = "crossbeam-epoch"
 version = "0.9.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1709,13 +1751,43 @@ dependencies = [
 ]
 
 [[package]]
+name = "daachorse"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6f55d7153ba3b507595872a3874803f07a8a81d1e888abed8e5db7da0597d6e2"
+
+[[package]]
+name = "darling"
+version = "0.20.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fc7f46116c46ff9ab3eb1597a45688b6715c6e628b5c133e288e709a29bcb4ee"
+dependencies = [
+ "darling_core 0.20.11",
+ "darling_macro 0.20.11",
+]
+
+[[package]]
 name = "darling"
 version = "0.23.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "25ae13da2f202d56bd7f91c25fba009e7717a1e4a1cc98a76d844b65ae912e9d"
 dependencies = [
- "darling_core",
- "darling_macro",
+ "darling_core 0.23.0",
+ "darling_macro 0.23.0",
+]
+
+[[package]]
+name = "darling_core"
+version = "0.20.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0d00b9596d185e565c2207a0b01f8bd1a135483d02d9b7b0a54b11da8d53412e"
+dependencies = [
+ "fnv",
+ "ident_case",
+ "proc-macro2",
+ "quote",
+ "strsim",
+ "syn",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


membphis:
## Independent audit outcome (merge-gate record)

A cold independent audit reviewed this PR against the six-angle checklist (correctness, reliability, security, leakage, breaking changes, e2e coverage). Core contract held up: chain-composition semantics, mask/UTF-8 correctness, feature-off/env-unset inertness, no matched-value leakage in any log/error/count, and independently falsifiable e2e assertions. Findings and their resolutions:

**Fixed in `162c4f0`:**

- **HIGH — unbounded candidate span**: `\d+(?:\.\d+)+` matched an arbitrarily long `1.1.1…` run as one span, turning each model call into a max-truncation inference and stalling the single inference lane deployment-wide. Now double-bounded: spans over 64 bytes are dropped before budget accounting, and the tokenizer is forced to 512-token truncation at load instead of trusting the operator's `tokenizer.json`. Unit test pins the cap.
- **MEDIUM — semaphore permit vs cancellation**: the permit lived in the async scope, so a client disconnect released it while the blocking inference still held the session, letting blocked threads accumulate under repeated cancels. The permit now moves into the `spawn_blocking` closure and releases only when inference actually finishes.
- **MEDIUM — streamed overflow was fail-closed**: the inherited `BufferFull` default emitted a `content_filter` block when a streamed response crossed 256 KiB — contradicting this guardrail's rewrite-never-block contract. `stream_output_policy` now overrides to fail-open on overflow (past-cap degrades to unmasked release, same as the other fail-open arms); a stricter chain member still wins the fold.
- **LOW ×4**: poisoned session mutex now recovers instead of silently disabling masking until restart; the threshold env is validated finite and in `[0, 1]` (`"NaN"` previously parsed and made the gate never fire); the aisix-server feature comment named the e2e harness's opt-in var instead of the binary's `GUARDRAIL_LOCAL_MODEL_DIR`; `window_bounds(ctx == 0)` latent edge guarded.

**Explicitly justified (accepted, not code-changed):**

- **MEDIUM — coverage CI now compiles the feature**: the `--all-features` coverage job builds `local-model`, so `ort-sys` downloads the prebuilt ONNX Runtime in CI. Accepted: the download is SHA256-pinned inside the crate (verified against its manifest), all new deps are MIT/Apache-2.0, and this makes the coverage job the one place CI compiles and unit-tests the feature code (the pure-function tests need no model files). If CDN flakiness shows up, the mitigation is pre-caching via `ORT_LIB_PATH`.
- **MEDIUM — acceptance e2e does not run in CI**: it needs the 119 MiB model files, so it is opt-in via `AISIX_LOCAL_GUARDRAIL_MODEL_DIR` and skips in CI by design; this PR's acceptance evidence is the recorded local runs against both debug and release binaries. A cached-model CI job belongs to the productization pass tracked on api7/AISIX-Cloud#1331, alongside the control-plane surface this experimental env-gated slice deliberately lacks.

With the HIGH and both code-relevant MEDIUMs fixed and the remaining two MEDIUMs justified above, the merge gate is satisfied.

membphis:
## Addition: multi-lane inference pool (#1001), implemented in `77ad1b9` + `b04186e`

`GUARDRAIL_LOCAL_MODEL_LANES` (default 1, clamp 32) builds N sessions behind `Semaphore(N)` with centralized free-list dispatch — deliberately no worker↔session binding (sessions are stateless loaded-model instances; the central queue load-balances the uneven per-worker accept distribution). Tokenizer is shared; each lane keeps one intra-op thread + disabled spinning; the audit-driven cancellation semantics (permit inside the blocking closure) and poison recovery are preserved per lane. Default (unset/1) is behavior-identical to before.

Measured on the 12-core avx2+vnni host (release):

| Lanes | Throughput (64-inference saturating batch) | Scaling | Resident memory | Load |
|---|---|---|---|---|
| 1 | 52.0 inf/s | 1.00× | 222.9 MiB | 868 ms |
| 2 | 103.8 inf/s | 2.00× | — | — |
| 4 | 194.1 inf/s | 3.73× (93%) | 529.4 MiB | 1744 ms |

Per additional lane: ~102 MiB resident (its weight copy + arena; the first lane's ~192 MiB includes the shared tokenizer and one-time overhead) and ~290 ms boot. Tests: lanes parse/clamp unit test, a 2-lane 8-way concurrency probe asserting lanes agree, an env-driven throughput probe (`--ignored`), and the acceptance e2e now runs against a 2-lane pool.

membphis:
## Delta audit outcome (lane-pool commits `162c4f0` / `77ad1b9` / `b04186e`)

A second independent audit pass covered the post-audit hardening and the multi-lane pool. **No HIGH, no MEDIUM — nothing merge-blocking.** Verified clean: all seven prior-audit fixes survive the pool rework (permit-inside-closure ordering is session-freed-before-permit-freed, which is what makes the pool invariant sound); the shared tokenizer's concurrent `encode(&self)` is the upstream-intended use (checked against the locked tokenizers source — the encode-path cache is a contention-skipping RwLock, non-serializing); defaults are behavior-identical with lanes unset; no new log carries matched content.

Three LOW findings, dispositions:

- **Taken** (`6c0e709`): `Embedder::load` now clamps `lanes.max(1)` — `parse_lanes` can't produce zero, but the config fields are `pub` and the future control-plane integration will construct configs programmatically; an empty pool would have panicked at the boot-time prototype embed with a confusing error.
- **Kept as designed** — try_lock scan-miss under churn: a full scan can miss while releases refill behind the cursor, briefly serializing on the `sessions[0]` fallback. Bounded (bargers capped by the semaphore, ~19 ms holds, no deadlock, self-correcting), and the code already labels the fallback defensive. The race-free index free-list is the hardening if MAX_LANES ever grows materially.
- **Kept as designed** — a usize-overflowing lanes value parses as malformed and lands on 1, not the 32 clamp; consistent with the documented "malformed → 1" rule (disabling stays the model-dir env, minimum-surprise for garbage input).

Merge gate remains satisfied.

Issue #998, opened 2026-08-19
Audio transcription: lofty WARNs fail the e2e log scan, and the streaming relay changes the SSE event count


jarvis9443:
`TestHappyPath/audio_stt_real_openai` and `TestHappyPath/audio_stt_stream_probe` both fail against a live OpenAI upstream. They gate on `OPENAI_API_KEY`, so CI skips them and only a local run with the key set sees this.

Confirmed pre-existing: reproduced on `48ed6abb` (the commit before #1335 merged) with identical symptoms, so this is not fallout from the MCP ACL work that surfaced it.

Two separate problems.

**1. `lofty` warnings are not in the DP allowlist.** Every transcription request logs one, and the harness fails the suite on un-allowlisted WARN/ERROR:

```
WARN request{...}: lofty::iff::chunk: Chunk exceeds reader size, stopping
WARN request{...}: lofty::mpeg::properties: MPEG: Using bitrate to estimate duration
```

Six on `audio_stt_real_openai`, one on the stream probe. These look like ordinary metadata-parsing observations on a real mp3 rather than something the operator can act on — if that reading is right the fix is to drop them to DEBUG in the audio path, not to allowlist them.

**2. The streaming relay changes the SSE event count and collapses delivery.** The probe records:

```
DIRECT  status=200 first_byte=5.802s last_byte=7.278s reads=172 sse_events=184 bytes=12002
VIA DP  status=200 first_byte=6.233s last_byte=6.233s reads=3   sse_events=185 bytes=12059
```

The caller still gets correct text, but 184 upstream events arrive as 185, and the 1.476s delivery spread over 172 reads becomes 0s over 3 reads — the incremental delivery a streaming client subscribes to is gone. The count mismatch is the sharper half: one event is added or split somewhere in the relay.

Repro:

```shell
export OPENAI_API_KEY=...
AISIX_E2E_DP_IMAGE=ghcr.io/api7/aisix:dev \
  go test ./e2e/cases/... -timeout 30m -run 'TestHappyPath/audio_stt' -args -e2e
```

Issue #911, opened 2026-08-07
A2A: gateway never sends the A2A-Version header, so 1.0-pinned agents reject every call


janiussyafiq:
An A2A agent registered with `protocol_version: "1.0"` is unreachable through `/a2a/<agent>`, every call comes back with:

```
{"error":{"code":-32009,"data":[{"@type":"type.googleapis.com/google.rpc.ErrorInfo",
 "domain":"a2a-protocol.org","metadata":{},"reason":"VERSION_NOT_SUPPORTED"}],
 "message":"A2A version '0.3' is not supported by this handler. Expected version '1.0'."}}
```

The identical request sent straight to the agent with `A2A-Version: 1.0` succeeds, and adding or omitting that header on the gateway call makes no difference.

The gateway is the A2A client toward the upstream, and it never sends the `A2A-Version` header:

- `crates/aisix-proxy/src/a2a.rs:139` drops the inbound request parts (`let (_parts, body) = request.into_parts();`), so a caller-supplied `A2A-Version` never survives.
- `HttpBridge::send` and `fetch_agent_card` in `crates/aisix-a2a/src/bridge.rs` set only `Content-Type` and the gateway-held credential.
- `A2aAgent.protocol_version` is stored, validated and surfaced through the Admin API, but nothing in the data plane reads it. `git log -S "A2A-Version" --all` comes back empty — the header has never been sent, going back to the A2A DP MVP (#717) that introduced the field.

The spec requires a client to send `A2A-Version` on every request, and requires an agent to interpret an absent value as `0.3` and answer `VersionNotSupportedError` when that version isn't supported. So the upstream silently downgrades to 0.3 and correctly rejects a 1.0-shaped body. The agent is behaving properly here; the gateway isn't. In practice this means `protocol_version` is non-functional and no A2A 1.0 agent can be reached through the gateway.

Reproduced on main (2d7132e) against a stub agent that mirrors a2a-python 1.x negotiation: registering it with `protocol_version: "1.0"` reproduces the error above verbatim, and the stub logs no `A2A-Version` header on the inbound request from the gateway.

The fix should be to carry `protocol_version` onto `A2aUpstream` and send it as `A2A-Version` on both `send` and `fetch_agent_card`. Tried locally: 1.0 agents then work whether or not the caller supplies the header, pinning `0.3` sends `0.3`, and the `aisix-a2a` tests still pass.

[...]

Issue #545, opened 2026-06-08
Guardrails silently skipped on /v1/completions, /v1/images, /v1/rerank (same bypass class as #719)


moonming:
## Summary
#719 closed the guardrail-bypass on `/v1/responses` and `/v1/embeddings` (input) + `/v1/responses` (output). While auditing inbound surfaces for #719, three more text-bearing surfaces were found to invoke **no** guardrails at all — the same content-block bypass class:

| Surface | Scannable user text | Guardrail invocation (origin/main) |
|---|---|---|
| `/v1/completions` | `prompt` | **none** (`completions.rs`: 0 guardrail refs) |
| `/v1/images` | `prompt` | **none** (`images.rs`: 0 guardrail refs) |
| `/v1/rerank` | `query` + `documents` | **none** (`rerank.rs`: 0 guardrail refs) |

A customer who configures an input content/DLP block on `/v1/chat/completions` can still bypass it by sending the same text to any of these surfaces. This is the gap #379's checklist anticipated ("All inbound API surfaces … each runs configured guardrail").

## Scope to verify / fix
- [ ] `/v1/completions` runs input guardrails on `prompt`
- [ ] `/v1/images` runs input guardrails on `prompt`
- [ ] `/v1/rerank` runs input guardrails on `query` + `documents`
- [ ] output guardrails where the surface returns scannable text (`/v1/completions` text, `/v1/rerank` is scores — likely N/A; `/v1/images` returns image data — N/A)

## Approach
Same pattern as the #719 fix: resolve the per-request chain (`RequestContext { model_id, api_key_id, team_id }`), synthesize a `ChatFormat` from the surface's text fields, run `check_input` before dispatch; `Block` → 422 `content_filter`. `/v1/messages`, `/v1/chat/completions`, `/v1/responses`, `/v1/embeddings` already do this after the #719 PRs (#541, #544).

`audio.rs` (audio input, not text), `count_tokens.rs` (no model call), and `passthrough.rs` (generic) are excluded.

Surfaced while implementing #719. Refs #719, #379.

moonming:
Fixed by #551 (merged to main): input guardrails now run on /v1/completions, /v1/rerank, /v1/images, and /v1/audio/speech — closing the same surface-skip bypass class as #719. Closing as fixed.

Issue #542, opened 2026-06-08
Guardrail input check runs after rate-limit reservation on /v1/messages, /v1/responses, /v1/embeddings (burns a quota slot on a content block)


moonming:
## Summary
`/v1/chat/completions` deliberately runs input guardrails **before** `quota::enforce` so a content-policy-blocked request does not consume an RPM/RPD/RPS/RPH slot (chat.rs: "Run before reservation so a blocked prompt [doesn't] count against quota"). `/v1/messages`, and now `/v1/responses` + `/v1/embeddings` (#541), run the guardrail check **after** `quota::enforce`.

`Reservation::Drop` only releases the concurrency permit; it does **not** roll back the request-count increment from `pre_commit`/`check_and_increment`. So a guardrail-blocked request on these three endpoints still burns a request-quota slot that is never refunded.

## Impact
Minor: content-policy refusals count against the caller's request quota on 3 of 4 inference surfaces; cross-endpoint inconsistency with chat. Not a security issue.

## Fix options
1. Move the chain resolve + `check_input` **before** `quota::enforce` on messages/responses/embeddings (mirror chat.rs), or
2. Make `Reservation::Drop` refund the request-count increment when the request is rejected pre-dispatch.

Surfaced by the independent audit of #541. Refs #719.

moonming:
Fixed by #561 (merged to main, commit 69c7685): the input guardrail check now runs **before** `quota::enforce` on /v1/messages, /v1/responses, and /v1/embeddings, so a content block no longer burns a rate-limit slot. Closing as fixed.

PR #1015, opened 2026-08-21
feat(proxy): typed POST /v1/images/edits — multipart image editing endpoint

moonming:
## Problem

`POST /v1/images/edits` was the one OpenAI image endpoint the gateway did not serve. The editing models (`gpt-image-1`/`gpt-image-2`) take the source image(s), an optional mask, the prompt, and every tuning parameter in one `multipart/form-data` body — reachable until now only through a passthrough route, which loses model aliasing, model-level rate limits, prompt guardrails, and per-image usage accounting.

Tracking: AISIX-Cloud#1360 (Phase 1).

## Design

Follows the audio-transcription multipart shape (drain → resolve `model` → swap upstream id → rebuild → forward), not the JSON bridge dispatch generations uses:

- **Layering identical to `/v1/images/generations`**: `AuthenticatedKey`, `allowed_models`, client-IP allowlist, key- + model-level rate limits (reserve after the guardrail check, #542).
- **Guardrails**: input chain scans and masks every `prompt` form field (#545/#932 parity). No output hook — the response is an image.
- **Provider boundary**: `openai` only, same 400 `requires OpenAI` reject as generations (#168 parallel). Cross-provider editing wires (different URL + body shapes) are Phase 2 in the tracking issue.
- **Forward-compatible form**: unknown fields (`n`, `size`, `quality`, `background`, `input_fidelity`, repeated `image[]`, `mask`) forward verbatim — new upstream parameters need no gateway release.
- **Usage**: the gpt-image `usage` token block feeds the UsageEvent + TPM/TPD commit (#911 [21]); absent block emits a zero-token event. Endpoint label `/v1/images/edits`; handler family `images`.
- **`stream=true` rejects 400** (partial-image SSE): buffering a stream the caller asked to watch grow is silent degradation; live relay (#998-style) is a tracked follow-up in AISIX-Cloud#1360.

Per the reference-implementations rule: of five direct comparators checked, four ship a typed multipart images/edits route (one with a dedicated multipart helper and Azure/Vertex coverage, one as a plugin route type, two OpenAI-signature multipart); the two aggregator-style products fold editing into chat completions instead. This lands with the typed-route camp, consistent with our typed audio/files multipart surfaces. Upstream contract: <https://platform.openai.com/docs/api-reference/images/createEdit>.

## CP note

No new config surface: the route dispatches on existing Model/ProviderKey resources (`provider: openai`), so no cp-admin.yaml change is needed. Docs (endpoint page + provider endpoint-coverage rows) tracked in AISIX-Cloud#1360's acceptance list.

## Tests

- **Unit (12)**: form reconstruction + alias swap (upstream never sees the alias), usage extraction, guardrail 422 with no pattern leak, non-OpenAI 400, `stream=true` 400, missing-model 400, non-multipart envelope 400, 401/403/404, upstream 5xx → 502.
- **e2e (2, real binary + etcd)**: multipart forward with file bytes/filename/fields intact and credential injected, response + usage relayed, `handler="images"` usage counter moves; provider-mismatch 400 with upstream never contacted.

Local runs: `cargo test -p aisix-proxy --lib` 992 passed; e2e file green against the debug binary; clippy/fmt clean.

## Summary by CodeRabbit

- **New Features**
  - Added image editing through the `/v1/images/edits` endpoint.
  - Supports multipart image uploads, model selection, authentication, authorization, quotas, prompt guardrails, and usage tracking.
  - Preserves uploaded images, masks, filenames, and form fields when forwarding requests.
  - Supports OpenAI providers and returns structured errors for unsupported or invalid requests.

- **Documentation**
  - Updated the feature list to include image editing.

- **Tests**
  - Added end-to-end coverage for forwarding, validation, provider restrictions, and usage reporting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ For a multi-replica cluster, point the gateway at etcd instead — `resources_fi
 Covered by 183 end-to-end scenario files (496 cases) that run against real gateway processes.
 
 - **OpenAI-compatible proxy** (`:3000`) — `chat/completions`, `completions`, `responses`,
-  `embeddings`, `rerank`, `images/generations`, `audio/{speech,transcriptions,translations}`,
+  `embeddings`, `rerank`, `images/{generations,edits}`, `audio/{speech,transcriptions,translations}`,
   `videos` (submit → poll → fetch), `files`, `batches`, `fine_tuning/jobs`, `realtime`,
   `GET /v1/models`, plus a root-level `/passthrough/:provider/*` escape hatch. Native SSE streaming,
   tool/function calling, JSON mode, vision/multimodal input, and reasoning-content support.
--- a/crates/aisix-proxy/src/images.rs
+++ b/crates/aisix-proxy/src/images.rs
@@ -111,6 +111,7 @@ pub async fn image_generations(
             // for completions / responses / rerank.
             let status = success.response.status().as_u16();
             emit_access_log(
+                "/v1/images/generations",
                 &model_name,
                 &success.provider,
                 &api_key_id,
@@ -151,6 +152,7 @@ pub async fn image_generations(
                     &state,
                     &snapshot,
                     &pk,
+                    "/v1/images/generations",
                     &request_id,
                     &success.model_id,
                     &model_name,
@@ -174,6 +176,7 @@ pub async fn image_generations(
             let status = err.status().as_u16();
             let elapsed = started.elapsed();
             emit_access_log(
+                "/v1/images/generations",
                 &model_name,
                 "unknown",
                 &api_key_id,
@@ -444,7 +447,7 @@ async fn dispatch(
 /// `usage: {input_tokens, output_tokens, total_tokens, ...}`; dall-e-2/3
 /// return no `usage` block → `None`. Wire shape:
 /// <https://platform.openai.com/docs/api-reference/images/object>
-fn extract_token_usage(body: &Value) -> Option<(u32, u32)> {
+pub(crate) fn extract_token_usage(body: &Value) -> Option<(u32, u32)> {
     let usage = body.get("usage")?;
     let input = usage.get("input_tokens").and_then(Value::as_u64)? as u32;
     let output = usage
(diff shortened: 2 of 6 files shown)


moonming:
**Independent audit disposition** (cold-context agent, verdict MERGEABLE-WITH-FIXES → fixes landed in 2b8c295):

| Finding | Disposition |
|---|---|
| HIGH-1 — `/v1/images/edits` missing from the LLM metrics tier (tokens booked into `aisix_llm_*_tokens_total` with no matching rows in `aisix_llm_requests_total`) | **Fixed**: registered in `LLM_ENDPOINTS`, the `ROUTES` guard, and the tier-split test |
| MEDIUM-1 — error-path UsageEvent dropped the requested model | **Fixed**: reads the attribution cell (populated past model resolution) |
| MEDIUM-2 — repeated-`image`/`mask` rebuild fidelity untested | **Fixed**: unit test pins two image parts + mask (bytes + filenames), e2e switched to `form.append` + mask with per-marker assertions |
| MEDIUM-3 — README endpoint list stale | **Fixed**: `images/{generations,edits}` |
| LOW-1 — `stream=true` reject placed before model resolution | **Fixed**: moved after resolve/ACL/IP gates — unknown model now 404s first, attribution real |
| LOW-2 — non-UTF-8 `prompt` skips guardrail scan but forwards | **Filed as #1016** — identical shape ships on both audio routes; fixing it here alone would leave the family inconsistent, so it lands as one family-wide PR |
| LOW-3 — `mime_str` parse failure silently drops the part content-type | **Justified**: mirrors the audio path byte-for-byte; diverging here would split the family convention. Covered by #1016's family pass if we harden it |
| LOW-4 — mixed explicit-commit / Drop-release reservation style | **Justified**: audit itself verified the two are functionally identical in both limiter stores; cosmetic |

Local re-run after fixes: 993 lib tests green, e2e file green against the rebuilt binary, clippy/fmt clean.

moonming:
**Bot-review triage** (Copilot + CodeRabbit on the first two commits; fixes in the commit just pushed):

| Finding | Disposition |
|---|---|
| Copilot — `record_success`/`mark_healthy` ran before the JSON-decode check | **Fixed**: marks moved after the parse; a 2xx-with-garbage answer no longer records healthy on a request the caller sees fail. (The decode failure still doesn't feed cooldown — it sits outside the retry unit, same as the audio family; failover semantics unchanged.) |
| Copilot — capture treated any part with a content-type as binary, so a `text/plain` prompt got hashed | **Fixed**: binary slots are name-based (`image`/`mask`), matching audio's name-based `file` rule; invalid-UTF-8 fallback unchanged |
| CodeRabbit (major) — e2e readiness gates exercised the behavior under test inside a catch-all | **Fixed**: both gates now probe `GET /v1/models` with the caller key seeded last, per the harness AGENTS.md |
| CodeRabbit (minor) — `/v1/images/edits` missing from `LLM_ENDPOINTS` + `ROUTES` | Already fixed in 2b8c295 (independent-audit HIGH-1) |

Local: 13 unit + 2 e2e green on the rebuilt binary, clippy/fmt clean.

PR #1018, opened 2026-08-21
fix(dispatch): openai keys with no api_base fall back to the default base on every route

### LINKED ISSUE #1017: openai provider: default api_base fallback is inconsistent across routes — chat/generations/videos fall back, audio and images/edits 400
`dispatch::resolve_base_url` errors when a provider key has no `api_base` ("cp-api must populate api_base for every catalog vendor"). That is correct for managed mode, where cp-api always writes the field — but for the open-source declarative workflow an OpenAI provider key without `api_base` behaves inconsistently by route:

| Route family | Unset `api_base` on an `openai` provider key |
| --- | --- |
| `/v1/chat/completions`, `/v1/images/generations`, and the other bridge-dispatched routes | Falls back to the bridge's built-in `https://api.openai.com` default |
| `/v1/videos` | Falls back to `OPENAI_DEFAULT_BASE` explicitly (`videos.rs` ~1087) |
| `/v1/audio/transcriptions` / `translations` / `speech`, `/v1/images/edits` | `400 invalid_request_error` from `resolve_base_url` |

So the same declarative config works on chat and breaks on audio/edits, and the OpenAI provider docs page's claim that an alias works "with `api_base` unset exactly like the chat alias" is only true for some routes. The docs now state the per-route requirement explicitly (image-editing endpoint page), but the product behavior should converge instead.

**Proposal:** for provider keys whose `provider` is `openai`, make the direct-HTTP call sites fall back to the same default base the bridge and the videos route use — either inside `resolve_base_url` (openai-only arm) or at each call site via the videos pattern. Sweep every `resolve_base_url` caller and pin the behavior with one e2e per route family (no-`api_base` OpenAI key succeeds).

Surfaced by the docs audit of the `/v1/images/edits` pages (api7/docs PR 2162 review).

moonming:
Closes #1017.

## Problem

An OpenAI provider key configured without `api_base` behaved inconsistently by route:

| Route family | Unset `api_base` on an `openai` key |
| --- | --- |
| Bridge-dispatched (chat, generations, …) | Falls back to the bridge's built-in `https://api.openai.com/v1` |
| `/v1/videos` | Falls back explicitly (`videos.rs`) |
| Direct-HTTP via `resolve_base_url` — audio ×3, `/v1/images/edits`, jobs, realtime | `400 invalid_request_error` |

Same declarative config: chat works, audio breaks. The OpenAI provider docs also promise an alias works with `api_base` unset "exactly like the chat alias".

## Fix

One chokepoint: `resolve_base_url` falls back to `OPENAI_DEFAULT_BASE` when `api_base` is unset/blank **and** the key's vendor is exactly `openai` (trim + case-insensitive). Every `resolve_base_url` caller converges at once.

Deliberately strict on the vendor string, mirroring the openai family bridge's refusal: an empty or OpenAI-compatible vendor (`deepseek`, `openrouter`, …) still 400s — this resolver also serves non-OpenAI-family paths (messages/count_tokens), and falling back would send another vendor's credential to api.openai.com. `videos.rs` keeps its equivalent local fallback untouched (keyed on the model's provider; behavior identical).

## Tests

- Unit: openai fallback; trim/case tolerance + whitespace-only `api_base`; OpenAI-compatible vendor still errors; legacy no-provider row still errors. 996 lib tests green.
- e2e: the 400-contract spec for family-adapter vendors (`invalid-upstream-config-400-e2e`) passes unchanged, plus the audio/images/edits specs. The fallback itself cannot be pinned e2e without dialing the real api.openai.com — the resolver unit tests are the contract (same coverage stance as the videos-route fallback).

## Docs note

The in-review docs PRs (api7/docs#2162 / docs.apiseven.com#487) currently document the 400; once this merges they will be updated to the fallback wording.

## Summary by CodeRabbit

- **Bug Fixes**
  - Requests now consistently use the correct provider-specific API endpoint, even after provider settings change.
  - Prevented stale endpoint reuse across audio, speech, token counting, image editing, messaging, and response requests.
  - OpenAI requests use OpenAI’s default API endpoint when no custom base URL is provided.
  - Blank or whitespace-only base URLs are handled correctly.
  - Vendor matching for OpenAI is case-insensitive.
  - Other vendors continue to require an explicitly configured API endpoint.
  - Missing endpoint errors now include clearer remediation guidance.

### DIFF
--- a/crates/aisix-proxy/src/audio.rs
+++ b/crates/aisix-proxy/src/audio.rs
@@ -990,6 +990,11 @@ async fn multipart_dispatch(
         url_cache_key,
         &[
             pk_entry.value.api_base.as_deref().unwrap_or(""),
+            // The resolver's output depends on the vendor since #1017
+            // (openai default-base fallback) — a fingerprint without it
+            // would keep serving a stale api.openai.com URL after the
+            // row is repurposed to another vendor.
+            pk_entry.value.provider.as_str(),
             upstream_path,
         ],
         || {
@@ -1664,7 +1669,11 @@ async fn speech_dispatch(
     let speech_url = aisix_gateway::url_cache::cached_endpoint_url(
         &pk_entry.id,
         "proxy/audio/speech",
-        &[pk_entry.value.api_base.as_deref().unwrap_or("")],
+        &[
+            pk_entry.value.api_base.as_deref().unwrap_or(""),
+            // Vendor in the fingerprint — see multipart_dispatch (#1017).
+            pk_entry.value.provider.as_str(),
+        ],
         || {
             let base = crate::dispatch::resolve_base_url(&pk_entry.value)?;
             Ok::<_, crate::error::ProxyError>(crate::dispatch::build_openai_url(
--- a/crates/aisix-proxy/src/audio.rs
+++ b/crates/aisix-proxy/src/audio.rs
@@ -988,15 +988,13 @@ async fn multipart_dispatch(
     let url = aisix_gateway::url_cache::cached_endpoint_url(
         &pk_entry.id,
         url_cache_key,
-        &[
-            pk_entry.value.api_base.as_deref().unwrap_or(""),
-            // The resolver's output depends on the vendor since #1017
-            // (openai default-base fallback) — a fingerprint without it
-            // would keep serving a stale api.openai.com URL after the
-            // row is repurposed to another vendor.
-            pk_entry.value.provider.as_str(),
-            upstream_path,
-        ],
+        // Every resolve_base_url input, via the shared constructor
+        // (#1017: the resolved URL depends on the vendor too), plus the
+        // per-call path.
+        &{
+            let [base, vendor] = crate::dispatch::pk_url_fingerprint(&pk_entry.value);
+            [base, vendor, upstream_path]
+        },
         || {
             let base = crate::dispatch::resolve_base_url(&pk_entry.value)?;
             Ok::<_, crate::error::ProxyError>(crate::dispatch::build_openai_url(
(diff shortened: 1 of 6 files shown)


moonming:
**Independent audit disposition** (cold-context agent, verdict MERGEABLE-WITH-FIXES → fixes in 7c5142f):

| Finding | Disposition |
|---|---|
| HIGH — cached-URL fingerprints omitted the vendor: a row cached with `provider: openai` + no `api_base` kept serving the api.openai.com URL after being repurposed to another vendor (stale-route credential egress until restart) | **Fixed**: vendor added to all six resolver-fed fingerprints (audio ×2, count_tokens, images_edits, messages, responses — rerank already carried it), plus a cache-level regression test simulating the repurpose with the exact fingerprint shape |
| MEDIUM — comment claimed the strict empty-vendor refusal "mirrors the family bridge"; the bridge actually falls back for the legacy empty-provider shape, leaving that shape route-dependent | **Fixed**: comment corrected to state the divergence accurately; residue filed as #1019 with three convergence options |
| LOW — misdeclared `provider: openai` on an anthropic-adapter path now dispatches and gets the upstream 401 instead of a gateway 400 | **Documented** in the resolver's comment (by declaration the secret is an OpenAI credential — no cross-vendor flow) |
| LOW — 400 body named control-plane internals ("cp-api") | **Fixed**: remediation detail moved to `tracing::error`, customer-visible body generic — matching the family bridge's posture |

Local after fixes: 997 lib tests, clippy/fmt clean, 4 e2e files (11 tests) green incl. the anthropic-upstream spec covering the touched messages fingerprint.

PR #1020, opened 2026-08-21
fix(guardrails): reject non-UTF-8 prompt multipart fields on every surface

### LINKED ISSUE #1016: guardrails: a non-UTF-8 `prompt` multipart field skips the input scan and mask but still forwards verbatim
On the multipart surfaces — `/v1/audio/transcriptions`, `/v1/audio/translations` (`audio.rs`, prompt scan ~L884 / redact ~L915), and `/v1/images/edits` (`images_edits.rs`, same shape) — the input-guardrail scan and the mask-action redaction both read the `prompt` field through `std::str::from_utf8(...).ok()`. A `prompt` part whose bytes are not valid UTF-8 is silently skipped by both passes, yet the rebuilt form still forwards those bytes to the provider verbatim.

That is a narrow but real guardrail bypass: a caller can smuggle text past a keyword/DLP input guardrail by encoding the prompt in any non-UTF-8 encoding the upstream model tolerates (or by prefixing a single invalid byte).

**Proposed fix (family-wide, one PR):** reject a non-UTF-8 `prompt` field with 400 `invalid_request_error` on all three routes before the guardrail pass. The upstream contract types `prompt` as text, so a compliant caller loses nothing.

Surfaced by the independent audit of #1015 (LOW, deliberate follow-up rather than a rider: the fix belongs to the audio family as much as the new route).

moonming:
Closes #1016.

## Problem

On the three multipart surfaces — `/v1/audio/transcriptions`, `/v1/audio/translations`, `/v1/images/edits` — the input-guardrail scan and mask-action passes read the `prompt` field through `std::str::from_utf8(...).ok()` and **skip** a field that fails to decode, while the rebuilt form forwards the original bytes verbatim. One invalid prefix byte (or a non-UTF-8 encoding) smuggles text past a keyword/DLP guardrail. On the edits route the input scan is the only control — the response is an image, no output hook.

## Fix

One chokepoint, `dispatch::require_utf8_prompt_fields`: any `prompt` part that is not valid UTF-8 answers `400 invalid_request_error`, called by both multipart dispatches **after model resolution** (unknown models keep answering 404) and **before the guardrail pass and quota reservation**.

Two deliberate choices:

- **Unconditional** — not gated on a guardrail chain being configured. A request's validity must not flip when a guardrail is attached later, and the JSON endpoints already behave exactly this way (serde rejects invalid-UTF-8 bodies with 400), so this converges the multipart surfaces to the rest of the proxy.
- **Reject, not lossy-decode.** Scanning a lossy-decoded copy while forwarding the original bytes leaves the same gap (a keyword in a foreign encoding never appears in the lossy text); forwarding the lossy copy mutates caller content. Validate-then-match is the standard boundary posture in content-inspection systems (the classic WAF anti-smuggling rule); of the comparators we checked, the Python-framework camp gets scan==forward parity incidentally from framework decoding but never rejects, and the plugin camp simply has no text guardrails on these route types at all — fail-closed is strictly stronger than both.

The upstream contracts type `prompt` as text, so a compliant caller loses nothing.

## Tests

- **Unit (5)**: the smuggle shape itself (0xFF + blocked keyword with a guardrail configured, upstream `expect(0)`) on transcriptions and edits; the unconditional reject with no guardrail on translations; 404 precedence for an unknown model; the pure helper (binary `image`/`mask` parts pass untouched). 1002 lib tests green.
- **e2e (3, real binary + etcd)**: one journey per wired route — raw-byte multipart with an invalid `prompt`, asserting 400 + `invalid_request_error` + the UTF-8 message + zero upstream hits.

No CP work: no config surface changes; behavior is a fail-closed validation tightening on malformed input only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Multipart audio and image-edit requests now return `400 Bad Request` when the `prompt` contains invalid UTF-8 data.
  * Invalid prompts are rejected before safety checks and upstream processing, preventing malformed requests from being forwarded.
  * Other binary multipart fields continue to be accepted normally.
  * Added coverage across transcription, translation, and image-edit requests.

### DIFF
--- a/crates/aisix-proxy/src/audio.rs
+++ b/crates/aisix-proxy/src/audio.rs
@@ -845,6 +845,11 @@ async fn multipart_dispatch(
     // Client-IP allowlist gate (#557): reject before quota / upstream.
     crate::dispatch::check_ip_access(&model_entry.value, &client_ctx.source_ip)?;
 
+    // #1016: a `prompt` part that is not valid UTF-8 would be skipped by
+    // the guardrail scan below yet forwarded verbatim — reject it before
+    // the guardrail pass instead of leaving that asymmetry.
+    crate::dispatch::require_utf8_prompt_fields(&fields)?;
+
     // #696: transcriptions/translations run the guardrail chain too. The
     // audio bytes aren't scannable text, but the optional `prompt` form
     // field IS caller text forwarded verbatim to the provider — scan it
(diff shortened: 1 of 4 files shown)


moonming:
**Independent audit disposition** (cold-context agent, verdict MERGEABLE; all 8 contract claims verified incl. the completeness sweep — `prompt` is the only scanned-but-skippable field on any multipart surface):

| Finding | Disposition |
|---|---|
| LOW — `/v1/files` batch blob is scanned lossy but forwarded verbatim (same asymmetry class, pre-existing, out of this PR's scope) | **Filed as #1022** with the purpose-aware fix (JSONL is contractually UTF-8) |
| LOW — audio side lacked a 404-precedence test | **Fixed** in the follow-up commit: `non_utf8_prompt_unknown_model_still_404` mirrored onto `/v1/audio/transcriptions` |
| LOW — before-quota placement verified by code order only | Accepted as-is per the audit's own assessment (a first-429 harness is disproportionate) |

1003 lib tests green after the rider; merging on CI green.

moonming:
CodeRabbit triage: the readiness-gate body IS consumed — `ProxyClient.listModels` goes through `json()`, which reads `res.body.text()` before returning `{status, body}` (`tests/e2e/src/harness/proxy.ts`). The comment assumed a raw `fetch`; no change needed. (Same pattern already merged in `images-edits-e2e.test.ts`.)

PR #1025, opened 2026-08-22
feat(guardrails): hand redaction rules the field name of a structured value

membphis:
## Problem

The mask write-back channel rewrites JSON string values **leaf by leaf**: each
leaf is decoded and handed to the redaction rules on its own. When the
sensitive value is a real JSON field — the shape a tool returns as
`structuredContent: {"version": "12.1"}` or takes as
`arguments: {"version": "12.1"}` — the rule receives the bare string `12.1`.
That is indistinguishable from `12.345` or `10.2.255.1`, so an operator rule
anchored on the field name can never match and the value ships unmasked.

The failure is silent, and it is inconsistent in a way an operator cannot
predict: the *same* rule fires when a tool happens to embed the document
inside a prose leaf (`"log": "config {\"version\": \"12.1\"} ok"`), because
there the key really is part of the text the rule sees.

## Change

Offer each object member to the redactor as its JSON spelling — `"key":"value"`,
the same compact form `Value::to_string()` already hands the `/mcp` input
**scan**, so one operator rule covers both the block decision and the mask.

Two passes, composed the way a chain composes its members (`fold_redactions`),
never raced against each other:

1. the **bare leaf** — exactly the pre-key-context rewrite. Always kept, so a
   rule anchored to the whole leaf (`^…$`, which can only ever match the bare
   form) keeps firing regardless of what pass 2 does.
2. the **member's JSON spelling** around pass 1's output.

Pass 2 is kept only when the synthesized prefix (`"key":"`) and the closing
quote come back byte-for-byte **and** the value actually changed — otherwise
the rule rewrote the wrapper rather than the data, or re-matched its own mask
token, and pass 1 stands. Because pass 2 can be discarded it runs through a new
unaudited channel (`Guardrail::redact_{input,output}_text_unaudited`) and the
kept rewrite is replayed through the audited path: an entry in
`guardrail_enforced_hits` must describe a mask the client really saw.

The wrapped pass is skipped for an empty value, and for a key or value carrying
`"` or `\` — their decoded bytes are not JSON string content, so the synthetic
would not be the document's own spelling and a `"([^"]*)"` rule would frame on
the embedded quote and mask only the head while reporting a full hit.

Applied where the object keys are the **tool's own schema**:

| surface | key context | why |
|---|---|---|
| `params.arguments.**` (`/mcp` request) | yes | the tool's input schema |
| `result.structuredContent.**` (`/mcp` response) | yes | the tool's output schema |
| `result.content[].text` / `description` / `title` / `resource.text` | no | MCP envelope names |
| `function.arguments` / `tool_use.input` (LLM families) | yes | the tool's input schema |

The envelope exclusion is deliberate: the scan path decodes content blocks
precisely so `content`/`type`/`text` cannot trip a rule, and wrapping a prose
block back up as `"text": "…"` would reintroduce that false positive. A unit
test pins it.

Array elements and the document root have no key and keep the previous
behaviour. JSON **number** leaves (`"version": 12.1`) remain out of scope by
design — rewriting one to a mask token would corrupt the document; that
decision is still open with the customer and is not touched here.

### Handler-family sweep

The same gap existed on the LLM families' JSON tool-argument walker
(`redact_value_strings`, reached from chat `function.arguments` and Anthropic
`tool_use.input`), so it is covered in this PR rather than deferred.

Doing so surfaced a real constraint worth pinning: the remote segment-moderation bridge (a remote guardrail whose PII action is anonymize) pairs its
collect and apply walks **by slot index**, so a walker must offer every text
exactly once and exactly as it stands. Key context would both change the text
sent to the provider and desynchronize the slots. The two probes now declare
that with `Guardrail::redacts_positionally()` (defaulted `false`) and keep the
bare-leaf walk; the existing positional test caught this before the guard
existed, and a new test pins the invariant directly. Remote entity detectors
take no field-name-anchored rules, so nothing is lost on that path.

## Verification

Real DP + etcd + a real MCP upstream (official SDK server) + the SLS mock, two
gateway instances sharing one upstream so the unguarded one is the
byte-for-byte baseline:

- **response** — `structuredContent: {version, 版本, elapsed, port, log, cells}`
  → `version`/`版本` masked, **full-body byte-diff** against the baseline proves
  `elapsed`, `port`, `cells` and every other byte are unchanged; the body still
  parses;
- **request** — the same fields as `tools/call` arguments → the upstream
  receives `"version":"***"` / `"版本":"***"`, negatives verbatim, byte-diff
  against the baseline instance;
- **false-positive suppression** — `{"elapsed": "12.345"}` and
  `{"port": "10.2.255.1"}` are zero-hit; that discrimination is the point of
  handing the rule the key;
- Chinese field name (`版本`) covered on both directions;
- unit: out-of-bounds rewrite falls back to the bare leaf; a wrapped hit does
  not steal a `^…$` leaf-anchored rule's turn; an idempotent second match is
  not counted twice; a discarded pass files no enforced hit; a value carrying
  `"` and an empty value skip the wrapped pass; the compact spelling matches
  the scan path; envelope names are not offered; the positional probe sees each
  leaf once verbatim.

`cargo test --workspace`, `cargo clippy --workspace --all-targets`, and the
guardrail/MCP e2e suites (20 files) pass.

## Independent audit

Audited cold by a separate agent per the repo's pre-merge rule. Findings and
what was done:

- **HIGH — a wrapped hit stole the leaf-anchored rule's turn.** With a chain
  carrying both an unanchored rule and a `^\d{4}$` rule, `"1234"` came back
  `Z234`: three of four secret digits shipped where the leaf rule had masked
  all four. Reproduced with a test, then fixed by the bare-first composition
  above. The test is kept as a regression.
- **MEDIUM — a discarded pass filed a phantom `masked` hit.** A key-anchored
  rule without a capture group swallows the wrapper, so the rewrite is dropped
  — but `fold_redactions` had already recorded an enforced hit, putting a mask
  on the usage event that no client ever saw. Fixed by the unaudited channel;
  verified the test fails without it.
- **MEDIUM — value containing `"` was masked only up to the embedded quote**
  while reporting a full hit. Fixed by the skip guard.
- **MEDIUM — spelling divergence** between the scan path (compact) and the mask
  path (spaced) meant a rule written without `\s*` blocked but never masked.
  Fixed by adopting the compact spelling; a unit test pins it.
- **MEDIUM — output-hook `action: block` is still key-blind** (`collect_string_leaves`
  drops keys). Filed as api7/aisix#1026 rather than folded in — it changes the
  block path and deserves its own review.
- **MEDIUM — monitor mode previews no would-mask hits** for field-anchored
  rules, because the probe runs over the scan text. Filed as api7/aisix#1027.
- **LOW — `redacts_positionally()` was not forwarded** by the chain or the three
  decorators. Added; harmless today (probes are passed bare) but the failure
  mode would be silent slot desync.
- **LOW — the relaxed e2e assertion** was legitimate but weaker. Strengthened
  with a count bound and a union assertion over both detectors.

Confirmed correct by the audit and left alone: UTF-8/char-boundary safety of
the value-span slice, JSON document integrity (the splice re-encodes), no new
remote exposure (the only concrete redactors are in-process), the segment
-bridge protection end to end, the path predicates, and the handler-family
sweep.

## Not in scope

- JSON number leaves — pending the customer decision.
- The semantic layer (AISIX-Cloud#1363).
- The LLM handler families' enforced-hit drain (aisix#1024) — separate work.
- Field names on the output-hook block decision (api7/aisix#1026) and monitor
  -mode preview of field-anchored rules (api7/aisix#1027) — both filed.
- A `api7/docs` page is owed: the synthesized `"key":"value"` spelling and the
  capture-group requirement are what an operator needs to write a working rule.
- No control-plane counterpart: this changes how existing `custom_patterns`
  rules are evaluated, not the config surface. No new field, no new enum.

## Summary by CodeRabbit

* **New Features**
  * Added field-aware redaction for structured JSON values, including nested fields and multilingual keys.
  * MCP tool arguments and structured responses now provide object field names as redaction context.
  * Added support for configuring structured fields in MCP test responses.

* **Bug Fixes**
  * Prevented discarded masking attempts and duplicate matches from inflating enforcement audits.
  * Improved handling of positional redactors, escaped values, empty values, and anchored rules.
  * Preserved unrelated values, array elements, root values, and original wire formatting.

* **Tests**
  * Expanded end-to-end coverage for structured-field masking in English and Chinese.

### DIFF
--- a/crates/aisix-guardrails/src/build.rs
+++ b/crates/aisix-guardrails/src/build.rs
@@ -728,6 +728,21 @@ impl Guardrail for MonitorGuardrail {
         self.observe_redaction("output", self.inner.redact_output_text(text));
         None
     }
+
+    // A speculative pass records nothing — not the enforcement audit, and
+    // not the monitor observation either: the caller may throw the rewrite
+    // away, and a would_mask hit for it would be as false as an enforced one.
+    fn redact_input_text_unaudited(&self, _text: &str) -> Option<Redaction> {
+        None
+    }
+
+    fn redact_output_text_unaudited(&self, _text: &str) -> Option<Redaction> {
+        None
+    }
+
+    fn redacts_positionally(&self) -> bool {
+        self.inner.redacts_positionally()
+    }
 }
 
 /// `mandatory: true` decorator. A remote guardrail that can't reach its
@@ -813,6 +828,18 @@ impl Guardrail for MandatoryGuardrail {
 
     fn redact_output_text(&self, text: &str) -> Option<Redaction> {
         self.inner.redact_output_text(text)
+    }
+
+    fn redact_input_text_unaudited(&self, text: &str) -> Option<Redaction> {
+        self.inner.redact_input_text_unaudited(text)
+    }
+
+    fn redact_output_text_unaudited(&self, text: &str) -> Option<Redaction> {
+        self.inner.redact_output_text_unaudited(text)
+    }
+
+    fn redacts_positionally(&self) -> bool {
+        self.inner.redacts_positionally()
     }
 }
 
@@ -927,6 +954,18 @@ impl Guardrail for LiveGuardrailChain {
 
     fn redact_output_text(&self, text: &str) -> Option<Redaction> {
         self.current().redact_output_text(text)
+    }
+
+    fn redact_input_text_unaudited(&self, text: &str) -> Option<Redaction> {
+        self.current().redact_input_text_unaudited(text)
+    }
+
+    fn redact_output_text_unaudited(&self, text: &str) -> Option<Redaction> {
+        self.current().redact_output_text_unaudited(text)
+    }
+
+    fn redacts_positionally(&self) -> bool {
+        self.current().redacts_positionally()
     }
 }
 
--- a/crates/aisix-guardrails/src/chain.rs
+++ b/crates/aisix-guardrails/src/chain.rs
@@ -698,6 +698,33 @@ impl Guardrail for GuardrailChain {
             self.audit.as_deref(),
         )
     }
+
+    // The speculative pair: same fold, no audit sink. A caller that may
+    // throw the rewrite away uses these, then replays the audited call on
+    // the rewrite it actually keeps.
+    fn redact_input_text_unaudited(&self, text: &str) -> Option<Redaction> {
+        fold_redactions(
+            text,
+            self.members.iter().filter(|m| m.guardrail.redacts_input()),
+            true,
+            None,
+        )
+    }
+
+    fn redact_output_text_unaudited(&self, text: &str) -> Option<Redaction> {
+        fold_redactions(
+            text,
+            self.members.iter().filter(|m| m.guardrail.redacts_output()),
+            false,
+            None,
+        )
+    }
+
+    fn redacts_positionally(&self) -> bool {
+        self.members
+            .iter()
+            .any(|m| m.guardrail.redacts_positionally())
+    }
 }
 
 /// Fold the texts through each segment-moderating member. Mirrors the
--- a/crates/aisix-guardrails/src/lib.rs
+++ b/crates/aisix-guardrails/src/lib.rs
@@ -532,6 +532,18 @@ pub trait Guardrail: Send + Sync + 'static {
         None
     }
 
+    /// `true` when this redactor consumes the offered texts POSITIONALLY —
+    /// it pairs slot *i* of one walk over a body with slot *i* of the next,
+    /// rather than deciding each text on its own content. A wire walker must
+    /// then offer every text exactly once and exactly as it stands: no repeat
+    /// call, no context synthesized around a value, or the two walks stop
+    /// lining up and masks land on the wrong fields. Only the collect/apply
+    /// probes of the segment-moderation bridge set this; a real guardrail
+    /// redacts each text independently and leaves it `false`.
+    fn redacts_positionally(&self) -> bool {
+        false
+    }
+
     // --- remote segment moderation (#932 bedrock follow-up) ---------------
     //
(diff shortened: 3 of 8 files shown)


membphis:
Closing: the requirement this was built for no longer exists.

The driving case was a sensitive value that IS a JSON field, where the field
NAME is the only thing identifying it (`{"version": "12.1"}` — the value alone
is indistinguishable from any other dotted number). Confirmation from the
customer is that their payloads carry the discriminator INSIDE the value (the
software name and its version in one string), and that a key+value combination
is not needed now or later.

That shape is already covered, and has been since the write-back channel
landed: the rules receive the whole leaf, so a rule anchored on the software
name or on a `version:`-style keyword matches directly. The e2e on `main`
already pins it in both directions, in English and Chinese, with same-shaped
negatives asserted byte-identical:

- a content text block, `阶段汇总 版本：2022.4 用时 12.345s`
- an embedded resource's `resource.text`, `Compile OK version: 12.1`
- a `structuredContent` string leaf, `config {"version": "12.1"} ok`
- request-side `arguments`, `build version: 12.1 … 工具版本：2022.4`
- negatives `Elapsed: 12.345s`, `0.13um`, `10.2.255.1` — untouched

With that requirement gone, this change is a behaviour change to every already
-deployed pattern across `/mcp` and the LLM tool-argument walker, in exchange
for a case nobody needs. The pre-merge audit also showed the approach has a
sharp edge worth recording for anyone who revisits it: offering the rules a
wrapped `"key":"value"` gives any unanchored rule the chance to consume the
turn of a rule anchored to the whole leaf (`^…$`), which can only ever match
the bare form — a `^\d{4}$` PIN mask silently degraded to a one-character mask
until the two passes were composed rather than raced.

The branch stays pushed. Two findings from the audit are unrelated to the
key+value idea and are tracked on their own: #1026 and #1027.

PR #1045, opened 2026-08-24
feat(obs): log the downstream connection identity a rollout needs

jarvis9443:
Four log fields that a rolling-update investigation needed and could not get. Observability only — no behaviour changes, no new config, no wire or schema change.

## The gaps

Reconstructing what happened to a request during a rolling update means joining three records, and nothing connected them:

- **No downstream peer address.** The gateway's own `request_id` is minted here, so nothing in front of it knows that value. Behind a layer-4 load balancer with the gateway on the host network, the accepted socket's remote address *with its port* is the only field that lines a gateway log line up with the fronting proxy's record of the same connection — every request otherwise arrives from the same handful of hosts.
- **No arrival line.** Everything the gateway writes about a request is written when the request *ends*. A request still running when the platform's grace period expires never gets there, so it leaves no trace at all — which is exactly the request an investigation is looking for.
- **No connection count on the drain, and no line when one is accepted during it.** Without them the in-flight count alone cannot distinguish work the gateway had already taken on from traffic a balancer is still routing at it.
- **The fronting proxy's `X-Request-Id` was discarded.** It is not adopted as the gateway's id by default, and it was not recorded either, so a search by an id from the ingress's logs returned nothing.

## What is added

**`peer` on the request span** — the accepted socket's remote address including its port. Deliberately not the resolved `source_ip` that `ClientContext` already carries: that answers "who is the caller", follows `proxy.real_ip`, and has no port. This one answers "which TCP connection".

**`downstream_request_id` on the request span**, read from `x-request-id` and screened by the same rule an adopted id passes, so nothing unsafe is written into a log line verbatim. Recorded, never adopted: which header may *become* this gateway's id stays governed by `proxy.request_id.accept_headers`. The ingress's id and the gateway's id are two fields that cannot be confused for one another.

Both ride the existing request span, so they reach the access log, the per-attempt `provider call completed` line, and every diagnostic in between — the same mechanism `request_id` already uses, rather than a second one threaded through the emit sites.

**A drain-time arrival line** (`request arrived while draining`, with method and path; identity comes from the span). Gated on the shutdown flag, so no line is added to steady-state logging. It skips `/livez` and `/readyz`: those are served on this same listener and the platform probes them every few seconds for the whole grace period, so including them would bury the handful of real arrivals — and a probe reaches its completion line regardless, which is the reason the arrival line exists for anything else.

**`open_connections`** on the signal-time line and on the drain heartbeat, plus an INFO when a connection is accepted after the signal. The count is raised and lowered by an RAII guard on the accepted stream and saturates at zero, so an unpaired decrement cannot wrap it. Only the proxy listener is counted; the admin and metrics listeners pass no drain state. At accept time there is no path to tell a probe from a client, so the count includes health probes and the accept line does not claim to know who connected — that judgement belongs to the arrival line, which has the path. The accept line covers what the arrival line cannot: a connection opened and never used.

Both new span fields carry caller-controlled or caller-influenced text and are recorded as strings, so they render quoted and escaped rather than raw into a space-delimited line.

Two things deliberately not done: `open_connections` is log-only and gets no Prometheus gauge (this repo only ships a metric family together with an e2e that asserts it in a real scrape, which is more than this needs); and no `max_drain_secs` knob — the platform stays the only hard bound.

Also documents, on `retire_connection`, why it is structurally unable to retire a **streaming** connection: a streamed response returns from the handler at the first token, so its head is already on the wire and a stream whose head went out before the signal can never carry `Connection: close`, however long it runs afterwards. No change to that function — the note exists so the next reader does not conclude a header fix would cover it.

## Also: the grace-period guidance was wrong

`config.example.yaml` and `config.managed.yaml` both told operators to size `terminationGracePeriodSeconds` to cover their *longest request*. That underestimates the real bound, and a deployment sized by it gets pods killed mid-request.

The gateway does not close a client's connection itself — it answers `Connection: close` and lets the client retire it, which is deliberate: a server-side close races the client dispatching onto the same connection. But that header rides the response head, and a streamed response commits its head at the first token, so a stream already running at SIGTERM can never carry it, and HTTP/1.1 offers no other way to retire the connection once the head is on the wire. It returns to the client's pool unmarked and is reused exactly once more; that successor's head IS generated during the drain, carries the header, and ends the chain.

The bound is therefore **two requests back to back**, each able to run to its own client timeout. Both files now say so and point at the Helm chart's "Termination and draining" section for the arithmetic, rather than duplicating a default here.

The same trap is now named on the access log itself: a streamed line's `latency_ms` is time-to-first-token, not how long the stream ran — the two differ by the length of the stream, and reading it as the end of the request is how a long-running stream gets mistaken for an idle connection.

## Tests

`tests/e2e/src/cases/drain-observability-e2e.test.ts` drives the real binary and asserts on its stderr, which is the only place these fields are consumed from. All five cases fail on `main` and pass here.

The assertions are built so none of them can pass for the wrong reason: the two negative cases (no downstream id sent; an unusable one sent) also assert `peer` is present on the same line, so neither passes merely because no connection fields are reported at all; and the drain case drives real health probes during the drain and fails if they produce arrival lines. Unit tests pin the counter's saturation and that the acceptor releases a connection's slot when the stream drops — every drain assertion is "at least one", so a leaked guard would otherwise keep them all green while the count climbed.

Note on volume: line *count* in steady state is unchanged, but `peer` rides the request span, so every line inside a request span grows by roughly the length of an `ip:port`.

The existing drain and request-id e2e suites pass unchanged, which is the check that behaviour did not move.

Ref api7/AISIX-Cloud#1394

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -364,8 +364,19 @@ downstream:
 # is left in flight. In-flight requests then drain with no deadline — an
 # inference call or an SSE stream may run for minutes — so the platform
 # (Kubernetes terminationGracePeriodSeconds, systemd TimeoutStopSec) is
-# the only hard bound. Make sure that bound is generous enough for your
-# longest request.
+# the only hard bound.
+#
+# Size that bound for TWO requests back to back, not one. The gateway
+# retires a client's connection by answering `Connection: close`, which
+# rides on the response head — and a streamed response commits its head
+# at the first token, so a stream already running at SIGTERM can never
+# carry it, however long it runs. That connection returns to the client's
+# pool unmarked and gets used once more; only THAT request's head is
+# generated during the drain, carries the header, and ends the chain. So
+# the tail is the stream still running at the signal plus the one request
+# its connection is reused for, each able to run to its own client
+# timeout. The aisix Helm chart's "Termination and draining" section
+# works a default out from that.
 shutdown:
   # Minimum seconds to keep accepting new connections after the signal,
   # while /readyz already reports 503.
--- a/config.managed.yaml
+++ b/config.managed.yaml
@@ -165,6 +165,11 @@ downstream:
 # in front has time to withdraw it, then waits for in-flight requests to
 # finish before closing the listener. Raise it above the detection latency
 # of your balancer's health check. In-flight drain itself is unbounded, so
-# terminationGracePeriodSeconds must cover your longest request.
+# terminationGracePeriodSeconds must cover two requests back to back: a
+# streamed response commits its head at the first token, so a stream
+# already running at SIGTERM never gets the `Connection: close` that
+# retires its connection, and the client reuses that connection once more
+# before the successor's head — generated during the drain — ends the
+# chain. See the aisix Helm chart's "Termination and draining" section.
 shutdown:
   min_drain_secs: 30
--- a/crates/aisix-obs/src/access_log.rs
+++ b/crates/aisix-obs/src/access_log.rs
@@ -15,6 +15,12 @@
 //!   produced nothing yet, so the token counts and `provider_request_id` are
 //!   necessarily absent, and `status` is the response-OPEN status: a stream
 //!   that later aborts, or whose consumer walks away, still logged `200`.
+//!   `latency` is time-to-first-token for the same reason, NOT how long the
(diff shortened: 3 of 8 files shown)


jarvis9443:
@coderabbitai review

Issue #591, opened 2026-06-11
Use readiness semantics or 503 for graceful shutdown health state


kayx23:
## Problem

The public `/livez` endpoint currently represents intentional graceful shutdown as `500 Internal Server Error`.

This is mechanically enough to fail Kubernetes HTTP probes, but the status is semantically misleading: graceful shutdown is an expected drain state, not an unexpected internal server failure. The process is still alive, but it should stop receiving new traffic.

## Current behavior

- `LivezState::mark_shutting_down()` flips the shared shutdown state.
- `livez_response()` treats that state as a failed liveness check and returns `500 Internal Server Error`.
- Both proxy and admin `/livez` routes use the same `livez_response()` implementation.
- Tests currently pin the 500 behavior for proxy, admin, and the shared health helper.
- The generated Admin OpenAPI also documents `/livez` shutdown failure as a 500 response.

Relevant code:

- `crates/aisix-proxy/src/health.rs`: `livez_response()` returns `StatusCode::INTERNAL_SERVER_ERROR` when shutdown is marked.
- `crates/aisix-proxy/src/lib.rs`: proxy `/livez` delegates to `livez_response()` and has a test expecting 500 while shutting down.
- `crates/aisix-admin/src/lib.rs`: admin `/livez` delegates to `livez_response()` and has a test expecting 500 while shutting down.
- `crates/aisix-server/src/main.rs`: signal handling marks the shared livez state as shutting down before sending the graceful-shutdown cancel signal.
- `crates/aisix-admin/src/openapi.rs`: `/livez` documents a 500 response for shutdown.

## Why this matters

Kubernetes distinguishes liveness from readiness:

- Liveness answers whether the process should be restarted.
- Readiness answers whether the instance should receive traffic.

For intentional drain or graceful shutdown, readiness failure is the more precise signal. If this state remains on `/livez`, `503 Service Unavailable` better communicates temporary unavailability or maintenance than `500 Internal Server Error`.

References:

- Kubernetes HTTP probes treat 200-399 as success and other status codes as failure: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
- Kubernetes distinguishes liveness probes from readiness probes: https://kubernetes.io/docs/concepts/workloads/pods/probes/
- RFC 9110 defines 503 Service Unavailable for temporary inability to handle requests due to overload or scheduled maintenance: https://datatracker.ietf.org/doc/html/rfc9110#section-15.6.4

## Suggested options

[...]

PR #1049, opened 2026-08-25
fix(health): a draining gateway stays live

jarvis9443:
`/livez` answered `503` once graceful shutdown began, and the shipped Helm chart wires `/livez` as the container's `livenessProbe` (`charts/aisix/templates/deployment.yaml`). Liveness decides whether to **restart** an instance, and a draining one is finishing the requests it already accepted — restarting it kills exactly those. Withdrawing traffic is `/readyz`'s job, and it already does it.

## This is the half of #591 that never landed

That issue spelled out the distinction and offered two options:

> 1. Add a `/readyz` endpoint for traffic eligibility … **Keep `/livez` focused on process liveness.**
> 2. If adding `/readyz` is too large for now, change the intentional shutdown response from 500 to 503

\#655 did the mechanics of option 1 (added `/readyz`) *and* option 2 (softened `/livez` from 500 to 503), but never removed the shutdown check from `/livez`. The comment it left behind reads:

```rust
// Graceful shutdown is an expected drain, not an internal error —
// 503 so Kubernetes stops routing without treating it as a crash loop (#591).
```

"so Kubernetes stops routing" is readiness reasoning on a liveness endpoint. Kubernetes does not route on liveness.

## Severity: latent, not live

Kubernetes stops probing liveness once a pod enters graceful termination, so a rolling update never acted on the answer. Verified rather than assumed — on kind (k8s 1.33.1), against the real `api7/aisix:0.10.0` image, with a `livenessProbe` on `/livez` at `periodSeconds: 1` / `failureThreshold: 1` (30x more aggressive than the chart default) and a 60s drain: `/livez` returned 503 throughout, `restartCount` stayed `0`, and no `Unhealthy` event followed `Killing`.

Where it does bite:

- The gateway also runs as a **single container under docker or systemd**, where a supervisor watching `/livez` restarts a healthy draining process mid-flight.
- Any monitor treating `/livez` as "is the process healthy" raised a false alarm on every rolling update.
- During a drain the two endpoints were indistinguishable, which defeats having both.

## What changed

- `livez_response` no longer consults the drain state, and no longer takes `LivezState` at all. Neither route handler takes its state extractor either. Not receiving the state is a stronger guarantee that the answer cannot depend on it than a comment saying so.
- The admin OpenAPI drops `/livez`'s `503` response and rewrites its description. Liveness now has no failure to document: a draining instance answers `200` like any other, and an instance that cannot answer does not reply at all.
- `/readyz` is untouched. The drain has to be visible somewhere, and readiness is where.

The chart needs no change — `/livez` for liveness and `/readyz` for readiness was always the right wiring; the endpoint's behaviour was what did not match.

## Behaviour change

`GET /livez` on both the proxy and admin listeners now returns `200` during a graceful drain instead of `503`. Anyone using `/livez` to detect draining should use `/readyz`, which has reported it since #655. Worth a release-note line.

## Tests

- `health-minimal-e2e` gains the real contract against the real binary: after SIGTERM, poll until `/readyz` reports 503 — which proves the drain has begun *and* the process is still serving — then assert `/livez` is `200` on both listeners. It fails against the old behaviour with `expected 503 to be 200`.
- That spec's app now spawns with `min_drain_secs: 5`. The harness default is `0`, so the process exited within milliseconds of SIGTERM — which is why the test it replaces passed: it counted a connection error as "unhealthy", so it never actually observed a 503.
- `livez_stays_ok_while_draining` pins both halves in one place: `/livez` 200, `/readyz` 503, same drained state.
- The proxy and admin route tests are inverted to assert `200`, and the OpenAPI test now asserts `/livez` documents no 503 while `/readyz` does.

Documentation is corrected in the paired api7/docs#2191 and api7/docs.apiseven.com#500, which also fix a line left stale by #1048 (HTTP/2 receives its GOAWAY when the drain starts, not when the listener closes).

## Summary by CodeRabbit

* **Bug Fixes**
  * Liveness endpoints now continue returning HTTP 200 with an `ok` response during graceful shutdown.
  * Readiness endpoints continue returning HTTP 503 while the service is draining or unavailable.
  * Updated health checks and API documentation to reflect the distinct liveness and readiness behavior.

### DIFF
--- a/crates/aisix-admin/src/lib.rs
+++ b/crates/aisix-admin/src/lib.rs
@@ -380,11 +380,12 @@ fn normalized_prometheus_path(path: &str) -> String {
     }
 }
 
+/// Takes no [`AdminState`], for the same reason the proxy listener's
+/// does not: see [`aisix_proxy::health::livez_response`].
 async fn livez(
-    axum::extract::State(state): axum::extract::State<AdminState>,
     axum::extract::Query(params): axum::extract::Query<std::collections::HashMap<String, String>>,
 ) -> Response {
-    aisix_proxy::health::livez_response(&state.livez_state, params.contains_key("verbose"))
+    aisix_proxy::health::livez_response(params.contains_key("verbose"))
 }
 
 async fn readyz(
@@ -1087,7 +1088,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn livez_returns_503_when_shutting_down() {
+    async fn livez_stays_200_when_shutting_down() {
         let state = build_state();
         state.livez_state.mark_shutting_down();
         let app = build_router(state);
@@ -1096,10 +1097,12 @@ mod tests {
             .body(Body::empty())
             .unwrap();
         let resp = run(app, req).await;
-        assert_eq!(resp.status(), StatusCode::SERVICE_UNAVAILABLE);
-        let bytes = to_bytes(resp.into_body(), 1024 * 1024).await.unwrap();
-        let text = std::str::from_utf8(&bytes).unwrap();
-        assert!(text.contains("livez check failed"));
+        assert_eq!(
+            resp.status(),
+            StatusCode::OK,
+            "the admin listener answers the same liveness question as the \
+             proxy one, and a draining process is not one to restart",
+        );
     }
 
     #[tokio::test]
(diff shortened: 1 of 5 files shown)


nic-6443:
Fixed — the comment cited the gateway's 30s default while this spec configures 5 in `beforeAll`.

Worth more than the comment, though: the poll deadline was 5000ms, i.e. exactly the window it was waiting inside. A drain that never started would have been reported as a timeout at the moment the window closed, which is an ambiguous failure. Dropped it to 3000ms so the poll gives up strictly inside the window and the assertion says what actually went wrong.

PR #1066, opened 2026-08-28
feat(telemetry): attribute usage events and their counters to the org member

jarvis9443:
The Logs question *"show me Alice's 429s in the last 24h"* had no answer. A usage event named the credential a request arrived on (`api_key_id`, plus a JWT subject) but never the person behind it, so answering it meant enumerating every key a member owns, querying each, and merging by hand — and a member who calls through both an API key and OIDC was still split across two identities that no single filter could join.

## What changes

**`user_id` on the UsageEvent wire contract**, stamped from the resolved `ApiKey.user_id` at request time.

A snapshot rather than something cp-api resolves from `api_key_id` at query time. Rebinding a key would otherwise re-attribute its whole history to the new owner, and deleting one would erase the attribution of everything it ever sent. The JWT path needs no separate handling: a token runs AS the key it resolves to, so one field covers both halves of the auth surface — which is precisely the case a query-time join on `api_key_id` gets wrong.

`apply_jwt_identity` becomes `apply_caller_identity` and takes the member alongside the JWT identity. The rename is the point: this family is one call per handler and a new one is easy to forget, so the compiler now refuses any site that has not been considered.

**`aisix_usage_events_emitted_total` gains two labels:**

- `user_id`, via `UsageEventLabels` — this is attribution, so it lands on `aisix_usage_event_drops_total` too and *"whose usage records were lost"* stays answerable. `UsageSink::try_emit` fills it from the event's own field, so the counter and the row cp-api persists cannot name different members.
- `status`, the raw code. `status_code` keeps its `2xx` / `4xx` / `5xx` bucketing, and the raw one is named `status` to match `aisix_proxy_requests_total`'s own raw-code label — deliberately not `http_status_code`, which contains `status_code` as a substring and would make every `status_code="…"` matcher ambiguous (the e2e prometheus guard caught exactly that).

```promql
sum(increase(aisix_usage_events_emitted_total{user_id="<member-id>", status="429"}[24h]))
```

OTLP spans carry the member as `aisix.user_id`, beside `aisix.api_key_id`.

## Compatibility

- **No breaking metric change.** `status_code` keeps its existing bucketed values, so dashboards and alerts written against `status_code="4xx"` keep working verbatim. The raw code functionally determines the family, so carrying both labels adds no series over carrying the raw code alone.
- **No new cardinality class.** `aisix_proxy_requests_total` already carries `user_id`, `user_name`, `api_key_id`, `team_id` and a raw `status` unconditionally; the usage family is strictly lower-dimensional than it, so this introduces no dimension the process was not already emitting.
- **The wire field is additive and optional** (`skip_serializing_if`), so an older cp-api ignores it and a key bound to no member sends nothing.

## Deliberate omissions

- **`user_name` is not duplicated onto the usage family.** It is 1:1 with `user_id` on the request family and joinable from there; reaching it here would mean widening `CallerIdentity` across the gateway crate for a display convenience.
- **The drop counter gains no status dimension.** `handler` / `status_code` / `http_status_code` / `inbound_protocol` are the emit counter's own arguments rather than attribution, and `emitted == delivered + dropped` is an invariant over the attribution dimensions — which is exactly why `user_id` *does* appear on both.

## Tests

E2E over a real `aisix` binary and a real Aliyun-SLS export (`usage-member-attribution-e2e.test.ts`):

- an owned key stamps its member;
- a JWT resolving to that member's **other** key stamps the same member;
- an unowned key stamps nobody, so a member filter cannot sweep up traffic that was never theirs;
- an upstream 429 is addressable as `{user_id, status="429"}` with the `4xx` family still present.

The existing `emits_and_drops_carry_the_same_attribution_labels` unit test now covers `user_id` automatically — it derives the attribution set by difference, so a future attribution label that reaches only one of the two counters still fails it.

The control-plane half (persisting the column, the member + exact-status Logs filters, the dashboard) follows in a paired AISIX-Cloud PR.

Fixes api7/AISIX-Cloud#1389

## Summary by CodeRabbit

* **New Features**
  * Usage metrics now include raw HTTP status codes and caller attribution.
  * Telemetry records the authenticated caller’s user ID when available.
  * Caller attribution is consistent across API, background, passthrough, and realtime requests.
* **Bug Fixes**
  * Corrected attribution for failed requests, including 4xx and 429 responses.
  * Unowned credentials no longer appear in member-specific metric series.
* **Tests**
  * Added end-to-end coverage for member attribution, status reporting, and unattributed traffic.

### DIFF
--- a/crates/aisix-obs/src/metrics.rs
+++ b/crates/aisix-obs/src/metrics.rs
@@ -260,11 +260,15 @@ pub const M_GUARDRAIL_LATENCY_SECONDS: &str = "aisix_guardrail_latency_seconds";
 ///   embeddings / responses / completions / rerank / audio /
 ///   images / messages). Fixed enumeration, low cardinality.
 /// - `status_code`: bucketed as `2xx` / `4xx` / `5xx`.
-/// - `http_status_code`: the raw code (`429`, `502`, ...), so a query can
-///   name one failure mode instead of a whole family (AISIX-Cloud#1389).
-///   It adds no series over `status_code` alone — the raw code determines
-///   the family — and `status_code` is kept so existing dashboards and
-///   alerts written against `status_code="4xx"` keep working unchanged.
+/// - `status`: the raw code (`429`, `502`, ...), so a query can name one
+///   failure mode instead of a whole family (AISIX-Cloud#1389). Named to
+///   match `aisix_proxy_requests_total`'s own raw-code label, so one
+///   spelling works across both families — and deliberately NOT
+///   `http_status_code`, which contains `status_code` as a substring and
+///   would make every `status_code="…"` matcher ambiguous. It adds no
+///   series over `status_code` alone (the raw code determines the family),
+///   and `status_code` is kept so dashboards and alerts written against
+///   `status_code="4xx"` keep working unchanged.
 /// - `user_id`: the org member behind the request, from
 ///   [`UsageEventLabels`].
 /// - `inbound_protocol`: `openai` / `anthropic`. Matches the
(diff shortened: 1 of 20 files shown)


jarvis9443:
Applied the `label_u16` nitpick in 20de40d7 — verified it against this file's own convention (`record_request` keys the same way, and `WorkerKey::label_u16`'s comment says why). The `to_string()` now only runs on a cache miss.

For the record: Copilot's review came back as a quota-limit notice with no content, so this PR has had no Copilot review.

PR #1067, opened 2026-08-28
ci(docker): run PGO only on stable release tags; drop the poc branch

jarvis9443:
## Problem

Every image build took the PGO path: ~30 min against ~11 min for a single
compile. Pushes to `main` are the bulk of that — roughly 40 per release
cycle — and those `:dev` images are never what a customer runs, so the
~20 min premium bought nothing on the builds that dominate the cost.

## Change

`PGO=on` is now derived from the ref. A stable `vX.Y.Z` tag takes the
three-phase build (instrumented → 12-shape training → profile-use rebuild);
PR, `main`/`:dev` and `-rc.N` candidates build in one compile.

The PR escape hatch is unchanged and now matters more — it is the only
automatic pre-tag exercise of the pipeline. A diff touching `Dockerfile`,
`Cargo.toml`, `Cargo.lock`, `rust-toolchain.toml`, `bench/pgo-training/**`
or this workflow still forces the full build pre-merge.

Flipping the default from on to off inverts the fail-safe direction, so the
proof-marker guard is decoupled from the mode step: its condition
re-derives "stable release tag" from `github.ref` rather than reading
`steps.pgomode.outputs.value`. A bug in the mode logic therefore cannot
skip the PGO build *and* its assertion in the same breath — a release tag
that somehow built without PGO fails the workflow instead of shipping
silently un-optimized. `ARG PGO=on` stays the Dockerfile default so a bare
`docker build` still reproduces the release shape.

A new `pgo` `workflow_dispatch` input runs the full build on demand. It is
the pre-flight for a release: run it on the release SHA before tagging and
a training-shape regression surfaces *before* the tag rather than after QA
has passed. It publishes only `:sha-<short>` and moves no pointer.

Separately, the `poc` branch trigger and the `:poc` tag are removed. The
branch no longer exists on the remote and the POC environment tracks
release tags.

## Behavior changes

- `-rc.N`, `:dev` and PR images are no longer PGO-built. The QA'd candidate
  is therefore not bit-identical to the released image. PGO changes code
  layout, inlining and block ordering — never semantics, and the workspace
  contains no `unsafe`, so there is no undefined behavior for a different
  inlining decision to expose. Functional QA still describes the shipped
  build; **performance numbers must come from a `vX.Y.Z` image or a local
  `--build-arg PGO=on` build**, never from an rc or `:dev`.
- The release tag is now the first build to run the training matrix on its
  commit. `RELEASING.md` documents the pre-flight that keeps that from
  being discovered after QA.
- `:poc` is no longer published.

## Tests

No test exception: this is CI-configuration only and has no runtime code
path to cover. The decision logic was exercised against every trigger
(main push, stable tag, `rc.1`/`rc.10`, ordinary PR, pipeline-touching PR,
dispatch with and without `pgo=true`) and matches the table above.

This PR touches `Dockerfile` and the workflow, so the escape hatch fires
and **this PR's own image build runs the full PGO path** — the change
proves the three-phase pipeline and the new mode logic in the same run.

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -5,22 +5,22 @@ name: docker-image
 # Tag strategy. Every push always carries a :sha-<shortsha> tag for
 # traceability; the other tags layer on top based on the trigger:
 #   - push to main              → :dev + :sha-<shortsha>
-#   - push to poc               → :poc + :sha-<shortsha>
 #   - push tag vX.Y.Z           → :X.Y.Z + :X.Y + :X + :latest
 #                                 + :sha-<shortsha>
 #   - push tag vX.Y.Z-rc.N      → :X.Y.Z-rc.N + :sha-<shortsha>  (NO rolling tags)
 #   - pull request              → build-only (no push), sanity check
 #   - workflow_dispatch         → :sha-<shortsha> + optional tag input
 #
-# Pre-release tags exist so a release candidate can be QA'd on the exact
-# artifacts customers receive before anything points at it. They must move
-# no rolling pointer: docker/metadata-action already skips :X.Y and :X for
+# Pre-release tags exist so a release candidate can be QA'd on the artifacts
+# customers receive before anything points at it — bar the PGO difference
+# noted at "Decide PGO build mode" below. They must move no rolling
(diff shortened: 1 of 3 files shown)


jarvis9443:
@coderabbitai review

PR #1122, opened 2026-09-03
fix(cooldown): make cooldown opt-in, and put a dropped routing candidate on the record

jarvis9443:
## Cooldown is opt-in

An absent `cooldown` block on a direct model used to mean cooldown **on**. `decide_cooldown` built a `CooldownConfig::default()` for a model that carried no block, and `enabled_or_default()` read the absent `enabled` flag as `true`.

The control plane's console omits the whole `cooldown` block whenever the operator does not enable it, and fills in no default — so the etcd document has no `cooldown` key at all. The net effect: **every direct model whose operator never asked for cooldown ran with cooldown**, taking the model out of rotation for 30s on an upstream 401, 408, 429, 500, 502, 503 or 504, and on timeout, transport, decode and stream-abort errors, while the console displayed the feature as disabled.

After this change, cooldown runs only when the operator explicitly enables it:

| stored configuration | before | after |
| --- | --- | --- |
| no `cooldown` block | on, 30s TTL, built-in trigger list | **off** |
| `cooldown` present, `enabled` unset | on, 30s TTL, built-in trigger list | **off** |
| `cooldown: {enabled: false}` | off | off |
| `cooldown: {enabled: true}` | on | on |

`decide_cooldown` returns early on a missing config rather than manufacturing a default one and asking it whether it is enabled — the shape that made the old behaviour invisible.

**Who is affected.** Any private-deployment or SaaS user who never configured cooldown on a direct model. Their models were being taken out of rotation after upstream failures and are no longer. Anyone who wants the previous behaviour enables cooldown explicitly on the model (`cooldown: {enabled: true}`); the defaults inside an enabled block are unchanged, so an enabled block with no other keys reproduces exactly what an unconfigured model used to get.

**What is deliberately unchanged.** The knobs *inside* an enabled cooldown — `default_seconds` (30), `max_seconds` (600), `honor_retry_after`, `trigger_statuses` (the same seven codes), `trigger_on_timeout`, `trigger_on_transport` — keep their values and their fallbacks. Those are parameters of a feature the operator opted into, which is a different thing from the feature switching itself on.

**Row loadability.** No field became required at the deserialization level and no parsing was tightened. Every stored model row still loads, including one carrying a partial cooldown block written before this change; only the meaning of the absent flag moved. `partial_cooldown_block_still_loads_the_row` pins that.

`enabled_or_default()` is renamed `is_enabled()` so a caller cannot pick up the old semantics under the old name.

### Documentation surfaces in this repo

The published resource schema and the Admin API reference stated the old contract in prose and are corrected with it: `schemas/resources/model.schema.json` no longer says "Omit this field to use the built-in cooldown behavior", and the `enabled` property's documented default flips from `true` to `false`. A test pins that default so the reference cannot drift back to advertising a feature the gateway does not switch on by itself.

## A dropped routing candidate is now on the record

A model group with targets A and B, where A returned an upstream status the group *was* configured to fail over on and no failover happened, left exactly one trace: a single attempt with error class `upstream_status`. Two different mechanisms produce that identical record — B was dropped from the candidate list before the dispatch loop started, or the configured `fallback_on_statuses` list never reached this gateway's snapshot — and nothing distinguished them after the fact. Neither mechanism wrote anything at any level, so raising verbosity in advance would not have helped.

Three additions, all at `WARN`, so they are present at the level a deployment is already running at (the default is `info`):

**1. The exclusion itself.** `filter_attempt_models` now returns the candidates it dropped, and `resolve_attempt_models` names each one:

```
WARN routing candidate excluded before dispatch
     virtual_model=my-group target_model=target-b reason="cooling" candidates=1
```

`candidates` is the size of the list the dispatch loop was actually left with. Only genuinely dropped candidates are reported: a cooling target that gets dispatched anyway because nothing healthier exists is not an exclusion, and `when_all_unavailable: try_anyway` reports none because it dispatches the whole list. The existing "all routing candidates are unavailable" warning now has the per-target detail beside it.

**2. Throttling.** The line sits on the per-request routing path, so unthrottled it would be one log per request for as long as a target stays out of rotation — a flood on a busy gateway, and a flood is what gets a diagnostic switched off before the incident that needs it. It is gated to one line per minute per (target, reason). A change of reason logs immediately rather than waiting out the previous window, so a target moving from cooling to background-unhealthy is not hidden behind the cooling line.

**3. Endpoint-family parity on the failed-attempt line.** The per-attempt failure WARN existed only on `/v1/chat/completions`. `/v1/messages`, `/v1/messages/count_tokens` and `/v1/responses` walk routing targets the same way and recorded nothing at all, so the same incident on an Anthropic-SDK or Codex client left no trace whatsoever. All four now go through one shared emitter, which additionally carries the group's effective `fallback_on_statuses` next to the `retryable` verdict:

```
WARN routing target attempt failed target_model=target-a target_attempt=1
     error=upstream returned HTTP 400: … retryable=false fallback_on_statuses=[400]
```

That field is what separates "the status was not in the list" from "the list never reached the gateway" — the second mechanism in the incident above.

### Log-message change

The two chat.rs messages `routing target attempt failed` and `streaming routing target attempt failed` are now one message. The streaming and non-streaming branches were never distinguished by anything a consumer reads, and one message keeps the four endpoints greppable together. An operator matching on the streaming variant's exact text should match on `routing target attempt failed` instead.

### Not added, and why

No new metric. `aisix_deployment_state` already reports a cooling or background-unhealthy target as Down, and `aisix_deployment_cooled_down_total` already counts cooldown entries, so the aggregate rate signal exists; what was missing was the per-request record naming the target, which is what the log line supplies.

### Deliberately deferred

Two adjacent blind spots of the same shape were found while reviewing this change and are **not** fixed here, because each needs a design call rather than a parallel edit:

- **Semantic routing.** `semantic::select_eligible` is the single-winner analogue of the group filter, and when its primary target is displaced for cooling or background-unhealthy it logs at `debug` and does not say which of the two. It cannot reuse this change's line verbatim: there, health is a *soft preference* — with no candidate available it dispatches the primary anyway — so "excluded before dispatch" and a `candidates` count do not describe what happened.
- **The two pre-filters ahead of the health filter.** `eligible_targets` (routing tags) and `targets_allowed_for_ip` (per-target `allowed_cidrs`) also drop candidates, and each surfaces only when it excludes *every* target; a partial exclusion is silent at any level. Neither is a mechanism in the incident that motivated this change.

## Testing

New spec `tests/e2e/src/cases/cooldown-opt-in-e2e.test.ts`, real binary + etcd + mock upstreams:

- a direct model with **no** cooldown configuration whose upstream returns 503 (a status in the old built-in trigger list) is still attempted on the next request, and `/status/models` reports it `healthy`;
- an enabled cooldown's exclusion appears in the gateway's output at the harness default log level, naming the target, `reason="cooling"`, the group, and the surviving candidate count;
- `/v1/messages`, `/v1/messages/count_tokens` and `/v1/responses` each write a failed-attempt line naming their target, with the group's `fallback_on_statuses`.

Each of the three was verified by mutation — restoring the old cooldown default, suppressing the exclusion line, and removing the sibling emitters (together, and then `/v1/responses` alone) — and each went red.

Unit coverage: `is_enabled` across the three "not asked for" shapes, the partial-block load, `decide_cooldown` returning `None` for an unconfigured model on every status in the old trigger list, the exclusion records `filter_attempt_models` returns on each of its branches, and the throttle gate's two obligations. The existing cooldown unit tests passed `None` and asserted a cooldown fired; they now run against an explicitly enabled config, so they still measure the knob defaults rather than passing vacuously.

Existing e2e specs whose subject is cooldown firing now enable it explicitly on the model that has to cool: `cooldown-contract`, `runtime-status`, `status-models`, `runtime-mixed-filtering`, `retry-on-429-vs-background-ignore`, `consistent-hash-routing` and `semantic-member-gates`.

Part of api7/AISIX-Cloud#1499.

Fixes api7/AISIX-Cloud#1499

### DIFF
--- a/crates/aisix-admin/src/openapi.rs
+++ b/crates/aisix-admin/src/openapi.rs
@@ -2344,7 +2344,7 @@ fn add_schema_defaults(doc: &mut Value) {
         ),
         (
             "/components/schemas/CooldownConfig/properties/enabled",
-            serde_json::json!(true),
+            serde_json::json!(false),
         ),
         (
             "/components/schemas/CooldownConfig/properties/honor_retry_after",
@@ -2978,6 +2978,13 @@ mod tests {
             schemas["CooldownConfig"]["properties"]["trigger_statuses"]["default"],
             serde_json::json!([401, 408, 429, 500, 502, 503, 504]),
             "runtime default cooldown trigger statuses should be visible in OpenAPI"
+        );
+        // Cooldown is opt-in (AISIX-Cloud#1499): the reference must not
+        // advertise a feature the gateway does not switch on by itself.
+        assert_eq!(
+            schemas["CooldownConfig"]["properties"]["enabled"]["default"],
+            serde_json::json!(false),
+            "cooldown must be documented as off unless the operator enables it"
         );
         assert_eq!(
             schemas["Routing"]["properties"]["when_all_unavailable"]["default"],
--- a/crates/aisix-core/src/models/model.rs
+++ b/crates/aisix-core/src/models/model.rs
@@ -112,10 +112,10 @@ pub struct BackgroundModelCheck {
     pub stale_after_seconds: u64,
 }
 
-/// Request-path cooldown settings for a direct model after retryable upstream failures.
+/// Request-path cooldown settings for a direct model after retryable upstream failures. Cooldown is opt-in: it runs only when `enabled` is `true`.
 #[derive(Debug, Clone, Serialize, Deserialize, schemars::JsonSchema, PartialEq, Eq, Default)]
 pub struct CooldownConfig {
-    /// Whether cooldown is active for this model. Set to `false` to keep the model in rotation regardless of upstream failures.
+    /// Whether cooldown is active for this model. Cooldown is off unless this is set to `true`, so a model that omits it stays in rotation regardless of upstream failures.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub enabled: Option<bool>,
     /// Cooldown TTL in seconds when the upstream did not supply a `Retry-After` header or `honor_retry_after` is `false`.
@@ -148,8 +148,13 @@ const DEFAULT_COOLDOWN_SECONDS: u64 = 30;
 const DEFAULT_COOLDOWN_MAX_SECONDS: u64 = 600;
 
 impl CooldownConfig {
-    pub fn enabled_or_default(&self) -> bool {
-        self.enabled.unwrap_or(true)
+    /// Cooldown is opt-in, so an absent `enabled` reads as `false` — the
+    /// same answer a model with no `cooldown` block at all gets. Taking a
+    /// direct model out of rotation is a user-visible availability
+    /// decision, and the gateway does not make it for an operator who
+    /// never asked (AISIX-Cloud#1499).
+    pub fn is_enabled(&self) -> bool {
+        self.enabled == Some(true)
     }
 
     pub fn default_seconds_or_default(&self) -> u64 {
@@ -302,7 +307,7 @@ pub struct Model {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub background_model_check: Option<BackgroundModelCheck>,
 
-    /// Direct-model-only request-path cooldown configuration. Omit this field to use the built-in cooldown behavior.
+    /// Direct-model-only request-path cooldown configuration. Cooldown is opt-in: omit this field, or leave `enabled` unset, and the model is never taken out of rotation by request-path failures.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub cooldown: Option<CooldownConfig>,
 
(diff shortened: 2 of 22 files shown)


jarvis9443:
On the CRITICAL flagged in the security scan (upstream error body reaching the new `error = %err` field on `log_attempt_failure`) — I checked it against the code and I'm not redacting it in this PR, because redacting this one line would close nothing and would make the remaining sinks harder to read.

The mechanism: `BridgeError::UpstreamStatus`'s `Display` is the single source for the upstream-derived text, and in the *same* `Err` arm that now calls `log_attempt_failure`, `attempt_error_from_proxy(&e)` already puts that same string into the per-attempt `AttemptRecord.error_message`, which is emitted as telemetry (`UsageEvent`) and lands in whatever exporter the deployment has configured. `access_log_error` puts it on the access log's `error` field on the same request. Both go through `sanitize_error_message` (`crates/aisix-proxy/src/attempt.rs:262`), which strips control characters and truncates — it does not redact secrets, which is the scan's own observation.

So the exposure is not introduced here. It predates this PR on the same three endpoints, and it is wider than the new log line: telemetry is persisted, stderr generally is not. `/v1/chat/completions` has also been logging `error = %err` at WARN in this exact shape for a long time; this PR moved that call into a shared emitter and wired the siblings to it, which is why the scan attributes causality to it.

Fixing it properly means one redaction policy applied where the string is produced — `sanitize_error_message` — so `UsageEvent.error_message`, the access-log `error` field and this WARN all get the same treatment. That is a behaviour change to what we persist and surface on every endpoint, with its own tests, and it does not belong in a cooldown-semantics change. Handing it to the maintainers to track on its own.

PR #1126, opened 2026-09-03
fix(security): AWS SigV4 headers are credential slots; full-version image tags only; count unscannable-body bypasses

jarvis9443:
Three independent changes on the 1.0.0 critical path, batched into one PR per this repo's one-PR-per-repository rule.

---

## 1. Security: the AWS SigV4 headers are credential slots again

**This is the important one.** `x-amz-security-token`, `x-amz-date` and `x-amz-content-sha256` were blocked outright by `RESERVED_UPSTREAM_HEADERS` in `config.rs`. When the forwarding rules moved into `crates/aisix-core/src/forwarded_headers.rs`, six of that constant's names became `CREDENTIAL_SLOT_HEADERS` — the set a glob cannot reach — and the three `x-amz-*` names landed in **no list at all**: not the credential slots, not `NON_FORWARDABLE_HEADERS`, not `NEVER_FORWARD_FROM_CLIENT`.

A provider key configured `forward_client_headers: ["x-*"]` therefore relayed the caller's AWS session token to the upstream. That contradicts the module's own stated rule — a glob is a statement about the operator's own headers, not consent to hand a third party the caller's credential.

Bedrock upstreams were never exposed: the SigV4 signer owns all three names and drops any supplied value before signing. The reach was every *other* upstream, which is exactly where an AWS credential has no business arriving.

The fix restores the credential-**slot** rule, not the old absolute block: a glob does not reach the three, and an operator who names one exactly still forwards it — identical to how the other six behave, and strictly more capable than the pre-split behaviour, which allowed neither.

`provider_key` and `passthrough_route` enumerate the credential slots in the field descriptions the control plane and the published API reference read, so both gain the three names and the schemas are regenerated. `mcp_server` refers to "a credential slot" without enumerating and needed no change.

**Scope note.** No released build ever shipped the glob-reachable state — it existed only between the refactor and this commit, both after `v0.13.0`. This is a defect fixed before release, not an advisory: no user could have observed it, and the release notes should not describe it as a behaviour change.

**Testing.** The tests name the three headers explicitly rather than iterating the constant. A loop over `CREDENTIAL_SLOT_HEADERS` can only confirm that whatever is *in* the list behaves and can never notice a name that dropped *out* of it — which is precisely how this went unseen. Both new tests go red without the fix; the gateway-level one fails by showing all three headers arriving at the upstream under `["x-*"]`, and also pins the positive direction (an exactly-named `x-amz-security-token` still forwards).

---

## 2. Image tags: publish the full version only, no `:X.Y` or `:X`

---

A release tag push published `:X.Y.Z`, `:X.Y`, `:X` and `:latest`. The two abbreviations are snapshots dressed as lines: `:1` published at 1.0.0 stops moving the day 1.1.0 ships, so anyone who pinned it silently keeps an old image under a name that suggests it tracks the major. The full version and `:latest` are the two honest answers.

The project is about to cut its first non-`0.x` version, which is the first time a `:1` would be created — and the workflow that runs on a tag push is the one at the tagged commit, so this has to be on `main` before that tag point is chosen.

`.github/workflows/docker-image.yml` — the metadata-action `tags:` block drops:

```
type=semver,pattern={{major}}.{{minor}}
type=semver,pattern={{major}}
```

`type=semver,pattern={{version}}`, `type=raw,value=latest`, `type=raw,value=dev`, `type=sha,prefix=sha-` and `type=ref,event=pr` are unchanged. The tag-strategy comment at the top of the file and the comment above the block are updated to match.

Two places that described or advertised the removed tags:

- `.github/workflows/release-draft.yml` stamps a rolling-tags sentence into every stable release body. It named the `:X.Y` tag, which would otherwise be advertised on the release page while never being pushed. It now names `latest` only; the `-rc` branch is unchanged.
- `RELEASING.md` listed the published tag set.

**Publish paths swept.** `docker/metadata-action` is invoked in exactly one place (`docker-image.yml`); it is the only workflow that builds or pushes an image, and it covers both the GHCR publish and the Docker Hub mirror. `release-draft.yml` mentions the action only in prose. No other tag-publishing path exists in the repo, and nothing in the repo pins an abbreviated tag.

**Behavior change.** A release tag now publishes `:X.Y.Z`, `:latest` and `:sha-<short>` to GHCR and the Docker Hub mirror — no `:X.Y`, no `:X`. Already-published abbreviations are untouched; this only stops new ones being created. Anything pinning an abbreviated tag must move to the full version or `:latest`.

**Testing.** CI-only with no runtime behavior, so there is nothing for a fail-before/pass-after test to bite on, and no test or linter in the repo asserts the published tag set. Verified by inspection instead: both workflows still parse, the edited `release-draft.yml` shell block still passes `bash -n`, and the downstream steps consuming `steps.meta.outputs.tags` are unaffected by there being fewer tags — the release-path steps address the image by digest, the PR-path steps take `head -n1`, and the cosign step dedups per repository and signs the digest.

---

## 3. `aisix_guardrail_bypasses_total` counts an unscannable-body bypass

`aisix_guardrail_bypasses_total` only ever counted bypasses a guardrail member reported: the increment hangs off the per-execution record, and an unscannable body has no execution by construction — the body could not be decoded, so no member ran. The chain recorded that pass-through to the audit log (and so to `UsageEvent.guardrail_bypassed_reason`) and to nothing else.

The sibling counter is the other way round: `aisix_guardrail_blocks_total` deliberately includes fail-closed paths that happen before a member executes. So one situation — the body could not be scanned — was counted when the chain refused and counted nowhere when it let the request through, which is the direction an operator reads `bypasses_total` to find. #1115 widened exactly that fail-open half in this release, so the traffic newly passing through was the traffic the counter could not see.

`GuardrailMetricsSink` gains `record_guardrail_bypass(reason)`, and `GuardrailChain::record_bypass` now feeds both receivers. A second trait method rather than a synthetic execution, deliberately: there is no member, kind or duration to report, and inventing them would put a phantom row in the per-execution latency histogram. Required rather than defaulted, so a sink cannot silently drop these.

Which pass-throughs count is unchanged and still decided in one place. `record_unevaluable_{input,output}_bypass` already distinguishes a chain whose members all fail open on that side (a bypass) from one where no member reads that side (never offered to screen it, so not a bypass), read off the same member set the refusal gate uses. The counter reads that decision rather than deriving its own.

### Behavior change

`aisix_guardrail_bypasses_total` now also counts bypasses that happen before any guardrail member executes — today that is `reason="unscannable_body"`, a body the gateway could not give the guardrails at all, which a fail-open chain lets through. The `reason` label gains that value; it is a bounded DP-internal tag and is clamped on the way in, so cardinality is unchanged in kind.

**Operators are affected.** The `reason="unscannable_body"` series was always zero for this cause and starts being non-zero, and the unlabelled total rises accordingly on any deployment that receives bodies the scanner cannot read. An alert threshold on `aisix_guardrail_bypasses_total` may fire on traffic that was always passing unscreened and simply was not counted. Nothing about which traffic is allowed or refused changes — this is a counting fix only.

Two limits are now stated where the counter is declared, because the change makes both easier to get wrong:

- `blocks_total` and `bypasses_total` are **not the same unit** and must not be summed or compared. The first fires once per request, off the terminal usage event; the second fires once per bypass event, so one outage on a chain with two fail-open members on both hooks adds 4 for a single request.
- The new producer reaches this counter only. With no member execution there is no `aisix_guardrail_latency_seconds` row, so that family's `result="bypassed"` slice stays execution-only and the two no longer agree on a total.
- The counter still does not cover every unscannable pass-through: `jobs::scan_output_blob` and `passthrough_route` scan a lossy copy with no failure policy involved and record nothing by design, so a zero is not proof that nothing went unread.

### Follow-up owed in this effort

**For the credential-slot fix (change 1).** The credential-slot names are enumerated verbatim in six places in `AISIX-Cloud: openapi/cp-admin.yaml` (provider_key `default_headers` + `forward_client_headers`, mcp_server create/update, passthrough_route ×2) and in four pages in each docs repo (`models/upstream-request-headers.md`, `mcp-gateway/upstream-authentication.md`, `reference/resources-file.md` ×2). All still list six names. This is description-only drift — the CP's `internal/cpapi/resources/forward_client_headers.go` mirrors the transport block lists and deliberately keeps no credential names, so the CP accepts `x-amz-security-token` in `forward_client_headers`, projects it, and the DP forwards it on an exact match. Nothing is functionally broken; the published contract just under-describes which names a glob no longer reaches. Tracked outside this PR.

The published metric reference and the guardrail behaviour guide contradict this change and need the paired EN + ZH docs PRs (`api7/docs` and `api7/docs.apiseven.com`): `docs/ai-gateway/traffic-controls/guardrails/behavior.md` states that a proxy-raised `unscannable_body` pass "reaches the usage record only", and `src/components/MetricReference/data.ts` describes the counter as timed fail-open *executions* and enumerates a closed `reason` list that omits `unscannable_body`. Tracked outside this PR.

### Testing

The chain unit test that pins which pass-throughs are bypasses now asserts both sinks, so the usage field and the counter cannot disagree. All three assertions were verified to go red under mutation: dropping the sink forwarding, counting every pass-through, and letting the counter derive its own predicate off a separate member set.

The bypass-reason e2e additionally asserts the `reason="unscannable_body"` delta in `GET /metrics` after driving a real unscannable request through a real DP — the positive direction only; the discrimination between a bypass and a never-offered pass-through is pinned in the unit test, where an output-only chain can be built without seeding a second route.

## Summary by CodeRabbit

- **Changes**
  - Docker releases now publish full semantic version tags, along with `latest` and SHA-based tags.
  - Major and minor version aliases are no longer published; release messaging and documentation reflect this policy.
  - Guardrail bypasses now record bounded reasons in audit logs and metrics, including unevaluable request bodies.
  - Bypass metrics are verified alongside successful request handling and usage events.
  - AWS SigV4 credential headers require explicit, exact-name forwarding and are no longer forwarded through broad wildcard patterns.


jarvis9443:
@coderabbitai the E2e Test Quality Review warning is looking at the wrong base. `7297b004` is this branch's merge base — it is `origin/main`'s current HEAD, not a commit in this PR — so the `CLAUDE.md` change it names was already on main before the branch existed and is not part of this diff:

```
$ git merge-base origin/main HEAD
7297b004ba42c7ba51b457c98d444c2e84b72fe8
$ git log --oneline origin/main -1
7297b004 docs: an absent optional config block means the feature is off (#1124)
$ git diff --name-only origin/main...HEAD
.github/workflows/docker-image.yml
.github/workflows/release-draft.yml
RELEASING.md
crates/aisix-core/src/models/guardrail.rs
crates/aisix-guardrails/src/build.rs
crates/aisix-guardrails/src/chain.rs
crates/aisix-obs/src/metrics.rs
tests/e2e/src/cases/guardrail-bypass-reason-e2e.test.ts
```

There is nothing to remove or revert — the suggested resolution would drop a commit off main. The two changes actually in the PR are batched deliberately: this repo's contributing rules ask for one PR per repository per effort rather than splitting related work across several bot-reviewed PRs.

PR #1128, opened 2026-09-04
fix(etcd): accept configuration snapshots up to 16 MiB

jarvis9443:
A full etcd range response can exceed tonic's 4 MiB default even when each configuration entry is small. In that case, the data plane cannot apply its initial snapshot and remains unready.

This raises the bounded etcd gRPC decoding limit to 16 MiB for bootstrap reads, watch responses, and Admin API reads.

Tests cover responses larger than 4 MiB through real etcd on both the configuration provider and Admin store paths.


jarvis9443:
Closing this PR because this emergency image is being built locally only.

Issue #1080, opened 2026-08-31
config-last-known-good: the stale-serving instant is re-stamped across a restart, not restored


jarvis9443:
`config-last-known-good-e2e` asserts that the stale-serving instant is continuous across a restart:

```
expect(rejection.serving_stale_since).toBe(staleSinceBeforeRestart);
```

It failed on api7/aisix#1077 with a one-second discontinuity, and passed on a plain rerun of the same commit:

```
AssertionError: expected '2026-08-31T10:30:39Z' to be '2026-08-31T10:30:38Z'
https://github.com/api7/aisix/actions/runs/33382316847/job/99457936725
```

A one-second delta that clears on rerun is the signature of a value being **re-stamped** rather than restored, where the re-stamp usually lands in the same wall-clock second as the original. The restart in that test takes well under a second, so the assertion passes whenever the two stamps fall inside the same second and fails when they straddle a boundary. That makes the observable rate a function of restart duration, not of load — which matches how rarely this shows up.

What I ruled out while looking:

- The snapshot cache preserves the instant in both directions — `snapshot_cache.rs` carries `since_unix_secs` through `store` and through the load path, so the value survives the round trip on disk.
- `Supervisor::note_stale` (`supervisor.rs:641`) uses `entry().or_insert_with(...)`, which keeps the original pin, and its comment says so explicitly.

What I did not manage to pin down, and where I would look next:

- The resync path at `supervisor.rs:889-916` runs `stale.retain(|k, _| entry_keys.contains(k) && rejected_keys.contains(k))` before re-pinning newly rejected keys with `since_unix_secs: now_unix_secs()`. If a restored entry is ever absent from `rejected_keys` at the moment that first post-restart resync runs, `retain` drops it and the loop below re-adds it with a fresh stamp. Whether that window exists depends on the ordering between the cache restore and the first resync's rejection list being populated.

Not blocking anything: it is a report-fidelity property (`serving_stale_age_seconds` would be off by a second), not a serving one — the pinned bytes themselves are unaffected. Filing rather than fixing because the config-serving path is the wrong place to land a guess, and the mechanism needs someone to reproduce the ordering rather than infer it.

Found while babysitting an unrelated PR; the failure has nothing to do with that change.

nic-6443:
Another occurrence, on #1082 at `17deaa1a`, with the same one-second signature:

```
AssertionError: expected '2026-09-01T04:07:05Z' to be '2026-09-01T04:07:04Z'
[...]

PR #1142, opened 2026-09-07
fix: multi-line SSE model echo, exact-limit token windows, and three never-emitted metric families

### LINKED ISSUE #950: ratelimit: token windows admit one extra request when usage lands exactly on the limit
## Description

Token-window admission checks in both backends admit a request while the committed count is exactly at the limit, and reject only when it is strictly over:

- memory: `FixedWindowCounter::is_exceeded` (`crates/aisix-ratelimit/src/window.rs`) — `if self.count > limit`
- Redis: the acquire script's token loop (`crates/aisix-ratelimit/src/store/redis.rs`) — `if cur > limit`

Token limiting is post-paid (check at admission, commit actual usage after the response), so crossing the cap by one in-flight response is inherent. But the strict `>` adds an avoidable case: a caller whose committed usage lands on the cap **exactly** (`count == limit`, budget fully consumed) still gets one more request admitted, overshooting by up to a full response beyond the crossing that post-paid semantics already implies.

This applies uniformly to every token window (`tpm`, `tpd`; policy-derived and key-level alike) — it is not specific to any one window and predates #771/#949, which only added a new window over the same counters.

## Proposal

Flip both comparisons to `>=` ("budget fully consumed ⇒ stop admitting") in the same change, with an exact-limit regression case per backend — e.g. cap 10, upstream reports exactly 10, the next request must be 429. Request counters are pre-paid (`would_be > limit` with the increment included) and are already exact; they need no change.

Both backends must move together (`memory ↔ redis` swap must not change observable limits).

## Priority

Low — one-request overshoot at an exact boundary; found during review of #949.

### LINKED ISSUE #1060: Three Metrics emit methods have no caller: aisix_redis_failures_total / aisix_otlp_fanout_drops_total / aisix_otlp_fanout_failures_total are never emitted
Three emit methods on `Metrics` in `crates/aisix-obs/src/metrics.rs` have no caller anywhere in the workspace — not in production code, not even in a test:

| method | family it would emit |
|---|---|
| `record_redis_failure` | `aisix_redis_failures_total` |
| `record_otlp_fanout_drop` | `aisix_otlp_fanout_drops_total` |
| `record_otlp_fanout_failure` | `aisix_otlp_fanout_failures_total` |

Each one's definition is its only occurrence. So all three families are never emitted at runtime and cannot be scraped — an operator querying `aisix_redis_failures_total` gets an empty result whether Redis is healthy or failing constantly, which is exactly the "indistinguishable from no traffic yet" symptom `CLAUDE.md` already warns about:

> **An emit function on `Metrics` with no caller is invisible to every check we run.** Its methods are `pub`, so dead-code analysis never fires; unit tests call it directly and pass; the only symptom is a series that never appears in a scrape [...] (Twice now: `record_proxy_request` until #888, then `record_deployment_request` + `record_routing_fallback` until #972.)

This is the third instance of that same class.

These three are worth wiring up rather than deleting. Redis failures and OTLP fan-out drops/failures are real operational signals — a shared rate-limit or cache backend degrading, and telemetry being silently discarded, are both things an operator needs to see and currently cannot. The call sites are the Redis client error paths in `aisix-cache` / `aisix-ratelimit` and the fan-out paths in `aisix-obs`.

Per the same rule, "shipped" here means an e2e asserts the series in `GET /metrics` after driving real traffic (or a real induced failure), not merely that `Metrics` can emit it.

One related but harmless case, for completeness: `record_llm_request` also has no production caller, but its families are emitted by `record_proxy_and_llm_request`, which is the live path used by `aisix-proxy/src/request_metrics.rs`. It is a redundant method, not a missing series.

Found while removing a separate piece of dead metrics code in #1059 (the never-emitted `M_LLM_API_LATENCY` constant), which added a guard for the constant-level version of this defect. That guard cannot catch this one: these constants *are* referenced — by the uncalled methods.

### LINKED ISSUE #1105: SSE readers that see only the first data: line: model echo, usage, provider_request_id and TTFT lost on a multi-line frame
#1102 fixed the buffered `/v1/responses` readers to walk the whole frame payload as the SSE specification requires — several `data:` lines in one frame concatenate with `\n`. Two sites in the same family stayed on "first `data:` line only", with different consequences and different reasons.

**1. `model_echo::restamp_sse_frame`.** A multi-line `message_start` or `response.created` frame does not parse, so the frame is released untouched and the client sees the upstream model id rather than the alias it addressed. Same class as #1086 / #1087 / #1088. Deferred in #1102 for a real reason: the splice needs a byte range into the original buffer, and a joined payload cannot supply one. Closing it means splicing per `data:` line, or carrying line offsets alongside the joined payload.

**2. `drain_responses_sse_frames` (`responses.rs:2728`).** Still calls `extract_sse_data_line`. This is the `/v1/responses` real-time forwarding path — the one most requests take. If an upstream writes `response.completed` across several `data:` lines, the JSON does not parse and the request records **no usage**, an empty `provider_request_id`, and no TTFT sample. This site is not covered by the deferral reason above: it only reads, it never splices, so a joined payload is enough for it.

Site 2 is the one to weigh first, because its failure is silent and it is a billing one.

The argument for leaving it is that no currently proxied provider splits a terminal frame across `data:` lines. That is the same argument that failed in #998: the SSE decoder recognised only `\n\n` while the real upstream framed with CRLF, streaming transcription billed zero tokens until someone noticed, and the tests stayed green throughout because the mocks used LF.

So settle this against a real provider stream rather than against a reading of the code — capture raw SSE bytes from each proxied Responses-API upstream and check whether any terminal frame uses more than one `data:` line. If none does, this is a tracked TODO alongside site 1. If any does, site 2 is a live billing gap.

jarvis9443:
Four independent defects, one commit each, plus a documentation-and-test rider.

## SSE frames whose payload spans several `data:` lines (Fixes #1105)

The event-stream spec lets one frame spell its payload over several `data:` lines, joined with a newline. The model restamp read only the first, so such a frame did not parse as JSON, was forwarded untouched, and the caller was answered with the upstream provider's own model id instead of the alias they addressed — on `/v1/messages` `message_start` and on the `/v1/responses` snapshot frames alike.

The restamp now reads the whole joined payload and writes each rewritten line back into the line it came from, so the framing survives byte for byte and a continuation line stays a continuation line. Both readers of the framing are now one parser: `redact::frame_payload` is expressed in terms of a new `redact::data_line_ranges`, which the restamp takes for the byte offsets it splices into. The first-line-only `messages::extract_sse_data_range` had no other caller and is removed — two scans of the same framing is what let them disagree about where a payload starts.

Swept the rest of the workspace for the same shape: the `aisix-gateway` SSE decoder already accumulates every `data:` line per event; the redactor's frame re-render deliberately collapses them into one after re-serialising; every other `data:` reader is test code. `aisix-a2a`'s relay parses each `data:` line as its own JSON-RPC envelope by design (its module doc states the assumption) and is deliberately left alone: changing it would move that relay from line framing to frame framing, and with that the point at which a malformed line fails the stream. Filed as #1143 rather than left as a note in this description.

## Token windows admitted one extra request at exactly the limit (Fixes #950)

Both backends admitted a request while the committed token count sat exactly on the cap, refusing only once it was strictly over. Token limiting is post-paid — usage is committed after the response — so crossing the cap by one in-flight response is inherent; admitting at equality overshot by a further whole response, for a caller whose budget was already entirely spent.

`FixedWindowCounter::is_exceeded` and the Redis acquire script now both read `>=`, deliberately in the same change: swapping the memory backend for the shared one must not change the observable limit. Request counters are pre-paid (`would_be > limit`, increment included) and were already exact — untouched.

**Behavior change.** `tpm`/`tpd` now refuse the request that arrives with the budget exactly consumed, where before one more was let through. This applies to every token window, policy-derived and key-level alike. A window explicitly configured to `0` now refuses every request rather than admitting one, which matches how a `0` request window (`rpm` and friends) has always behaved.

## Three metric families that were never emitted (Fixes #1060)

`aisix_redis_failures_total`, `aisix_otlp_fanout_drops_total` and `aisix_otlp_fanout_failures_total` had emit methods on `Metrics` and no caller anywhere, so an operator querying any of the three got an empty result whether the subsystem was healthy or failing constantly. Every failure they describe is fail-open by design — the shared rate limiter degrades to per-replica counting, a cache read becomes a miss, telemetry is discarded — so these series are the only place the degradation shows.

Redis: `RedisStore` (rate limit), `RedisCache` and `RedisSemanticCache` take an optional `Metrics` handle and count every failed operation. The label is subsystem-qualified (`ratelimit_acquire`, `cache_get`, `semantic_lookup`, …) because a deployment points all three at one Redis and the operator's first question is which of them is degraded. The rate limiter's existing warning is one-shot per outage, so it can say an outage started but never how long or how hard; its two detached tasks (`release`, `add_tokens`) had no log at all and the counter is their only trace. `main` builds `Metrics` before the stores so each gets the handle.

OTLP fan-out: the per-exporter pipeline counts one failure per failed export *attempt* (retries included) and counts dropped *records* by reason — `queue_full`, `worker_stopped`, `retries_exhausted`, `permanent_error`. `SinkStats` already accounted drops, but that is the heartbeat's view and resets whenever an exporter is reconfigured, so it cannot answer whether an exporter has ever lost data. `record_otlp_fanout_drop` gained a record count so one failed batch counts as many records as it carried, which is what makes it comparable with what the exporter did deliver.

## Guardrail scope follows the entry the caller addresses (Ref api7/AISIX-Cloud#1090)

Documentation and test only, no behaviour change: the chain has always been resolved once, from the addressed entry, before a target is picked.

A guardrail attached to a model runs only for requests addressed to that model. When the model is reached as a member of a routing, semantic, or ensemble group, its guardrails do not run — the operator attaches the guardrail to the group instead. That is what keeps an input guardrail's verdict independent of which member a failover lands on. The rule now lives in the `GuardrailScopeType` description (the artifact the published API reference renders and the control plane reads) and replaces the passage in the proxy's `AGENTS.md` that still described member-scoped guardrails as an open question.

## Testing

Every case was mutated to red before being trusted.

- `#1105`: unit cases for the new range scan on every framing variant a provider may emit, plus e2e — a `message_start` written over two `data:` lines, asserted to come back carrying the alias with both `data:` lines intact. Reverting the restamp to first-line-only fails it.
- `#950`: exact-limit cases for both backends (memory and a live Redis), plus a one-short-of-the-limit case so the assertion is about the budget being spent rather than about any usage at all. Reverting either comparison fails them.
- `#1060`: tests are on the production path, not on `Metrics`. A rate-limit integration test drives a real store through a relay that starts answering Redis errors mid-test; an e2e drives real traffic to an OTLP receiver that refuses more times than the pipeline will retry, then asserts both families in `GET /metrics`.
- Guardrail scope: a routing group over the guardrail-scoped model — addressed to the group the request is not screened, addressed to the member it is. Making model scope match everything fails it.

The `#1060` Redis relay answers a Redis error rather than dropping the socket on purpose: a dropped socket sends the client's connection manager into its own multi-minute reconnect backoff, so the operation under measurement would not return for the length of it. Both shapes reach `RedisStore` as the same `Err`.

## Not included: #1080

`config-last-known-good`'s stale-serving instant is **already restored correctly on `main`**, so nothing here changes it and the issue is not closed by this PR.

The mechanism the issue proposes cannot occur: `apply_resync`'s re-pin loop iterates `stats.rejections`, so a key dropped by the preceding `retain` (which drops exactly the keys that are *not* rejected) can never be re-added by it in the same call. A fresh stamp therefore requires the pin to be absent from the restored state, which means a cache file predating the pin.

Verified end to end against real etcd: the cache file at restart carries `since_unix_secs` matching the pre-restart `serving_stale_since`, with the rejected bytes alongside it, and the assertion held on 25 consecutive runs. Two durability gaps that could leave that file behind the pin — a spawned cache write nothing awaited at shutdown, and `store` committing an older snapshot after a newer one — were both closed by #1134 on 2026-09-05, after the issue was filed and after the CI failure it reports.

## Summary by CodeRabbit

* **Bug Fixes**
  * Rate limits now reject requests when usage exactly reaches the configured limit.
  * Streaming responses with multiline event payloads preserve framing, content, aliases, and usage data.
  * Guardrails consistently follow the directly addressed model or group entry.

* **Observability**
  * Added metrics for Redis cache, rate-limit, and telemetry delivery failures, retries, and dropped records.
  * Redis-backed services continue operating according to existing failover behavior while reporting failures.

* **Documentation**
  * Clarified guardrail scope and behavior for routed, semantic, and ensemble requests.


jarvis9443:
@coderabbitai thanks — taking the three pre-merge checks in turn. One was right and I've pushed for it; two I'm not acting on, with the evidence below.

**E2E completeness — agreed, fixed in cc7658ee.** You're right that the Redis and token-limit changes stopped at the backend layer, and for the metric that's not a style preference: this repo's rule is that a metric family is shipped when an e2e asserts it in `GET /metrics` after real traffic, precisely because an emit exercised only by a unit test is the shape that shipped uncalled three times before (which is what #1060 is about). So there are now two API-level cases. `completions-tpm-commit-e2e` drives a caller whose upstream reports usage landing exactly on its `tpm` cap and asserts the next request is 429 with `x-ratelimit-scope: tpm` and `remaining: 0`. `ratelimit-cluster-e2e` runs a gateway on the shared Redis backend through a relay the test cuts mid-run, asserts the next request is still served (the fail-open contract, and why the outage is otherwise invisible), and asserts `aisix_redis_failures_total` rises in the scrape. Both go red with their fix reverted. The direct store tests stay as the lower layer.

**`/v1/responses` for #1105 — already fixed on `main`, before this PR.** That's the site the issue calls "site 2", and #1110 closed it. On `origin/main` today, `drain_responses_sse_frames` (`responses.rs:2752`) reads `crate::redact::frame_payload(&frame)` — the whole joined payload — under a comment naming exactly the failure you're describing, and `extract_sse_data_line` no longer appears anywhere in that file. What was left, and what this PR fixes, is the restamp site: it needed byte offsets to splice into, which a joined payload could not supply, so it stayed first-line-only. That's the one remaining reader, and the sweep for others is in the PR description.

**Guardrail changes out of scope — deliberate, and linked.** They carry `Ref api7/AISIX-Cloud#1090` in both the commit message and the PR description. Splitting them out would work against the repo's own batching rule: one PR per repository per effort, because the review bots here bill and rate-limit per PR, so fanning an effort into small PRs burns review quota and stalls on throttled reviews. The commit is documentation and one test with no behaviour change, which is why it rides along rather than waiting.

PR #1154, opened 2026-09-08
fix(routing): avoid deadlock when exclusion log throttle expires

ormandj:
When a routing target is excluded repeatedly, the first exclusion after its 60-second log throttle expires can block the request worker indefinitely before provider dispatch. In `should_log_exclusion`, the `DashMap::get()` guard used as the match scrutinee remains alive in the fallback arm, where `insert()` attempts to acquire the same shard's write lock. This also prevents the request from reaching the upstream timeout and can accumulate blocked workers during an outage.

Use `DashMap::entry()` to check and update the timestamp under one guard. This removes the lock upgrade and ensures concurrent callers emit at most one exclusion log per group/model/reason per interval. Routing order, cooldown settings, and public interfaces remain unchanged.

Validation:
- The expiry regression reproduces the original deadlock and fails through a separate two-second watchdog instead of hanging the test suite.
- Concurrent fresh-entry and expired-entry tests require exactly one caller to pass the throttle. A further regression verifies suppressed calls do not refresh the timestamp and postpone logging indefinitely; deliberately introducing that defect makes the test fail.
- All 1,209 proxy unit tests pass on upstream main, including 35 health tests and 76 routing tests. All 33 health tests and 68 routing tests also pass against v1.1.0 with the same fix.
- `cargo fmt --all -- --check` and `git diff --check` pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved rate limiting for health-related logging to prevent race conditions and deadlocks.
  * Ensured suppressed log attempts do not extend the throttle window.
  * Corrected throttling behavior so the gate reopens after the configured interval.
  * Limited concurrent logging to one permitted writer per interval.

* **Tests**
  * Added coverage for throttle expiration, suppressed calls, and concurrent access.

### DIFF
--- a/crates/aisix-proxy/src/health.rs
+++ b/crates/aisix-proxy/src/health.rs
@@ -1644,6 +1644,20 @@ mod tests {
     }
 
     #[test]
+    fn exclusion_log_gate_does_not_extend_the_window_when_suppressed() {
+        let t = ModelRuntimeStatusTracker::new();
+        let key = ("g-1".to_string(), "m-1".to_string(), "cooling");
+        let logged_at = Instant::now() - EXCLUSION_LOG_INTERVAL / 2;
+        t.exclusion_log.insert(key.clone(), logged_at);
+
+        for _ in 0..8 {
+            assert!(!t.should_log_exclusion("g-1", "m-1", "cooling"));
+            // Busy traffic must not postpone the next log indefinitely.
+            assert_eq!(*t.exclusion_log.get(&key).unwrap(), logged_at);
+        }
+    }
+
+    #[test]
     fn exclusion_log_gate_reopens_after_the_interval_without_blocking() {
         let t = Arc::new(ModelRuntimeStatusTracker::new());
         t.exclusion_log.insert(
--- a/crates/aisix-proxy/src/health.rs
+++ b/crates/aisix-proxy/src/health.rs
@@ -721,12 +721,20 @@ impl ModelRuntimeStatusTracker {
         model_id: &str,
         reason: &'static str,
     ) -> bool {
-        let now = Instant::now();
         let key = (virtual_name.to_string(), model_id.to_string(), reason);
-        match self.exclusion_log.get(&key) {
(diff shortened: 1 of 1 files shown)


ormandj:
@jarvis9443 @nic-6443 — could you review this routing deadlock fix? Repeated target exclusions can permanently block request workers once the 60-second logging throttle expires, before provider dispatch or the upstream timeout. The regression reproduces the original deadlock, the fix and concurrent/suppression tests pass, and all 1,209 proxy unit tests pass.

The [CLA clarification questions on #1047](https://github.com/api7/aisix/pull/1047#issuecomment-5410952049) also remain open. Thanks!

jarvis9443 (approved):
Verified independently, approving.

On `main`, `should_log_exclusion` keeps the `DashMap::get()` read guard alive as the match scrutinee temporary into the fallback arm that calls `insert()` on the same shard; dashmap 6.1.0's shard lock is a non-reentrant RwLock, so the first exclusion after the 60s window expires parks that worker forever, and every later write to the same shard parks behind it. Applying only your tests onto main's implementation reproduces it (the expiry test times out, the concurrency test sees 3 writers instead of 1); both pass on this branch. I also swept every DashMap/DashSet use in the workspace and this was the only get-then-mutate-under-guard site, so nothing else needs the same treatment.

The red `e2e (work-stealing)` leg failed on `metric-label-selection-e2e` startup validation with an assertion text identical to a failure on an unrelated branch yesterday, i.e. a known flake; re-running it.

jarvis9443:
Merged as 7d6d14bb — thanks for the thorough report and the fix; it will ship in the next release.

ormandj:
> Merged as [7d6d14b](https://github.com/api7/aisix/commit/7d6d14bbf5f5ec4577466da48ce7763d519d81bb) — thanks for the thorough report and the fix; it will ship in the next release.

Anytime, I look forward to contributing more over time! Thank you for creating this project and all of your hard work!

PR #1155, opened 2026-09-08
fix(a2a): parse multi-line SSE events

### LINKED ISSUE #1143: aisix-a2a: an SSE frame whose JSON-RPC envelope spans several data: lines fails the stream
`aisix-a2a`'s SSE relay reads one `data:` line as one JSON-RPC envelope, so an agent that spells a long envelope across several `data:` lines fails the whole stream rather than being read.

`sse_events` (`crates/aisix-a2a/src/bridge.rs`) splits the byte stream on `\n` and hands each line to `parse_sse_data_line`, which does `strip_prefix("data:")` and parses the remainder as a complete envelope. The event-stream spec joins a frame's `data:` lines with `\n` into one payload, so a continued envelope arrives as two invalid fragments — and the first one returns `A2aError::Request("malformed JSON-RPC event")`, whose arm `return`s and ends the stream mid-task.

The assumption is deliberate and documented in that function's doc comment ("A2A carries one JSON-RPC envelope per `data:` line"), which is why #1142 fixed the same reader class in `aisix-proxy` and left this one alone rather than changing it silently. Closing it is not a one-line edit: it moves the relay from line framing to frame framing, and with that the point at which a malformed line fails the stream — today the first bad line fails immediately, whereas a frame reader has to buffer to the blank line before it can decide. That trade is worth making deliberately.

Same class as #1105 and #1100.

mat973252-coder:
Fixes #1143.

An A2A agent can split a JSON-RPC envelope across several `data:` fields in one SSE event. The relay previously parsed each line separately and terminated the stream on the first incomplete JSON fragment.

Buffer until the event's blank line, join its data fields with newlines, and parse the envelope once, following the [SSE event-stream rules](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation). This handles LF/CRLF framing and network chunk boundaries while ignoring comments and metadata. The existing 16 MiB limit now bounds the accumulated event. Existing compatibility for an unterminated final event is retained.

Malformed JSON is now reported when the complete event arrives, or at EOF, rather than after its first data line. Existing configurations need no changes.

Regression coverage includes multiline payloads, interspersed metadata, split CRLF/JSON writes, malformed JSON, and an oversized event made of individually small lines. The HTTP round-trip regression fails against the original reader with `malformed JSON-RPC event` and passes with this change.

### Validation

Validated on Rust 1.93.1 in a Linux checkout, at commit `7f73bb63b569288a3284559dcbb28d741c911567`:

- `cargo fmt --all --check` passed.
- `cargo check --locked --workspace` passed.
- `cargo clippy --locked --workspace -- -D warnings` passed.
- A2A: 34 unit tests and 23 HTTP integration tests passed.
- `cargo test --locked --workspace --no-fail-fast` has one failing test: `aisix-mcp::bridge::tests::connect_timeout_bounds_an_unreachable_upstream`.

The same full-workspace test command, in the same environment, on unmodified upstream `main` (`c8e96aa`) fails only that same MCP test. It exceeds the test's eight-second bound while reaching the unreachable test address. No tests were skipped or weakened to hide this failure. Local proxy/network behavior affects this test, so this draft does not claim a green full-workspace test run; upstream CI still needs to validate it.

The separate repository E2E suite has not been run locally. This change is limited to the A2A reader and its tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming event handling for responses split across multiple data fields or network chunks.
  * Correctly reassembles multiline event payloads, including streams using CRLF line endings and keep-alive messages.
  * Prevents oversized events from being processed and reports a clear stream error when the size limit is exceeded.

### DIFF
--- a/crates/aisix-a2a/src/bridge.rs
+++ b/crates/aisix-a2a/src/bridge.rs
@@ -590,17 +590,17 @@ impl A2aBridge for HttpBridge {
     }
 }
 
-/// Parse an upstream SSE body into the JSON-RPC envelope of each `data:` field.
+/// Parse an upstream SSE body into one JSON-RPC envelope per event.
 ///
-/// Deliberately minimal: A2A carries one JSON-RPC envelope per `data:` line, so
-/// `event:` / `id:` / `retry:` fields and comments are metadata this gateway has
-/// no use for and passes over. A `data:` line that is not JSON ends the stream
-/// with an error rather than being skipped — a caller that silently dropped
-/// events would report a truncated task as a complete one.
+/// Join an event's `data:` fields with newlines before parsing, as specified by
+/// https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation.
+/// Metadata and comments are ignored. Malformed JSON fails the stream once the
+/// event ends, rather than silently reporting a truncated task as complete.
 fn sse_events(resp: reqwest::Response) -> impl futures::Stream<Item = A2aEvent> + Send {
     async_stream::stream! {
         let mut bytes = resp.bytes_stream();
         let mut pending: Vec<u8> = Vec::new();
+        let mut line_start = 0;
         loop {
             let chunk = match bytes.next().await {
                 Some(Ok(chunk)) => chunk,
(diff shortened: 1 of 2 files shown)


mat973252-coder:
Hi @jarvis9443, a gentle follow-up on this fix when you have a chance.

I've rebased the branch onto the latest `main` at `4875c2d`; the PR head is now `7625bec`. The rebase was conflict-free, and `git range-diff` confirms the original fix is unchanged.

Validation on the rebased revision (Rust 1.93.1, Linux/WSL):
- `cargo test -p aisix-a2a --locked`: all 57 tests pass (34 unit + 23 HTTP integration tests).
- `cargo fmt --all -- --check`: pass.
- `cargo clippy -p aisix-a2a --all-targets --locked -- -D warnings`: pass.

The regression coverage includes multiline SSE data, chunk boundaries, split CRLF, malformed trailing data, and the event-size limit. These are local checks, not a claim that the full project CI has passed.

The new [CI run](https://github.com/api7/aisix/actions/runs/35674741804) and the other fork workflows are still awaiting approval. Could a maintainer approve them and help review this when convenient? If there's anything you'd like changed in the approach or additional coverage you'd find useful, I'm happy to follow up. Thanks for your time!

jarvis9443:
Thanks @mat973252-coder, this is a clean fix and the regression coverage made it easy to land. We added a few small follow-ups on top before merging: bare-CR line endings, stripping a leading BOM, and an e2e case that covers the gateway's re-framing of multi-line events. Merged now, thanks again!

PR #1161, opened 2026-09-10
fix: report the build version in outbound User-Agent headers

jarvis9443:
Upstream requests advertise `aisix/0.1` even when the gateway is running a newer release. Set the default `User-Agent` in the shared HTTP client builder to `aisix/<build version>`, using the same release-aware version source as `aisix --version` and the `Server` response header.

Remove the fixed values from the OpenAI, Anthropic, Azure OpenAI and Vertex bridges, the proxy HTTP client, and the worker and Provider Key TLS pools. Clients using this shared builder now inherit the versioned default; heartbeat and telemetry retain their existing identities. The exporter client keeps its `aisix-dp/` prefix but also uses the release build version instead of the Cargo package version. No configuration change is required.

Regression coverage sends real requests through the gateway and etcd to a local upstream across chat completions, Anthropic messages and Responses, with both OpenAI and Anthropic upstreams and in both serving modes and with both default and Provider Key TLS pools. Each case checks the outgoing UA against the binary's own version. The source guard also rejects dispatch clients that override the shared default. The OTLP HTTP test exercises the production exporter client and checks the version received over HTTP.

CI stamps the E2E binary and the targeted exporter test with a version different from the Cargo package version, so accidentally reverting to the package version fails the regression checks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -298,6 +298,11 @@ jobs:
           name: coverage-unit
           path: lcov-unit.info
           retention-days: 7
+      - name: exporter User-Agent uses the release version
+        env:
+          # Differ from Cargo's version so the old source cannot pass.
+          AISIX_BUILD_VERSION: 0.0.0-ci
+        run: cargo test -p aisix-obs --lib otlp_http_sink::tests::otlp_sink_posts_one_request_with_all_spans -- --exact
 
   build-bin:
     name: build aisix (instrumented)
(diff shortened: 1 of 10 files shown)


jarvis9443:
Added the Anthropic bridge scenario to the same E2E matrix. It sends `/v1/chat/completions` through an Anthropic provider with a bare upstream base URL, covering both serving modes and both default and Provider Key TLS pools. Each case checks the exact outgoing User-Agent against the binary version. CI also stamps the test binary with a version distinct from the Cargo package version.

PR #1167, opened 2026-09-11
perf(etcd): adapt watch coalescing to configuration apply cost

jarvis9443:
Spaced configuration updates can repeatedly clone and hash a large snapshot because each gap exceeds the fixed 20 ms quiet period. Adapt that period to the previous apply's processing cost, within the existing 150 ms maximum coalescing wait. Cheap applies and writes following an idle interval retain the 20 ms wait.

The event cap, validation, last-known-good handling, canonical hash format and per-batch revision publication stay unchanged; no configuration changes are required. Regression coverage pins the quiet-period cap and idle reset, and exercises spaced model updates followed by API-key revocation and restoration against a real gateway and etcd. CPU savings are measured separately from functional E2E assertions to avoid timing-dependent tests.

Fixture readiness now waits for caller keys written after their dependencies, so separately delivered watch batches cannot start jobs, TLS, cache, last-known-good or header-forwarding assertions with a partial fixture.

Fixes api7/AISIX-Cloud#1569

## Summary by CodeRabbit

- **Improvements**
  - Configuration updates now adapt batching delays based on recent processing time, helping manage expensive updates while keeping routine changes responsive.
  - Adaptive timing is bounded and resets after inactivity for more predictable update behavior.

- **Tests**
  - Expanded end-to-end coverage for configuration propagation, API-key authorization, caching, TLS access, and last-known-good configuration handling.
  - Test setup now more reliably waits for configuration changes to become available before validating requests.

### DIFF
--- a/crates/aisix-etcd/src/supervisor.rs
+++ b/crates/aisix-etcd/src/supervisor.rs
@@ -1663,6 +1663,7 @@ impl<P: ConfigProvider> Supervisor<P> {
         // stream). Held here so the next loop iteration handles it
         // exactly as if it had just arrived.
         let mut pushed_back: Option<Option<Watched>> = None;
+        let mut apply_timing = ApplyTiming::default();
 
         loop {
             if *cancel.borrow() {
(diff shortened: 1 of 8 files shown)


jarvis9443:
The full-scale stability acceptance remains open. This PR changes only in-memory watch coalescing; snapshot persistence is unchanged, and the agreed benchmark uses the default snapshot-off configuration. Snapshot writer redesign is deferred.

The focused real-gateway/etcd tests check bounded coalescing, hash consistency, revision publication, rejected-update handling and API-key revocation/restoration. A separate controlled comparison at 70,350 projected rows and 14 updates/s measured fewer applies and lower CPU with identical final hashes; it is not a substitute for full-stack acceptance.

The authorized acceptance sequence uses published main images: a fresh baseline, all 1/5/14 updates/s and batch/shared-reference pressure gates, then an uninterrupted 24-hour run and separate recovery tests. Those gates remain mandatory after this change is merged and published. The existing failed pressure run is retained and is not reported as passing. No snapshot-persistence improvement or completed stability acceptance is claimed by this PR.

PR #1174, opened 2026-09-12
perf(metrics): reuse sorted registry series across scrapes

jarvis9443:
High-cardinality scrapes repeatedly collected and sorted the entire registry even when its series membership had not changed. Cache the sorted series handles and rebuild them when a counter, gauge or histogram is registered. Upkeep shares the same directory.

Each scrape still reads current values, drains pending observations and evaluates rolling summary windows. Default labels, exposition, histogram buckets and scrape cancellation behavior are unchanged. Existing-series recording adds no cache lock or generation update; only insertion invalidates the directory.

A release ABBA mechanism comparison with 53,260 synthetic keys and 723 MB exposition reduced warmed render CPU from 0.8543 s to 0.5666 s (33.68%). All 24 canonical outputs matched; peak RSS remained about 2.97 GB. Initial upkeep plus first-render CPU was comparable. This measures recorder work, not full gateway stability acceptance.

Tests cover live values, membership invalidation across all three kinds, directory reuse, concurrent registration/rendering, reference-recorder equivalence and rolling windows. Real gateway coverage includes new label combinations after a warmed scrape. Timing-sensitive performance is measured separately from functional E2E assertions.

Fixes api7/AISIX-Cloud#1593

## Summary by CodeRabbit

- **Bug Fixes**
  - Metrics now consistently include newly registered series during scrapes.
  - Metric values remain current while cached metric catalogs are reused.
  - Concurrent streaming and non-streaming requests now report observations correctly for new model and stream label combinations.

- **Performance**
  - Repeated metrics rendering and maintenance operations are more efficient, especially for larger metric catalogs.
  - Metric collection remains accurate while reducing repeated catalog rebuilding during scrapes.

### DIFF
--- a/crates/aisix-obs/src/prometheus.rs
+++ b/crates/aisix-obs/src/prometheus.rs
@@ -208,6 +208,7 @@ impl DistributionSeries {
 
 struct Storage {
     distributions: DistributionBuilder,
+    generation: Arc<AtomicUsize>,
 }
 
 impl metrics_util::registry::Storage<Key> for Storage {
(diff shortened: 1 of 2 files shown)


jarvis9443:
The sequential `series_catalog_is_reused_while_values_and_membership_stay_live` test already isolates each insertion: it renders after the counter, then registers and renders the gauge without another counter registration, then does the same for the histogram. Removing only the gauge generation increment fails the `active 2` assertion; removing only the histogram increment fails the `latency_count 1` assertion. Both mutations were checked independently, along with counter invalidation and directory reuse. The concurrent test supplements these isolated assertions. No additional test is needed for this finding.

PR #1176, opened 2026-09-13
perf(config): share resource index strings across snapshots

jarvis9443:
Configuration snapshot copies currently duplicate every resource ID and name in the two lookup indices, and each name lookup allocates another ID string. Share immutable index strings across snapshots and between the name and ID indices to reduce allocation and destruction work during configuration changes.

Each snapshot retains independent maps, counts and generations. Renames, deletions, conflict checks and the configuration digest contract retain their existing behavior. Name lookups release the name shard before acquiring the ID shard. No configuration changes are required.

Allocation regressions cover snapshot copying and name lookup, together with mutation isolation. The real-etcd configuration-digest E2E also checks request routing through model rename, deletion and restoration. The MCP OpenAPI fixture now waits for its last caller key to authenticate before testing tool permissions, keeping its scenarios and assertions unchanged.

Fixes api7/AISIX-Cloud#1607
Fixes api7/AISIX-Cloud#1608

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved model lifecycle handling during configuration changes. Renamed models no longer respond under their former names, while new names remain available.
  - Deleting a model now properly removes its service availability.
  - Recreating a deleted model restores service availability as expected.
  - Improved configuration propagation so model availability is updated reliably across supported integrations.

- **Performance**
  - Improved resource indexing efficiency, reducing unnecessary memory allocations during lookups and table operations.
  - Resource table copies now share storage more efficiently without duplicating entries.


jarvis9443:
Both pre-merge warnings are addressed:

- The fixture correction is tracked by api7/AISIX-Cloud#1608 and linked in this PR's closing keywords. It fixes the actual `tools/list` 401 in CI34741150793. With the same CI binary, delaying the final key's notification reproduces that failure in both serving modes; the corrected gate passes all six unchanged cases in both modes, with and without the delay. This demonstrates the readiness gap, not the exact timing of the original CI failure.
- The ordering is an existing harness contract, documented in [`tests/e2e/AGENTS.md`](https://github.com/api7/aisix/blob/2bd03d4c5a5f51f0687e683c7f58bbd250eaad70/tests/e2e/AGENTS.md). This fixture writes the server, full key and scoped key through one prefix, in that order. [`Supervisor::apply_events`](https://github.com/api7/aisix/blob/2bd03d4c5a5f51f0687e683c7f58bbd250eaad70/crates/aisix-etcd/src/supervisor.rs#L1071) stages events in order and publishes all mutations together. The MCP handler reads the snapshot per request; [`OpenApiBridge::tools`](https://github.com/api7/aisix/blob/2bd03d4c5a5f51f0687e683c7f58bbd250eaad70/crates/aisix-mcp/src/openapi.rs#L160) generates tools synchronously from the server row, with no separate asynchronous discovery step to warm up. Therefore final-key authentication is sufficient for fixture propagation. It deliberately does not test MCP behavior: the existing tools/schema/call/ACL assertions do that and remain unchanged. A broken server row or tool generator should fail those assertions rather than exhaust a setup probe.

Keeping the last-key read-only gate. No assertion, timeout, retry policy or product code was changed by this follow-up.

PR #1178, opened 2026-09-13
perf(etcd): isolate configuration apply from async I/O workers

jarvis9443:
Large etcd configurations parse, clone snapshots and calculate configuration digests synchronously on Tokio I/O workers. Unrelated sockets can then wait for an entire apply operation even when another runtime worker is available.

Hand the async work queue back to Tokio while initial loads, reconnect loads, resyncs and incremental batches finish synchronously in place. This preserves canonical digests, publication order, adaptive batching and shutdown ordering without detached or concurrent configuration applies. Current-thread embedders retain their existing inline behavior; no configuration changes are required. On multithread runtimes, library callers must run the supervisor outside `LocalSet`, using `tokio::spawn` instead of `spawn_local`; both public async entry points document this requirement.

Add a regression for worker availability and in-flight cancellation, and exercise metrics scrapes and proxy requests concurrently with updates in the real DP/etcd digest lifecycle E2E. Scheduling performance is measured separately from functional E2E to avoid hardware-dependent latency assertions.

Fixes api7/AISIX-Cloud#1610


jarvis9443:
Fixed in 2760a5b2: unexpected join errors now fail the test, and the work-completion assertion remains.

Requiring only a cancelled result would reject a valid completion. With the pinned Tokio 1.52.1, the synchronous apply can finish without another suspension point and return `Ok(())` after `abort()`. The test now accepts success or cancellation while rejecting panics. Injecting a panic after the configuration work passed the old test but fails the updated check; the unchanged task passes.

jarvis9443:
Documented the restriction on both public async entry points in 687712be and added a regression for the supported call pattern from a `LocalSet`: `tokio::spawn` for initial loading and the watch loop, including an incremental update and graceful cancellation. Direct `spawn_local` variants reproduce the Tokio panic for both entry points; the documented path passes, as do the existing current-thread load and in-flight cancellation tests.

This matches the pinned Tokio 1.52.1 behavior. Both gateway serving modes already create a multithread control runtime and use `tokio::spawn` for the supervisor (`crates/aisix-server/src/main.rs`); the production scheduling implementation is unchanged.

PR #1196, opened 2026-09-16
fix(bedrock): give each test's SDK client its own connection pool

jarvis9443:
## Problem

`main` is red on the required `rust unit + coverage` check.
`aisix-provider-bedrock::bridge::tests::sdk_does_not_retry_on_its_own`
fails with

```
expected UpstreamStatus 500, got Transport("upstream dispatch failed")
```

It reproduces locally in roughly one run in eight when the bedrock test
binary runs its tests concurrently, and never in isolation. Printing the
error chain names the mechanism:

```
DispatchFailure { source: ConnectorError { kind: User, source:
  hyper::Error(User(DispatchGone), "runtime dropped the dispatch task") } }
```

## Root cause

The defect is in the test setup, not in the assertion — the test is
correct and stays exactly as it is.

The Bedrock SDK's HTTP client is a process-wide singleton
(`upstream_tls::aws_http_client()`), so every test in the binary shares
one hyper connection pool, while each `#[tokio::test]` owns a runtime
and drops it at the end of the test — the runtime the pool's connections
were spawned on. The mock servers bind ephemeral ports that the OS
recycles aggressively (one instrumented run built 68 SDK clients across
five distinct ports), so a later test routinely lands on an authority
the pool still holds an entry for. A connection checked out while the
neighbouring test's runtime is being torn down carries a dispatch task
that is already gone, and hyper fails the request before it reaches the
server — which the bridge maps to `BridgeError::Transport`, exactly what
the failure reports.

Nothing about the gateway changed to cause this; #1193 only shifted the
scheduling that decides whether the window is hit.

## Fix

One pool for the process is right for the gateway, which runs a single
tokio runtime for as long as it lives, and is only wrong for a test
binary. `upstream_tls` now exposes the builder behind that singleton,
and under `cfg(test)` each Bedrock SDK client builds a pool of its own.
Production is unchanged: the `cfg(not(test))` arm is the previous call,
so a release build compiles to exactly what it did before, still on the
shared TLS material `upstream.tls.ca_file` configures.

The guardrail's Bedrock client is the only other SDK client on the same
singleton, so it gets the same treatment in the same pass. It has not
been observed to fail — it is the identical construct with the identical
exposure.

Two source scans keep the split honest, because the builder is now a
`pub` function that a production call site must not reach — `build_client`
runs per Bedrock request, so calling it there would rebuild the connector
and re-read the platform trust store on every request. The existing
outbound-TLS scan now requires the call spelled in full
(`upstream_tls::aws_http_client()`), which the builder's name no longer
satisfies as a substring; and a new scan reads the call sites rather than
the file, requiring every call of `build_aws_http_client()` to sit in a
function carrying `#[cfg(test)]`. Both were checked by mutation: pointing
the production branch at the builder turns each of them red.

The `reqwest`-side process-wide clients (a2a bridge, MCP tool + OAuth,
proxy, JWKS) share the shape but a different HTTP stack, and none has
been observed to flake; they are deliberately left alone rather than
changed speculatively.

## Testing

A deterministic regression test is not available here: the failure is a
race between one test's runtime teardown and another test's pool
checkout, and every staged version of that sequence (same recycled port
across two runtimes, a runtime left alive but idle) resolves benignly —
the pool correctly discards a connection whose runtime is fully gone. A
connection-count test does not discriminate either, since the SDK stack
dials a fresh connection per call in both the fixed and unfixed builds.

What does discriminate is the failure rate itself, measured over
back-to-back runs of the bedrock test binary at the concurrency that
reproduces it:

- before: **38 failures / 300 runs**
- after: **0 failures / 300 runs**

`cargo test` for `aisix-gateway`, `aisix-guardrails` and
`aisix-provider-bedrock` passes, and `cargo clippy --all-targets` on the
three is clean.

## Behavior changes

None. The change is confined to `cfg(test)` builds.


jarvis9443:
Fixed in 1f03166. You're right that the file-level rule doesn't bind to the production branch — I reproduced exactly the case you described: rewrite `sdk_http_client`'s `cfg(not(test))` arm to call `build_aws_http_client()` while leaving one `aws_http_client()` mention behind in a comment, and the existing scan still passes.

So there's now a second scan that looks at the call sites instead of the file: every call of `build_aws_http_client()` has to sit in a function carrying `#[cfg(test)]`, walking back from the call to its signature and over the attributes above it. Under that mutation it fails and names `bridge.rs:354`.

The scanner's own file is skipped, since it necessarily spells the probe in its source; the builders it guards live in `upstream_tls`.

PR #1210, opened 2026-09-18
perf(telemetry): fill timed batches from the ready queue

jarvis9443:
Timed telemetry flushes can send an undersized batch while the receive queue already contains a backlog. Fill the nonempty buffer from immediately available events before the timed flush, up to the existing 100-event limit, to avoid unnecessary serial requests.

The queue bound, FIFO event contents, serial delivery, flush interval, failure handling and cancellation drain stay unchanged. No configuration changes are required.

Regression tests drive the real sender and timer against an HTTP receiver, covering backlog batching, complete event contents and order, failed sends, partial timed batches and cancellation. The managed sender boundary is covered here because the standalone DP harness has no managed control-plane receiver; unchanged business-path E2E suites remain part of CI. This change does not claim end-to-end loss elimination or a measured throughput improvement.

Fixes api7/AISIX-Cloud#1684

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved telemetry event processing by draining queued events in batches when the flush interval occurs.
  * Preserved event ordering and ensured pending events are handled during cancellation.
  * Improved handling of partial batches and failed batch delivery.


jarvis9443:
The test boundary is deliberate for this internal scheduling change. The regression drives the actual `run` future that `telemetry::spawn` invokes, holds a partial buffer, then makes the receive queue and timer ready together. It fails without the refill call (first batch 5 rather than 100) and with the cap removed. The receiver checks serialized contents and FIFO order, including repeated request IDs; separate cases check the original failed-send, partial-interval and cancellation behavior.

`spawn`/`UsageSink` wiring and proxy event emission are unchanged. The refill exits immediately for either empty or disconnected channels; the existing channel-close flush is unchanged. The PR description records the managed-receiver limitation and the unchanged business E2E coverage. Full CP+DP usage-presence and drop validation remains required after publication; these tests do not claim end-to-end loss elimination. Adding a broad proxy scenario would not pin this timer/receive scheduling regression more directly.

PR #1214, opened 2026-09-21
docs(readme): drop the --platform flag now that images are multi-arch

moonming:
One-line removal: the quickstart's `--platform linux/amd64` is obsolete since the published images became multi-arch (`latest` = linux/arm64 + linux/amd64).

## Verified

Ran the README quickstart **verbatim** (both YAML blocks and the `docker run` extracted from the edited README) against `ghcr.io/api7/aisix:latest` (1.3.0) on an Apple Silicon host:

- pulls native `linux/arm64`, no platform warning
- `GET /livez` → `ok`; `GET /v1/models` → `["my-model"]`
- `POST /v1/chat/completions` with `my-model` reaches OpenAI (401 from the upstream on the placeholder key — the gateway path is complete)

Also confirmed the MCP registry label is present on both arch variants of 1.3.0. The only other `--platform` mention in the repo is a Dockerfile comment about multi-arch *builds*, which is correct and untouched.

## Summary by CodeRabbit

* **Documentation**
  * Updated the Docker Quickstart command to rely on Docker’s default platform selection.


moonming:
Cold-context audit completed: **1 MEDIUM, 0 HIGH — fixed in 0bb6ec1** (commit message only; the one-line diff is unchanged).

- **M1 fixed** — the commit body said arm64 images exist "since v1.0". The auditor checked every tag manifest from 0.13.0 through 1.3.0: 1.0.0, 1.1.0 and 1.2.0 are all single-arch amd64; the first release tag carrying `linux/arm64` is **1.3.0-rc.1** (via #1194 / #1197). Reworded to "since v1.3.0 (#1194, #1197)" so a reader pinned to 1.0–1.2 isn't misled into dropping the flag.

Verified clean by the auditor: `latest`/`1.3.0` are a genuine OCI index with both arches and the MCP label on each variant; amd64 hosts resolve their native leg without the flag; the only other `--platform` in the repo is a correct Dockerfile comment about multi-arch builds; nothing in CI parses the README; shell continuation intact; every quickstart YAML field exists in `schemas/resources/*` at both `origin/main` and `v1.3.0`.

